PR #34, opened 2026-03-03
DYT-209: Support document updates in ingestion pipeline

msrdic:
## Summary
- When a document is re-ingested with the same `source` but different content, old chunks/entities/relationships are cleaned up and replaced — queries always return current information
- Uses `source` field as optional update key: non-empty source + different checksum = update; empty source = always create new (preserves current behavior)
- Checksum dedup runs first — identical content always skipped regardless of source
- Reuses document UUID on update (UPDATE not DELETE+INSERT) so external consumers keep stable references
- Entity/relationship cleanup removes the doc from `source_document_ids` arrays; deletes orphans where the array becomes empty; preserves shared entities referenced by other docs
- Adds `allow_update: bool = True` parameter to `remember()` for opt-out
- Adds `updated` field to `RememberResult` and `BatchResult`
- ADR: `docs/design/document-update-identity.md` documents the source-as-identity decision and trade-offs

### Files changed (13)
| Layer | Files | Change |
|-------|-------|--------|
| ADR | `docs/design/document-update-identity.md` | NEW — architectural decision record |
| Database | `alembic/versions/010_document_source_index.py`, `db/models.py` | Partial composite index on `(namespace_id, source)` |
| Storage | `postgresql.py`, `pgvector.py`, `neo4j.py`, `coordinator.py` | Source lookup, entity/relationship orphan cleanup |
| Engine | `engine.py`, `ingest.py`, `protocol.py` | Update detection in `remember()` and `stage_document()` |
| API | `memory_lake.py` | `allow_update` param, `updated` result field |
| Tests | `test_document_update.py`, `test_storage_coordinator.py` | 20 new + 4 new test cases |

## Test plan
- [x] All 1410 unit tests pass (0 failures, 4 skipped)
- [x] 20 new tests in `test_document_update.py` cover: new doc creation, checksum dedup, source-based update detection, document ID reuse, old chunk deletion, entity/relationship orphan cleanup, shared entity preservation, `allow_update=False` opt-out, batch updates, result metadata
- [x] 4 new tests in `test_storage_coordinator.py` cover: source lookup delegation, batch source lookup, cleanup orchestration
- [x] Coverage: 52.80% (threshold: 30%)
- [x] `make format` clean
- [x] `make lint` clean (only pre-existing weaviate/kuzu/logfire warnings)
- [ ] Manual verification: ingest a doc with source, re-ingest with changed content, verify old entities cleaned up

### DIFF
--- a/alembic/versions/010_document_source_index.py
+++ b/alembic/versions/010_document_source_index.py
@@ -1,0 +1,33 @@
+"""Add partial composite index on documents(namespace_id, source).
+
+Revision ID: 010_document_source_index
+Revises: 009_temporal_search_indexes
+Create Date: 2026-03-03
+
+Supports source-based document update detection: when a document is
+re-ingested with the same source but different content, the old data is
+cleaned up and replaced. The partial index (WHERE source != '') avoids
+indexing documents without a source, which cannot be auto-updated.
+"""
+
+from collections.abc import Sequence
+
+from alembic import op
+
+revision: str = "010_document_source_index"
+down_revision: str | Sequence[str] | None = "009_temporal_search_indexes"
+branch_labels: str | Sequence[str] | None = None
+depends_on: str | Sequence[str] | None = None
+
+
+def upgrade() -> None:
+    op.create_index(
+        "ix_documents_namespace_source",
+        "documents",
+        ["namespace_id", "source"],
+        postgresql_where="source != ''",
+    )
+
+
+def downgrade() -> None:
+    op.drop_index("ix_documents_namespace_source", table_name="documents")
--- a/docs/design/document-update-identity.md
+++ b/docs/design/document-update-identity.md
@@ -31,6 +31,7 @@ Use the existing `source` field (URL, file path, etc.) as an **optional update k
 - If a doc is first ingested without source, then later re-ingested with a source and different content → creates a second document (no retroactive identity assignment).
 - Concurrent updates to the same source race; **last writer wins** (no locking in v1).
 - On update: old chunks are deleted, entity/relationship `source_document_ids` are pruned, orphaned entities/relationships (sole source was this doc) are deleted.
+- Cross-backend cleanup (Neo4j + pgvector) is best-effort; if one backend fails mid-cleanup, stale references may persist until the next update of the same source.
 - The document UUID is **reused** on update so external consumers referencing doc IDs see continuity.
 
 ## Future Considerations
@@ -38,3 +39,4 @@ Use the existing `source` field (URL, file path, etc.) as an **optional update k
 - A dedicated `external_id` column could be added if `source` proves insufficient as an identity key.
 - Backfill support (`update_document_source()`) could be added to retroactively assign sources.
 - `SELECT ... FOR UPDATE` could be added if concurrent update races become problematic.
+- Error handling and retry logic for partial cleanup failures across backends.
--- a/docs/design/document-update-identity.md
+++ b/docs/design/document-update-identity.md
@@ -1,0 +1,40 @@
+# ADR: Document Update Identity
+
+## Status
+
+Accepted
+
+## Context
+
+Khora's ingestion pipeline treats every `remember()` call as a new document. The only dedup is checksum-based: identical content is skipped. When a document's content changes (same source URL, different text), the old chunks, entities, and relationships remain alongside the new ones, causing stale/contradictory query results.
+
+We need a mechanism to detect when a re-ingested document is an update of an existing one, clean up the old data, and replace it.
+
+Two competing approaches were considered:
+
+- **(A) Consumer-supplied stable ID:** Consumers pass an explicit `external_id`; Khora is fully agnostic about identity.
+- **(B) Auto-detect from existing metadata:** Khora uses an existing field (`source`) to detect updates without requiring new consumer-side changes.
+
+## Decision
+
+Use the existing `source` field (URL, file path, etc.) as an **optional update key** for v1.
+
+- When `source` is **non-empty** and a document with the same `(namespace_id, source)` exists but has a **different checksum** → treat as an update.
+- When `source` is **empty** → always create a new document (current behavior).
+- An `allow_update: bool = True` parameter allows consumers to opt out.
+- Checksum dedup runs **first** — if content is identical, skip regardless of source.
+
+## Implications
+
+- Documents ingested **without** `source` can never be auto-updated (no identity key).
+- Same content re-ingested under a different source URL is **skipped** by checksum dedup (not treated as a new doc for the new source).
+- If a doc is first ingested without source, then later re-ingested with a source and different content → creates a second document (no retroactive identity assignment).
+- Concurrent updates to the same source race; **last writer wins** (no locking in v1).
+- On update: old chunks are deleted, entity/relationship `source_document_ids` are pruned, orphaned entities/relationships (sole source was this doc) are deleted.
+- The document UUID is **reused** on update so external consumers referencing doc IDs see continuity.
+
+## Future Considerations
+
+- A dedicated `external_id` column could be added if `source` proves insufficient as an identity key.
+- Backfill support (`update_document_source()`) could be added to retroactively assign sources.
+- `SELECT ... FOR UPDATE` could be added if concurrent update races become problematic.
--- a/src/khora/db/models.py
+++ b/src/khora/db/models.py
@@ -215,6 +215,12 @@ class DocumentModel(Base):
     __table_args__ = (
         Index("ix_documents_namespace_checksum", "namespace_id", "checksum"),
         Index("ix_documents_namespace_source_type", "namespace_id", "source_type"),
+        Index(
+            "ix_documents_namespace_source",
+            "namespace_id",
+            "source",
+            postgresql_where="source != ''",
+        ),
     )
 
     def __repr__(self) -> str:
--- a/src/khora/engines/graphrag/engine.py
+++ b/src/khora/engines/graphrag/engine.py
@@ -20,6 +20,7 @@ from loguru import logger
 
 from khora.config import KhoraConfig, LiteLLMConfig
 from khora.core.models import Document, DocumentMetadata, Entity, MemoryNamespace, Organization, Workspace
+from khora.core.models.document import DocumentStatus
 from khora.extraction.embedders import LiteLLMEmbedder
 from khora.memory_lake import BatchResult, RecallResult, RememberResult, Stats
 from khora.query import HybridQueryEngine, QueryConfig, SearchMode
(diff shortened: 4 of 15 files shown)


msrdic:
## Code Review: DYT-209 — Document Update Pipeline

### HIGH Severity

**1. Non-atomic multi-backend cleanup can leave inconsistent state**
`coordinator.py:425-480`

`cleanup_document_references()` performs 5 sequential operations across 3 backends. If any step fails mid-way (e.g., Neo4j is down but pgvector is fine), the document row gets updated to `PENDING` and re-processed, but only *some* old data is cleaned up. This can result in duplicate entities/relationships after re-processing.

**Suggestion:** Wrap in a try/except that logs partial cleanup stats and still proceeds. Or catch and aggregate errors so the caller knows cleanup was partial.

---

**2. `get_document_by_source()` returns arbitrary row when multiple docs share the same source**
`postgresql.py:551-556`

The query uses `.first()` without an `ORDER BY`. Since there's no UNIQUE constraint on `(namespace_id, source)`, multiple documents can have the same source in the same namespace (e.g., if `allow_update=False` was used for one ingestion). The returned document is non-deterministic.

**Suggestion:** Add `.order_by(DocumentModel.updated_at.desc())` to return the most recently updated document. This ensures deterministic behavior.

---

**3. Same issue in batch: `get_documents_by_sources()` dict key collision silently drops documents**
`postgresql.py:574-582`

The result is `{m.source: ... for m in models}`. If multiple documents share a source, only the last one iterated ends up in the dict. Combined with no `ORDER BY`, the chosen document is non-deterministic.

**Suggestion:** Same as above — add an `ORDER BY` and/or document the assumption that source should be unique per namespace.

---

### MEDIUM Severity

**4. `remember_batch()` does not expose or pass through `allow_update`**
`engine.py:421-503`, `memory_lake.py:345-394`

`GraphRAGEngine.remember_batch()` doesn't accept an `allow_update` parameter and doesn't pass it to `ingest_documents()`. The pipeline picks it up via `kwargs.get("allow_update", True)` at `ingest.py:1346`, but the engine never sends it. No way for callers to opt out of update detection in batch mode.

**Suggestion:** Add `allow_update: bool = True` to `remember_batch()` in both the engine and `MemoryLake`, and pass through to `ingest_documents()`.

---

**5. Neo4j entity cleanup query may return no rows when only updates occur (no deletes)**
`neo4j.py:1464-1492`

The Cypher query uses `UNWIND to_delete AS d` after the aggregation. If all entities are just pruned (none deleted), `to_delete` is a list of nulls → after `WHERE d IS NOT NULL` → zero rows → `RETURN` never executes → `result.single()` returns `None` → method returns `([], [])` even though entities *were* updated.

**Suggestion:** Split into two queries (one for array removal, one for orphan deletion) or restructure the Cypher to avoid the UNWIND-gating-RETURN pattern.

---

**6. Cleanup stat merging uses `max()` — potentially misleading**
`coordinator.py:460-471`

When both graph and vector backends report cleanup counts, `max(graph_count, vector_count)` is used. If backends diverge (e.g., vector deletes 3, graph deletes 2), the stat is approximate.

**Suggestion:** Consider separate `graph_*` / `vector_*` stats, or document that counts are approximate.

---

### LOW Severity

**7. Duplicate `DocumentMetadata(...)` construction in 3 places**
`ingest.py:614-628`, `ingest.py:726-736`, `ingest.py:749-759`

Same 8-field metadata construction repeated in update and create paths of both `stage_document()` and `stage_documents_batch()`.

**Suggestion:** Extract `_build_metadata(doc_input, checksum)` helper.

---

**8. `source_type` hardcoded to `"api"` in engine but `"manual"` in pipeline**
`engine.py:284` vs `ingest.py:620`

`GraphRAGEngine.remember()` always sets `source_type="api"` for updates. Pipeline uses `doc_input.get("source_type", "manual")`. Different defaults for the same document depending on code path.

---

**9. No test for concurrent/duplicate source scenario**

No test covers the case where multiple documents exist with the same `(namespace_id, source)` — the scenario where findings #2/#3 manifest.

---

**10. Engine tests use `__new__()` to bypass `__init__`**
`test_document_update.py:97`

Works but fragile — if `__init__` adds new required attributes, tests break with confusing errors.

---

### Summary

| Severity | Count |
|----------|-------|
| HIGH | 3 |
| MEDIUM | 3 |
| LOW | 4 |

**Assessment: Request changes** — #2/#3 (non-deterministic source lookup) are correctness bugs that should be fixed before merge. #4 (batch `allow_update` gap) is important for API consistency. #5 (Neo4j Cypher edge case) should be verified.

msrdic:
## Post-Review Fixes

Addressed 6 findings from the structured code review, with 1 deferred to a follow-up ticket.

### Changes

**Fix 1: Best-effort cleanup with error tracking** (`coordinator.py`)
- Each of the 5 cleanup steps in `cleanup_document_references()` is now wrapped in individual try/except
- On failure, a warning is logged and the step name is appended to `stats["errors"]`
- Remaining steps continue executing so partial progress is not lost

**Fix 2 & 3: Deterministic source lookups** (`postgresql.py`)
- `get_document_by_source()` — added `ORDER BY updated_at DESC` so `.first()` returns the newest document when duplicates exist
- `get_documents_by_sources()` — added DESC ordering with dict-building loop that keeps only the first (newest) doc per source

**Fix 4: `allow_update` on `remember_batch()`** (`protocol.py`, `engine.py` x3, `memory_lake.py`)
- Added `allow_update: bool = True` to `MemoryEngineProtocol`, `GraphRAGEngine`, `SkeletonEngine`, `VectorCypherEngine`, and `MemoryLake`
- `GraphRAGEngine` passes it through to `ingest_documents()`; `MemoryLake` passes it to the engine

**Fix 7: `_build_document_metadata()` helper** (`ingest.py`)
- Extracted helper function that builds `DocumentMetadata` from a doc input dict + checksum
- Replaced all 4 identical construction sites in `stage_document()` and `stage_documents_batch()`
- Cleaned up now-unused `DocumentMetadata` imports in function bodies

**Fix 9: Unit tests for duplicate-source and partial failure** (`test_document_update.py`)
- `test_get_document_by_source_returns_newest` — verifies ORDER BY is present in compiled SQL
- `test_get_documents_by_sources_keeps_newest` — verifies batch lookup retains newest doc per source
- `test_cleanup_partial_failure_continues` — verifies cleanup continues when individual backend steps fail
- `test_cleanup_without_any_backends` — updated to assert empty `errors` list

**ADR update** (`document-update-identity.md`)
- Added cross-backend best-effort cleanup to Implications
- Added retry logic for partial failures to Future Considerations

### Deferred

- **DYT-212**: Neo4j cleanup stats reporting issues (Cypher UNWIND edge case + `max()` stat merging) — stats-only, no correctness impact

### Verification

- `make format` ✅
- `make lint` ✅ (only pre-existing warnings)
- `make test` ✅ — 1413 passed, 53% coverage

msrdic:
## Audit: PR #34 (DYT-209) — Document Update Pipeline

Full structured review of the PR in its current state (2 commits).

---

### HIGH

**1. Batch same-source race: double cleanup + silent data loss**
`src/khora/pipelines/flows/ingest.py:713-730`

In `stage_documents_batch()`, if two docs in the same batch share the same `source` (and neither is checksum-deduped), both iterations find `source in source_matches` and operate on the **same** Document object:
1. First iteration: cleanup old doc, mutate object, `update_document()`
2. Second iteration: cleanup **again** (double-delete), mutate same object, `update_document()` again

First document's content is silently lost and cleanup runs twice.

**Fix:** Track processed sources in the loop and skip/create-new for subsequent docs with same source.

---

**2. `BatchResult.updated` is never populated**
`src/khora/engines/graphrag/engine.py:531-540`

`BatchResult` has `updated: int = 0` field, but `remember_batch()` constructs it without passing `updated=`. The pipeline's `ingest_documents()` doesn't track or return how many docs were updates vs. new creates. So `BatchResult.updated` is always `0` even when documents were actually updated.

**Fix:** Have `ingest_documents()` return `updated_documents` count and pass it through to `BatchResult`.

---

**3. `GraphRAGEngine.remember()` metadata inconsistency**
`src/khora/engines/graphrag/engine.py:281-288, 302-308`

`_build_document_metadata()` helper was added to `ingest.py` and replaced 4 call sites, but `GraphRAGEngine.remember()` has 2 more `DocumentMetadata(...)` constructions that use `source_type="api"` and don't set `author`, `language`, or `content_type` — behavioral inconsistency with the pipeline helper which defaults those fields.

**Fix:** Align the `remember()` metadata construction with the pipeline's defaults, or extract a shared helper.

---

### MEDIUM

**4. Coordinator cleanup stats use `max()` — already tracked in DYT-212**
`src/khora/storage/coordinator.py:484-485, 496-497` — `max(graph_count, vector_count)` can undercount if backends diverge.

**5. Concurrent updates to same source have no protection**
`src/khora/engines/graphrag/engine.py:267-297` — Two simultaneous `remember()` calls with the same source can both cleanup the same doc concurrently. Acknowledged in ADR as "last writer wins (no locking in v1)".

**6. `test_storage_coordinator.py` uses loose `>=` assertions**
`tests/unit/test_storage_coordinator.py:399-401` — Mock returns deterministic values, so `>=` should be `==` to catch regressions.

**7. Missing test: `allow_update=True` with no existing source match**
`tests/unit/test_document_update.py` — No test for the common happy path: first ingestion with a source that doesn't match anything yet (should create new doc).

---

### LOW

**8. `_make_storage_mock()` missing `errors` key**
`tests/unit/test_document_update.py:66-74` — Mock `cleanup_document_references` return dict doesn't include `"errors": []` which production now always includes.

**9. PostgreSQL test mocks are brittle**
`tests/unit/test_document_update.py:526-547` — 16-field MagicMock objects to simulate ORM models; adding/renaming columns would silently break.

---

### Verdict

**REQUEST_CHANGES** — 3 HIGH findings need fixes before merge:
1. Batch same-source double-cleanup is a correctness bug
2. `BatchResult.updated` is dead code (always 0)
3. Metadata inconsistency between `remember()` and pipeline paths

msrdic:
## Audit Fix Commit: `b485cbb`

Addresses the 3 HIGH-severity findings from the PR audit:

### Fix 1: Batch same-source double-cleanup
Added `processed_sources: set[str]` tracking in `stage_documents_batch()`. When two docs in the same batch share a source, the first enters the update path normally; the second is skipped with a warning and returns `None`.

### Fix 2: `BatchResult.updated` never populated
`stage_documents_batch()` now returns `(results, updated_count)` tuple. The count flows through `ingest_documents()` → `remember_batch()` → `BatchResult(updated=...)`.

### Fix 3: Metadata inconsistency in `remember()`
Extracted `_make_metadata()` closure in `GraphRAGEngine.remember()` that sets `content_type="text/plain"`, `language="en"`, `author=""` — matching the pipeline helper `_build_document_metadata()` defaults.

### Verification
- All 1389 unit tests pass
- New test: `test_batch_duplicate_source_skips_second`
- `make format` + `make lint` clean

msrdic:
Outdated.

PR #40, opened 2026-03-05
DYT-233: Fix document_status and event_type enum uppercase values

sale87:
## Summary
- Add `values_callable` to `document_status` and `event_type` enum columns in `db/models.py`
- Same fix as DYT-231 (tenancy_mode) — SQLAlchemy sends `.name` (PENDING) instead of `.value` (pending)
- Fixes `invalid input value for enum document_status: "PENDING"` during `genesis create`

Linear: https://linear.app/deyta/issue/DYT-233/fix-document-status-and-event-type-enums-sending-uppercase-instead-of

## Test plan
- [ ] Run `genesis create` — documents should be inserted without enum errors
- [ ] Verify event_type inserts also work with lowercase values

### DIFF
--- a/src/khora/db/models.py
+++ b/src/khora/db/models.py
@@ -177,7 +177,12 @@ class DocumentModel(Base):
     )
     content: Mapped[str] = mapped_column(Text, nullable=False)
     status: Mapped[str] = mapped_column(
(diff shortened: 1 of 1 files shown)


probablyrobot (approved):
Approved.

PR #42, opened 2026-03-06
DYT-258: Parameterize entity extraction API

msrdic:
## Summary

Make Khora's entity extraction pipeline ontology-agnostic by accepting `entity_types` and `relationship_types` as optional parameters, per [ADR-003](https://github.com/DeytaHQ/hegemon/blob/main/architecture/decisions/003-make-khora-ontology-agnostic.md).

- Add `entity_types: list[str] | None` and `relationship_types: list[str] | None` to the full extraction call chain: `MemoryLake` → engine protocol → 3 engine implementations → ingest pipeline → extract task → LLM extractor → prompt
- Fix precedence: explicit params > expertise config > skill config > defaults
- Fix `_render_extraction_prompt()` to accept pre-resolved relationship types instead of re-computing internally
- Thread params through cross-chunk relationship extraction (`_extract_cross_chunk_relationships`)
- Skeleton engine accepts params for protocol compliance (no entity extraction)
- All parameters optional with `None` defaults — fully backward compatible
- 9 new unit tests covering defaults, custom types, precedence, empty list normalization, and API threading

### Files changed (10 source + 1 test)

| Layer | Files |
|-------|-------|
| Public API | `memory_lake.py` |
| Engine protocol | `engines/protocol.py` |
| Engine impls | `engines/graphrag/engine.py`, `engines/skeleton/engine.py`, `engines/vectorcypher/engine.py` |
| Pipeline | `pipelines/flows/ingest.py`, `pipelines/tasks/extract.py` |
| Extractor | `extraction/extractors/base.py`, `extraction/extractors/llm.py` |
| Tests | `tests/unit/test_parameterized_extraction.py` |

## Test plan

- [x] Unit tests pass (1470 passed, 4 skipped)
- [x] Coverage: 53.77% (threshold: 30%)
- [x] Lint: ruff + black + isort clean
- [x] Type check: ty passes (13 pre-existing diagnostics, 0 new)
- [ ] Manual verification: call `lake.remember(content, entity_types=["DRUG", "GENE"], relationship_types=["TARGETS"])` and confirm custom types appear in extraction prompt

### DIFF
--- a/src/khora/engines/graphrag/engine.py
+++ b/src/khora/engines/graphrag/engine.py
@@ -223,6 +223,8 @@ class GraphRAGEngine:
         source: str = "",
         metadata: dict[str, Any] | None = None,
         skill_name: str = "general_entities",
+        entity_types: list[str] | None = None,
+        relationship_types: list[str] | None = None,
     ) -> RememberResult:
         """Store content in the memory engine.
 
@@ -287,6 +289,8 @@ class GraphRAGEngine:
             skill_name=skill_name,
             embedding_model=self._config.llm.embedding_model,
             extraction_model=self._config.llm.extraction_model or self._config.llm.model,
+            entity_types=entity_types,
+            relationship_types=relationship_types,
         )
         timings["pipeline_ms"] = (time.perf_counter() - start) * 1000
         timings["total_ms"] = (time.perf_counter() - total_start) * 1000
@@ -389,6 +393,8 @@ class GraphRAGEngine:
         deduplicate: bool = True,
         infer_relationships: bool = True,
         on_progress: Callable[[int, int], None] | None = None,
+        entity_types: list[str] | None = None,
+        relationship_types: list[str] | None = None,
     ) -> BatchResult:
         """Store multiple documents with automatic optimization.
 
@@ -461,6 +467,8 @@ class GraphRAGEngine:
             shared_embedder=shared_embedder,
             shared_entity_index=shared_entity_index,
             enable_expansion=infer_relationships,
+            entity_types=entity_types,
+            relationship_types=relationship_types,
         )
         timings["ingest_pipeline_ms"] = (time.perf_counter() - start) * 1000
         timings["total_ms"] = (time.perf_counter() - total_start) * 1000
--- a/src/khora/engines/protocol.py
+++ b/src/khora/engines/protocol.py
@@ -54,6 +54,8 @@ class MemoryEngineProtocol(Protocol):
         source: str = "",
         metadata: dict[str, Any] | None = None,
         skill_name: str = "general_entities",
+        entity_types: list[str] | None = None,
+        relationship_types: list[str] | None = None,
     ) -> RememberResult:
         """Store content in the memory engine.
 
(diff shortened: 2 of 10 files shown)


msrdic:
## Code Review: DYT-258 — Parameterize entity extraction API

### HIGH Severity

No high-severity findings.

### MEDIUM Severity

**M1. Defensive fallback in `_render_extraction_prompt` uses falsy check instead of `is None`**
- File: `src/khora/extraction/extractors/llm.py`
- The new code: `if not relationship_types:` will treat `[]` as falsy and fall back to defaults. However, by this point `relationship_types` should already be resolved (the caller normalizes `[]` → `None` → `DEFAULT_*`). This creates a second, implicit normalization layer. If someone calls `_render_extraction_prompt` directly with `relationship_types=[]`, the behavior silently falls back to defaults rather than producing an empty list. This is consistent with the PR's intent but is a hidden contract — anyone adding a new call site could be surprised. Consider using `if relationship_types is None:` for clarity and consistency with the rest of the PR.

**M2. `_extract_multi_batch` same falsy guard issue**
- File: `src/khora/extraction/extractors/llm.py`
- Same pattern: `if not relationship_types:` falls back to defaults. Same concern as M1 — should use `is None` for consistency with the precedence logic used everywhere else.

**M3. Missing `relationship_types` threading in `_extract_cross_chunk_relationships` context**
- File: `src/khora/pipelines/flows/ingest.py`
- The `extract_multi` call inside `_extract_cross_chunk_relationships` now receives `entity_types` and `relationship_types`, but the `expertise` kwarg is not passed (wasn't before either). This means the cross-chunk extraction always uses the default expertise path. This is a pre-existing gap, not introduced by this PR, but worth noting since the new parameterization partially papers over it.

### LOW Severity

**L1. `pass  # use as-is` pattern is unusual**
- Files: `src/khora/extraction/extractors/llm.py`, `src/khora/pipelines/tasks/extract.py`
- The `if entity_types is not None: pass` pattern works but is slightly unconventional. Could use `if entity_types is None:` with the body restructured. Very minor style nit.

**L2. No test for skill-level fallback in `extract_entities` task**
- File: `tests/unit/test_parameterized_extraction.py`
- Tests verify explicit params override defaults and thread through the pipeline, but there's no test for the precedence step: `explicit > expertise > skill > default`. Specifically, no test verifies that when `entity_types=None` and no expertise is set, `skill.entity_types` is used.

**L3. `resolved_entity_types = entity_types or None` redundant with prior normalization**
- File: `src/khora/pipelines/flows/ingest.py`
- In `stream_extract_and_embed_entities`, the code does `resolved_entity_types = entity_types or None` but the caller (`extract_entities` task) already normalizes `[]` → `None`. This double normalization is harmless but adds noise.

**L4. Test file uses one-class-per-test pattern**
- File: `tests/unit/test_parameterized_extraction.py`
- Each test is wrapped in its own class with a single method. Simple functions with `@pytest.mark.unit` would be more concise. Very minor style preference.

### Summary

| Severity | Count |
|----------|-------|
| HIGH     | 0     |
| MEDIUM   | 3     |
| LOW      | 4     |

**Overall assessment: ✅ Approve**

The PR correctly threads `entity_types` and `relationship_types` through the entire pipeline (MemoryLake → Engine → ingest pipeline → extract task → LLM extractor → prompt). The precedence logic (`explicit > expertise > default`) is correct and consistent. Empty-list normalization is handled properly. All paths are covered including cross-chunk extraction. The medium findings are about consistency (`not x` vs `is None`) rather than correctness — the code works correctly today because callers always pre-resolve types before hitting those internal methods.

PR #86, opened 2026-03-18
DYT-595: Add missing enum values to document_status on create_tables()

msrdic:
## Summary
- Databases created by older Khora versions may lack `pending` and/or `archived` in the PostgreSQL `document_status` enum. Since `Base.metadata.create_all` only creates the type if absent, library upgrades never added the missing values — causing `asyncpg.InvalidTextRepresentationError` on document inserts in downstream consumers (Poros).
- Added `sync_enum_values()` helper (`src/khora/db/schema.py`) that issues `ALTER TYPE … ADD VALUE IF NOT EXISTS` in autocommit mode after `create_all`
- Called from all three `create_tables()` methods: `PostgreSQLBackend`, `PGVectorBackend`, `PostgreSQLEventStore`
- Added Alembic migration `014_sync_document_status_enum` for managed deployments
- Added unit tests for the new helper

## Test plan
- [x] Unit tests pass (1585 passed)
- [x] Lint + type check clean (`make lint`)
- [x] Coverage ≥30% (52.84%)
- [ ] Manual verification: deploy to staging and confirm documents with `status='pending'` insert successfully

### DIFF
--- a/alembic/versions/014_sync_document_status_enum.py
+++ b/alembic/versions/014_sync_document_status_enum.py
@@ -1,0 +1,39 @@
+"""Add missing 'pending' and 'archived' values to document_status enum.
+
+Revision ID: 014_sync_document_status_enum
+Revises: 013_drop_previous_version_id
+Create Date: 2026-03-18
+
+The Python DocumentStatus enum defines PENDING, PROCESSING, COMPLETED, FAILED,
+and ARCHIVED, but databases created by older Khora versions (via create_tables()
+or earlier migrations) may only have a subset of these values.
+
+ALTER TYPE ... ADD VALUE IF NOT EXISTS is used so this migration is safe to run
+on databases that already have the values.
+
+Note: ADD VALUE cannot run inside a transaction, so autocommit mode is required.
+"""
+
+from collections.abc import Sequence
+
+from alembic import op
+
+# revision identifiers
+revision: str = "014_sync_document_status_enum"
+down_revision: str = "013_drop_previous_version_id"
+branch_labels: str | Sequence[str] | None = None
+depends_on: str | Sequence[str] | None = None
+
+
+def upgrade() -> None:
+    # ADD VALUE IF NOT EXISTS is idempotent — safe if values already present.
+    # Must run outside a transaction block.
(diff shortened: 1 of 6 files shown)


msrdic:
## Code Review — DYT-595

### HIGH

**1. Parameterized bind won't work for DDL statements** — `src/khora/db/schema.py:32`

```python
stmt = text(f"ALTER TYPE {type_name} ADD VALUE IF NOT EXISTS :val")
await conn.execute(stmt, {"val": value})
```

PostgreSQL does not support parameter binding in DDL statements. `ALTER TYPE ... ADD VALUE` expects a string literal, not a placeholder. asyncpg uses the extended query protocol which requires `PREPARE`, but only `SELECT/INSERT/UPDATE/DELETE/MERGE/VALUES` can be prepared. **This will fail at runtime against a real database.**

The unit tests use mocks and don't catch this.

**Suggested fix:** Since `value` comes from a hardcoded Python enum (not user input), direct string interpolation is safe:
```python
stmt = text(f"ALTER TYPE {type_name} ADD VALUE IF NOT EXISTS '{value}'")
await conn.execute(stmt)
```

---

### MEDIUM

**2. Overly broad `except Exception` could mask connection errors** — `src/khora/db/schema.py:35`

Catches everything — including connection drops, auth failures, or timeouts. Consider catching `sqlalchemy.exc.ProgrammingError` for the "type doesn't exist" case and letting infrastructure errors propagate.

**3. `sync_enum_values` called from event store unnecessarily** — `src/khora/storage/event_store.py:113`

The event store only deals with `MemoryEventModel` — it doesn't use `document_status`. Calling `sync_enum_values` here adds an unnecessary round-trip.

---

### LOW

**4. `pass` in migration downgrade** — `alembic/versions/014_sync_document_status_enum.py:39`

Intentional — PostgreSQL doesn't support `DROP VALUE`. Already documented. No action needed.

---

| Severity | Count |
|----------|-------|
| HIGH     | 1     |
| MEDIUM   | 2     |
| LOW      | 1     |

**Verdict: Request changes** — #1 must be fixed before merge.

sale87 (approved):

PR #134, opened 2026-03-25
DYT-1114: Fix VectorCypher engine dropping non-scalar chunk metadata

filippetrovic:
## Summary
- Remove the `isinstance(v, (str, int, float, bool))` filter from three locations in the VectorCypher engine (`_process_document`, `_process_document_streaming`, `remember_batch`) that silently dropped non-scalar metadata values (lists, dicts, None) when propagating document metadata to chunks
- Both storage backends handle nested types correctly: pgvector uses JSONB (native nested support), Neo4j uses `serialize_dict()` which JSON-serializes the entire dict
- This aligns VectorCypher with GraphRAG's existing behavior, which already preserves all metadata types
- Added 5 regression tests in `test_vectorcypher_bugs.py` covering list, dict, None, mixed-type preservation, and a source-level guard against reintroduction

## Test plan
- [x] All 1931 unit tests pass
- [x] Pre-commit hooks pass (black, isort, ruff, ty, pyupgrade)
- [x] New regression tests verify non-scalar metadata types are preserved
- [x] Source-level test ensures the isinstance filter pattern does not reappear

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -227,3 +227,9 @@ scripts/slack-notify.py
 
 # TTOJ agent skills (deployed per-machine, not tracked)
 .agents/
+.ttoj-worktrees/
+# Claude Code user-specific files
+.claude/CLAUDE.md
+.claude/settings.local.json
+.claude/projects/
+.claude/worktrees/
--- a/docs/AI_CHANGELOG.md
+++ b/docs/AI_CHANGELOG.md
@@ -24,3 +24,4 @@
 - 2026-03-24: DYT-882: Remove auto-initialize from pyo3 (manylinux compat)
 - 2026-03-24: DYT-984: Replace __slots__ detail tests with behavioral checks
 - 2026-03-25: DYT-882: Fix publish-accel uv cache failure
+- 2026-03-25: DYT-1114: Fix VectorCypher engine dropping non-scalar chunk metadata
--- a/src/khora/engines/vectorcypher/engine.py
+++ b/src/khora/engines/vectorcypher/engine.py
@@ -612,7 +612,7 @@ class VectorCypherEngine:
                         "chunk_index": i,
                         "start_char": raw_chunk.start_char if hasattr(raw_chunk, "start_char") else 0,
                         "end_char": raw_chunk.end_char if hasattr(raw_chunk, "end_char") else len(raw_chunk.content),
-                        **{k: v for k, v in doc_metadata.items() if isinstance(v, (str, int, float, bool))},
+                        **doc_metadata,
                     },
                 )
                 temporal_chunks.append(temporal_chunk)
@@ -978,7 +978,7 @@ class VectorCypherEngine:
                     "chunk_index": i,
                     "start_char": raw_chunk.start_char if hasattr(raw_chunk, "start_char") else 0,
                     "end_char": raw_chunk.end_char if hasattr(raw_chunk, "end_char") else len(raw_chunk.content),
-                    **{k: v for k, v in doc_metadata.items() if isinstance(v, (str, int, float, bool))},
+                    **doc_metadata,
                 },
             )
             temporal_chunks.append(temporal_chunk)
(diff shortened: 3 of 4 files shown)


filippetrovic:
## Consolidated Review — DYT-1114: Fix VectorCypher non-scalar chunk metadata

**Reviewers:** Security, Performance, Correctness, Test, Docs, Devil's Advocate + Security Audit
**Verdict:** REQUEST_CHANGES — 1 HIGH finding must be fixed before merge

---

### HIGH

- **H1 [Correctness / Performance / Devil's Advocate]** `src/khora/engines/vectorcypher/engine.py:615,981,1568` — **Metadata key collision: `**doc_metadata` overwrites internal keys.** The `**doc_metadata` spread is placed *after* the fixed keys (`chunk_index`, `start_char`, `end_char`) in the dict literal. In Python, later keys win — so if `doc_metadata` contains any of these keys, the internal values are silently overwritten. This corrupts chunk positional data used for fragment reassembly and source attribution. **Fix:** Reverse the order so fixed keys come last:
  ```python
  metadata={
      **doc_metadata,
      "chunk_index": i,
      "start_char": ...,
      "end_char": ...,
  }
  ```

### MEDIUM

- **M1 [Test / Devil's Advocate]** `tests/unit/engines/test_vectorcypher_bugs.py:240-305` — **Tests verify dict construction, not engine behavior.** The 4 data-type tests construct metadata dicts in isolation using plain Python spreading. They prove Python's dict semantics, not that the VectorCypher engine preserves non-scalar metadata through the actual `remember()` → storage → `recall()` pipeline. The `test_no_isinstance_filter_in_engine` source-inspection test is also fragile (won't catch functionally equivalent reimplementations of the filter). Consider adding a mock-based test that calls the engine method and inspects the `TemporalChunk.metadata` produced.

- **M2 [Security / Performance]** `src/khora/engines/vectorcypher/engine.py:615,981,1568` — **No metadata size/depth validation.** With the filter removed, arbitrarily large or deeply nested metadata can now flow to storage. `json.dumps()` has no depth limit, and JSONB/Neo4j accept unbounded payloads. In batch mode (~line 1537), this compounds across all documents in memory. Not exploitable for injection, but a DoS vector if callers pass very large metadata. **Note:** This is a pre-existing concern (the old filter didn't limit size of scalar values either) and is out of scope for this bugfix PR. Consider addressing separately.

- **M3 [Devil's Advocate]** `src/khora/engines/vectorcypher/engine.py:615,981,1568` — **Non-JSON-serializable types will cause runtime errors.** The old filter implicitly excluded `datetime`, `UUID`, custom objects, etc. Now these will reach `json.dumps()` in `serialize_dict()` and raise `TypeError`. The public API type hint is `dict[str, Any]` which permits these types. **Note:** This is also a pre-existing issue — GraphRAG has the same behavior. Both engines will fail on non-JSON types. Out of scope for this PR but worth a follow-up.

### LOW

- **L1 [Docs]** `src/khora/memory_lake.py:376` — The `remember()` docstring says only "Optional metadata" without specifying supported types. Adding a note like "JSON-serializable dict" would help callers avoid passing non-serializable values.

- **L2 [Test]** `tests/unit/engines/test_vectorcypher_bugs.py:232` — Test class docstring mentions "datetime" as a supported type but no test covers it. Either remove the mention or add a test (noting that `datetime` is actually *not* JSON-serializable without a custom encoder).

### INFO

- No injection risks (SQL or Cypher). Both storage paths use parameterized queries. JSONB via SQLAlchemy ORM; Neo4j via `serialize_dict()` passed as a Cypher parameter.
- No auth/authz changes, no new endpoints, no secret handling, no dependency changes.
- AI_CHANGELOG entry is correct and follows format.
- The change is backward-compatible — previously working scalar metadata continues to work.

---

## Security Audit

**Verdict: PASS**

Scanned all 16 OWASP categories against the PR diff. No injection, auth, crypto, access control, SSRF, XSS, deserialization, file handling, redirect, mass assignment, race condition, ReDoS, supply chain, or schema enforcement findings. The only security-adjacent concern (DoS via large metadata, M2 above) is pre-existing and not introduced by this PR.

---

## Overall Assessment

**REQUEST_CHANGES** — Fix H1 (key collision) before merge. M1-M3 and L1-L2 are recommendations for follow-up, not blockers.

probablyrobot (approved):

PR #201, opened 2026-03-31
DYT-1447: Fix transaction abort on fresh DB in do_run_migrations

filippetrovic:
## Summary
- **Bug fix**: `run_migrations()` was raising `InFailedSQLTransactionError` on a fresh PostgreSQL database (no `khora_alembic_version` table).
- **Root cause**: `do_run_migrations` tried `SELECT version_num FROM khora_alembic_version` inside a `context.begin_transaction()` block. When the table doesn't exist, PostgreSQL immediately puts the entire transaction into ABORTED state. The surrounding `try/except` caught the Python exception but left the transaction broken, so the subsequent `context.run_migrations()` call failed.
- **Fix**: Wrap the version-table query in a `SAVEPOINT`/`ROLLBACK TO SAVEPOINT`. On failure (table absent or any error), `ROLLBACK TO SAVEPOINT` restores transaction health and ahead-detection is skipped — one fewer round-trip than the `information_schema.tables` pre-check approach, no existence query needed.

## Test plan
- [x] `test_proceeds_when_version_table_absent` — verifies 4 execute calls (advisory lock → SAVEPOINT → failed SELECT → ROLLBACK TO SAVEPOINT) and that `run_migrations` is called
- [x] `test_proceeds_when_version_select_fails` — covers permission-denied and other SELECT failures after SAVEPOINT
- [x] `_setup_conn` updated to model the new 4-call sequence (advisory lock → SAVEPOINT → version SELECT → RELEASE SAVEPOINT)
- [x] All 49 migration tests pass
- [x] `ruff check` and `ruff format` clean on changed files

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -74,6 +74,7 @@ Format: versions match git tags (`git tag vX.Y.Z`). Versions before 0.5.1 were i
 
 ### Bug fixes
 
+- Fix `run_migrations()` raising `InFailedSQLTransactionError` on a fresh PostgreSQL database — existence check now uses `information_schema.tables` (SQL-standard, search_path-aware) instead of querying the version table directly inside a transaction (#201, DYT-1447)
 - Fix `_parse_uuid` for non-UUID SurrealDB record IDs (#168)
 - Fix SurrealDB ingestion performance regression (#160)
 - Fix `temporal_chunk` tags: coerce JSON strings to native arrays (#158)
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -198,6 +198,7 @@ Versions from git tags — no manual bumps. `git tag vX.Y.Z && git push origin v
 - **Advisory lock:** `run_migrations()` uses `pg_advisory_xact_lock` (ID `6001515088189075507`), 60s timeout
 - **Migrations bundled** in `src/khora/db/migrations/`, not `alembic/`. Root `alembic.ini` is dev-only
 - **Skip-ahead:** When multiple services share a DB with different Khora versions, `run_migrations()` detects if the DB revision is unknown (ahead) and skips gracefully — returns `MigrationResult(success=True, skipped=True)`. Signaled via `_DatabaseAheadError` from `env.py` to `session.py`
+- **Fresh-DB behavior:** On a PostgreSQL database with no `khora_alembic_version` table yet, `run_migrations()` / `MemoryLake(run_migrations=True)` correctly creates all tables from scratch. Prior to v0.6.6 (DYT-1447), querying the missing version table inside an explicit transaction caused `InFailedSQLTransactionError`. The fix uses `information_schema.tables` to check table existence before querying it — never issuing a statement that could abort the transaction.
 
 ### UUID & Type Handling
 - **ORM:** all 52 UUID columns use `as_uuid=True` — native `uuid.UUID`, never `str()` wrap
--- a/src/khora/db/migrations/env.py
+++ b/src/khora/db/migrations/env.py
@@ -113,6 +113,8 @@ def do_run_migrations(connection: Connection) -> None:
         # Use pg_catalog to check existence first — querying a missing table inside
         # an explicit transaction puts PostgreSQL into ABORTED state, which would
         # prevent context.run_migrations() from running (InFailedSQLTransactionError).
+        # pg_catalog is a system catalog that is always readable by any connected user,
+        # so this check never triggers a transaction abort regardless of DB state. (DYT-1447)
         table_exists = connection.execute(
             text(
                 "SELECT EXISTS(SELECT 1 FROM pg_catalog.pg_tables WHERE schemaname = 'public' AND tablename = :table)"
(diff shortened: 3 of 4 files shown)


filippetrovic:
## 🔍 Team Review — PR #201

> **Overall: REQUEST CHANGES** — Core fix is correct. Two HIGH findings must be addressed before merge.

---

### CRITICAL
_None._

---

### HIGH

**H1** `env.py:118` **[Correctness / Database]** — **Hardcoded `schemaname = 'public'` breaks non-default-schema deployments**

The pg_catalog query filters by `schemaname = 'public'`. If Khora is deployed against a PostgreSQL instance where `search_path` points to a custom schema, the version table will exist but the existence check returns `False`. Ahead-detection is silently skipped — migrations run as if on a fresh DB, potentially re-running applied migrations or ignoring a DB ahead of known revisions.

Three paths to fix (in order of recommendation):
1. Use `information_schema.tables WHERE table_name = :table` (SQL-standard, respects search_path, consistent with `versions/001–013` in this repo)
2. Drop the `schemaname = 'public'` filter from the pg_catalog query (searches all schemas)
3. Read schema from Alembic's `version_table_schema` config attribute if set

**H2** `CLAUDE.md` **[Docs]** — **Missing gotcha for fresh-database migration behavior**

The "Migrations & Schema" gotchas document skip-ahead but not the fresh-DB fix. Since `run_migrations()` / `MemoryLake(run_migrations=True)` are public API and fresh-DB was broken until this PR, the fix warrants a gotcha entry.

---

### MEDIUM

**M1** `env.py:122–125` **[Correctness]** — No exception handling if version SELECT fails after table confirmed present. Original `try/except Exception` caught these; new code propagates unhandled, aborting the transaction.

**M2** `env.py:118` **[Database]** — Inconsistency: migration files `001/005/007/008/013` use `information_schema.tables` (SQL-standard, search_path-aware); this PR introduces `pg_catalog.pg_tables`. Switching to `information_schema` would also fix H1.

**M3** `CHANGELOG.md` **[Docs]** — DYT-1447 user-facing reliability fix missing from `[Unreleased]` section.

**M4** `tests/` **[Design]** — SAVEPOINT approach would be simpler (no schema assumption, 2-call mock instead of 3-call, more idiomatic PostgreSQL). Worth considering as an alternative or follow-up.

---

### LOW

**L1** `tests/test_migration_bundling.py:827` — Mock row uses `__getitem__` for tuple access; silently breaks if code switches to attribute access.

**L2** `tests/test_migration_bundling.py` — No test covers pg_catalog query failures (permission denied, connection dropped).

---

## Security Audit — PASS ✅

All 16 OWASP categories reviewed. New pg_catalog query uses named bind parameters (`:table`). `VERSION_TABLE` is a module-level constant. No injection risk. No secrets, no auth changes, no HTTP clients. Clean.

---

## Auto-fixes applied (committed to branch)

| Finding | Change |
|---|---|
| Fragile `str(text(...))` assertion | Replaced with bound-parameter dict check: `call_args_list[1][0][1] == {"table": env.VERSION_TABLE}` |
| Convoluted `.replace()` string logic | Removed — covered by above |
| Comment clarity | Added: pg_catalog is always readable + `(DYT-1447)` ticket ref |

---

### Summary

| Severity | Count |
|---|---|
| CRITICAL | 0 |
| HIGH | 2 (H1: schema assumption, H2: docs) |
| MEDIUM | 4 |
| LOW | 2 |
| Security | PASS |

filippetrovic:
## Review Response — DYT-1447

Thanks for the thorough review. Here's the disposition of each finding:

### Fixed

| Finding | What was done |
|---|---|
| **H1** `env.py:118` — Hardcoded `schemaname = 'public'` | Replaced `pg_catalog.pg_tables WHERE schemaname = 'public'` with `information_schema.tables WHERE table_name = :table` (approach 1 — SQL-standard, respects `search_path`) |
| **H2** `CLAUDE.md` — Missing fresh-DB gotcha | Added "Fresh-DB behavior" bullet to the Migrations & Schema gotchas section |
| **M1** `env.py:122–125` — No exception handling on version SELECT | Wrapped version SELECT in `try/except`; swallows errors after table confirmed present and proceeds without ahead-detection rather than aborting the transaction |
| **M2** `env.py:118` — Inconsistency with `information_schema` | Fixed by H1 fix — now uses `information_schema.tables` consistently with the rest of the codebase |
| **M3** `CHANGELOG.md` — DYT-1447 missing from `[Unreleased]` | Added entry to the Bug fixes section |
| **M4** `tests/` — SAVEPOINT approach | Added a code comment in `env.py` documenting the SAVEPOINT alternative (SAVEPOINT + rollback on failure, avoids the existence check entirely) as a future simplification. The current `information_schema` + exception-handling approach is kept per the H1 approach-1 direction. |

### Acknowledged / Won't Fix

| Finding | Disposition |
|---|---|
| **L1** Mock row uses `__getitem__` | **Accepted / Won't fix** — The mock intentionally mirrors the current tuple-access pattern (`row[0]`). If the code switches to attribute access, the test will fail loudly, which is the desired signal. No silent breakage. |
| **L2** No test for pg_catalog query failures | **Won't fix in this PR** — With the switch to `information_schema`, the existence check is a simple `SELECT EXISTS(...)` that cannot fail under normal operating conditions (no permissions required). Pg_catalog failure tests no longer apply. The new `test_proceeds_when_version_select_fails` covers post-confirmation failures (M1). |

msrdic (approved):

probablyrobot (approved):

PR #293, opened 2026-04-08
DYT-1948: Add configurable query timeout for Neo4j get_entity_neighborhoods

filippetrovic:
## Summary

Production benchmarks show `DualNodeManager.get_entity_neighborhoods` has a bimodal latency distribution: p50 = 30ms, p95 = 150ms, **p99 = 27.3s (max 86s)**. The outliers are caused by combinatorial explosion in the variable-length Cypher path traversal on highly-connected entities. These runaway queries exhaust the Neo4j connection pool and cascade into 34s+ recall latencies and HTTP 500s.

This PR adds a configurable per-transaction timeout that bounds the worst ~0.5-1% of outliers without affecting the common case.

### What changed

- **`Neo4jConfig.query_timeout: float | None`** (default `5.0`, `gt=0`). Settable via env var. `None` disables; `0`/negative rejected by validator.
- **`DualNodeManager.get_entity_neighborhoods`** wraps the managed-read unit-of-work with `neo4j.unit_of_work(timeout=...)` when configured. On timeout, logs a warning (namespace_id, entity_count, depth, error code — no PII) and returns `{}` (downstream callers already handle empty neighborhoods).
- **Exception filter is narrow:** only catches `ClientError` with `.code` in the explicit `_NEO4J_TIMEOUT_CODES` tuple — catches BOTH `Neo.ClientError.Transaction.TransactionTimedOut` (server-default) AND `…TransactionTimedOutClientConfiguration` (client-configured, which is what we actually emit via `unit_of_work`). Syntax errors, auth failures, and constraint violations still propagate normally.
- **Threaded through** `VectorCypherEngine.initialize()` → `DualNodeManager` and `VectorCypherRetriever` via a new keyword-only `neo4j_query_timeout` kwarg.
- **Fully backward-compatible:** `DualNodeManager(driver)` without the kwarg gets `query_timeout=None` (no behavior change). Existing tests unchanged.

### Why `unit_of_work(timeout=...)` and not `tx.run(timeout=...)`

The ticket suggested `tx.run(timeout=...)` but that kwarg does not exist on the Neo4j **Python** driver (the JS/Go drivers have it; Python does not). The correct idiom is `@unit_of_work(timeout=...)` applied to the read closure inside `session.execute_read`. Verified against `neo4j==6.1.0` source.

### Why retries don't amplify the budget

`Neo.ClientError.Transaction.TransactionTimedOut*` hydrates to `ClientError` with `is_retryable() == False`, so `session.execute_read`'s managed retry loop does NOT re-execute on timeout. The 5s budget is a hard ceiling, not an amortized 15s (3× 5s). Verified empirically against the installed driver.

### Out of scope (deliberate)

Per the ticket, only `get_entity_neighborhoods` gets the timeout. Other read methods (`get_relationships_between`, `get_chunks_by_entities`, `get_neighborhoods_batch`, etc.) do not have the same pathological query pattern and are not touched. Peras-side changes (exposing `query_timeout` via `KhoraSettings`) are tracked separately as peras is a different repo.

## Test plan

- [x] `make format && make test` — **2488 passed, 5 skipped**, 53.39% coverage (above 30% gate)
- [x] 21 new tests across 3 files:
  - `tests/unit/test_config_backends.py` — default 5.0, override, `None`, round-trip via discriminated-union loader, **parametrized rejection of `0`/`0.0`/`-1`/`-0.5`**
  - `tests/unit/engines/test_dual_nodes.py` — `_query_timeout` storage, keyword-only enforcement, `unit_of_work` called with correct timeout, `unit_of_work` skipped when `None`, **parametrized test over both members of `_NEO4J_TIMEOUT_CODES` verifying empty-dict return and warning log**, non-timeout `ClientError` re-raised, non-`Neo4jError` re-raised, timeout-disabled happy path
  - `tests/unit/engines/test_vectorcypher_retriever.py` — default kwarg is None, `neo4j_query_timeout=3.0` forwarded to inner `_dual_nodes`, driver=None short-circuits
- [x] Lint clean: ruff, black, isort, ty typecheck (all via pre-commit)
- [x] Manual verification: `Neo4jConfig(query_timeout=0)` now raises `ValidationError`; `Neo4jConfig().query_timeout == 5.0`

### DIFF
--- a/docs/AI_CHANGELOG.md
+++ b/docs/AI_CHANGELOG.md
@@ -5,3 +5,4 @@
 - 2026-04-03: DYT-1733: Normalize simple path relevance scores to [0,1]
 - 2026-04-01: DYT-1798: Switch discovery models to LiteLLM YAML config
 - 2026-04-01: DYT-1884: Add KhoraError exception hierarchy, fix silent swallowing
+- 2026-04-08: DYT-1948: Add configurable query timeout for Neo4j get_entity_neighborhoods
--- a/src/khora/config/schema.py
+++ b/src/khora/config/schema.py
@@ -78,6 +78,22 @@ class Neo4jConfig(BaseModel):
         description="Seconds of idle time after which connections are checked for liveness "
         "before being returned from the pool. None disables the check.",
     )
+    query_timeout: float | None = Field(
+        default=5.0,
+        gt=0,
+        description=(
+            "Per-transaction timeout in seconds for bounded read queries "
+            "(currently applied only to get_entity_neighborhoods). Bounds the "
+            "worst ~0.5-1% of outlier queries that otherwise hold connections "
+            "for 27+ seconds and exhaust the pool. The Neo4j server terminates "
+            "transactions exceeding this duration, raising ClientError with "
+            "code Neo.ClientError.Transaction.TransactionTimedOut* — the "
+            "client catches this and returns an empty neighborhood dict. "
+            "Set to None to disable entirely. Values <= 0 are rejected "
+            "(the driver would interpret 0 as 'run forever', which defeats "
+            "the purpose)."
+        ),
+    )
     entity_write_concurrency: int = Field(default=12, description="Max concurrent entity write transactions")
     relationship_write_concurrency: int = Field(default=8, description="Max concurrent relationship write transactions")
 
--- a/src/khora/config/schema.py
+++ b/src/khora/config/schema.py
@@ -81,17 +81,26 @@ class Neo4jConfig(BaseModel):
     query_timeout: float | None = Field(
         default=5.0,
         gt=0,
+        le=300,
         description=(
-            "Per-transaction timeout in seconds for bounded read queries "
-            "(currently applied only to get_entity_neighborhoods). Bounds the "
-            "worst ~0.5-1% of outlier queries that otherwise hold connections "
-            "for 27+ seconds and exhaust the pool. The Neo4j server terminates "
-            "transactions exceeding this duration, raising ClientError with "
-            "code Neo.ClientError.Transaction.TransactionTimedOut* — the "
-            "client catches this and returns an empty neighborhood dict. "
-            "Set to None to disable entirely. Values <= 0 are rejected "
-            "(the driver would interpret 0 as 'run forever', which defeats "
-            "the purpose)."
+            "Per-transaction timeout in seconds for bounded read queries. "
+            "Currently applied ONLY to DualNodeManager.get_entity_neighborhoods "
+            "in the VectorCypher engine — Neo4jBackend.get_neighborhood and "
+            "get_neighborhoods_batch in storage/backends/neo4j.py do NOT "
+            "currently respect this setting (see DYT-1948 follow-up). Bounds "
+            "the worst ~0.5-1% of outlier queries that otherwise hold "
+            "connections for 27+ seconds and exhaust the pool. The Neo4j "
+            "server terminates transactions exceeding this duration, raising "
+            "ClientError with code Neo.ClientError.Transaction.TransactionTimedOut* "
+            "— the client catches this and returns an empty neighborhood dict. "
+            "Set to None to disable entirely. Values <= 0 are rejected (the "
+            "driver would interpret 0 as 'run forever', which defeats the "
+            "purpose). Values above 300 seconds (5 minutes) are also rejected "
+            "as a sanity cap: any recall budget that tolerates a 5-minute "
+            "neighborhood lookup is better served by batched offline jobs, "
+            "not interactive reads. Override via env var "
+            "KHORA_STORAGE__GRAPH__QUERY_TIMEOUT (note the double underscores "
+            "— StorageSettings nests graph config via env_nested_delimiter='__')."
         ),
     )
     entity_write_concurrency: int = Field(default=12, description="Max concurrent entity write transactions")
--- a/src/khora/engines/vectorcypher/dual_nodes.py
+++ b/src/khora/engines/vectorcypher/dual_nodes.py
@@ -24,7 +24,7 @@ from neo4j import unit_of_work
 from neo4j.exceptions import ClientError
 
 from khora.storage.backends.mixins import deserialize_dict, serialize_dict
-from khora.telemetry import trace
+from khora.telemetry import trace, trace_span
 
 if TYPE_CHECKING:
     from neo4j import AsyncDriver
@@ -100,6 +100,13 @@ class DualNodeManager:
         self._driver = driver
         self._database = database
         self._query_timeout = query_timeout
+        # Pre-bind the unit_of_work decorator once. The factory call is
+        # cheap but non-zero and the produced decorator is fully reusable
+        # — see neo4j.unit_of_work: it closes over (metadata, timeout) and
+        # returns a plain wrapper function that can decorate any number of
+        # transaction callables. Hoisting shaves an allocation per call
+        # on the hot neighborhood lookup path.
+        self._timed_unit_of_work = unit_of_work(timeout=query_timeout) if query_timeout is not None else None
 
     async def ensure_indexes(self) -> None:
         """Create indexes for Chunk and TimeNode nodes."""
@@ -570,14 +577,15 @@ class DualNodeManager:
             )
             return [record.data() async for record in result]
 
-        # Apply the configured transaction timeout via unit_of_work. The Neo4j
-        # Python driver's `tx.run()` does NOT accept a timeout kwarg — timeouts
-        # must be set at transaction-begin time, which unit_of_work does by
-        # attaching metadata the session reads when starting the managed tx.
-        # TODO(follow-up): add a timeout counter / trace attribute so ops can
-        # alert on timeout frequency rather than relying on result_count=0.
-        if self._query_timeout is not None:
-            _work = unit_of_work(timeout=self._query_timeout)(_work)
+        # Apply the configured transaction timeout via the pre-bound
+        # unit_of_work decorator (hoisted in __init__). The Neo4j Python
+        # driver's `tx.run()` does NOT accept a timeout kwarg — timeouts
+        # must be set at transaction-begin time, which unit_of_work does
+        # by attaching metadata the session reads when starting the
+        # managed tx. Hoisting means we pay the factory cost once per
+        # DualNodeManager rather than once per query.
+        if self._timed_unit_of_work is not None:
+            _work = self._timed_unit_of_work(_work)
 
         try:
             async with self._driver.session(database=self._database) as session:
@@ -587,6 +595,22 @@ class DualNodeManager:
             # tuple, not a prefix match) so we don't swallow syntax errors,
             # auth failures, or constraint violations that are also ClientError.
             if exc.code in _NEO4J_TIMEOUT_CODES:
+                # Emit a dedicated child span so operators can alert on
+                # timeout frequency in Logfire/OTEL via a span-name filter
+                # (the ".timeout" suffix). The parent @trace span records
+                # result_count=0 on timeout, but a distinct span carries
+                # the timeout-specific attributes ops needs for dashboards
+                # (configured timeout, entity count, depth, error code).
+                with trace_span(
+                    "khora.neo4j.get_entity_neighborhoods.timeout",
+                    timeout_s=self._query_timeout,
+                    entity_count=len(entity_ids),
+                    depth=depth,
+                    code=exc.code,
+                    namespace_id=str(namespace_id),
+                    timeout_occurred=True,
+                ):
+                    pass  # attributes set via kwargs; no inner work
                 logger.warning(
                     "Neo4j get_entity_neighborhoods timed out after {timeout}s "
                     "(namespace_id={ns}, entity_count={n}, depth={d}, code={code}); "
@@ -596,10 +620,8 @@ class DualNodeManager:
                     n=len(entity_ids),
                     d=depth,
                     code=exc.code,
+                    timeout_occurred=True,
                 )
-                # TODO(DYT-1948-followup): emit dedicated telemetry counter for
-                # query timeouts. @trace already records result_count=0 which
-                # is the minimum viable signal but we want explicit alerting.
                 return {}
             raise
 
--- a/src/khora/engines/vectorcypher/dual_nodes.py
+++ b/src/khora/engines/vectorcypher/dual_nodes.py
@@ -20,6 +20,8 @@ from typing import TYPE_CHECKING, Any
 from uuid import UUID, uuid4
 
 from loguru import logger
+from neo4j import unit_of_work
+from neo4j.exceptions import ClientError
 
 from khora.storage.backends.mixins import deserialize_dict, serialize_dict
 from khora.telemetry import trace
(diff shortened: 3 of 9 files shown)


filippetrovic:
# /team:review — Multi-Perspective Review of PR #293

**Reviewers spawned:** Security, Performance, Correctness, Test, Database, Docs, Devil's Advocate + dedicated Security Audit pass.

**Note:** the Security Reviewer agent did not return findings (technical issue with message delivery); the dedicated Security Audit pass below covers the same ground with deeper OWASP Top 10 coverage.

---

## Consolidated Findings

### CRITICAL
*(none)*

### HIGH

**H1. [Devil's Advocate]** Silent failure without observability — on timeout the method returns `{}`, indistinguishable from "actually no neighbors found". The `@trace` decorator logs `result_count=0`, but ops cannot alert on timeout frequency specifically. A `TODO(DYT-1948-followup)` is left in the code for a dedicated counter, which means we ship a silent failure mode until the follow-up lands. **Recommendation:** file the follow-up ticket immediately after merge and attach it to the PR; or consider adding a minimal telemetry counter (`neo4j_query_timeouts_total{method="get_entity_neighborhoods"}`) in this PR.

**H2. [Database]** Server-side timeout interaction risk — if a production Neo4j deployment has `db.transaction.timeout` set lower than the 5.0s client value (e.g., 3.0s), the server's setting wins silently (per driver docs for Neo4j 4.2–5.2). Operators debugging why the configured client timeout doesn't fire would get confused. **Recommendation:** document this interaction in the `Neo4jConfig.query_timeout` field description.

**H3. [Test]** Integration test gap — all tests mock `session.execute_read` entirely. Unit tests verify that the `unit_of_work(timeout=...)` decorator is applied and that the exception branch returns `{}`, but they do NOT verify that the timeout actually fires against a real Neo4j server. If a future driver version changes how `unit_of_work` applies timeouts, our tests stay green while production regresses. **Recommendation:** check if the repo has a testcontainers fixture for Neo4j; if yes, add one end-to-end timeout test. If no, acceptable for this PR but file a follow-up for integration coverage.

### MEDIUM

**M1. [Database]** Scope mismatch with `Neo4jBackend` — the `query_timeout` field is on `Neo4jConfig`, but `Neo4jBackend` (`storage/backends/neo4j.py`) has analogous methods (`get_neighborhood`, `get_neighborhoods_batch`) with the same pathological variable-length path pattern that are NOT wrapped. Operators tuning `query_timeout` and hitting a backend path (not the VectorCypher engine path) will see no effect and be surprised. **Recommendation:** either (a) document explicitly that the field currently applies only to the VectorCypher engine's `get_entity_neighborhoods`, or (b) file a follow-up ticket to wrap the backend methods too.

**M2. [Docs]** `CLAUDE.md` Gotchas section should document the new behavior — current codebase notes pool tuning but not query timeouts. Add: "**Neo4j query timeout:** `get_entity_neighborhoods` times out at 5s by default, returning empty dict + warning log (no exception). Set `Neo4jConfig.query_timeout=None` to disable."

**M3. [Docs]** Env var documentation missing — `Neo4jConfig.query_timeout` can be set via env var (following the project's `KHORA_` prefix convention) but the field docstring does not mention the env var name. Users won't know how to set it without reading `pydantic-settings` docs.

**M4. [Docs]** `docs/architecture/storage-backends.md` (if it exists) should add a Neo4j query timeout subsection. Docs Reviewer flagged this explicitly.

**M5. [Test]** `test_wraps_work_with_unit_of_work_when_timeout_set` validates that `unit_of_work` was called with the correct `timeout=` kwarg, but `session.execute_read` is mocked to return `[]` and never actually invokes the wrapped function. The test asserts the decorator was applied but does not verify the wrapped function is actually passed through. **Recommendation:** have the mock `execute_read` capture its argument and verify it's the wrapped-and-decorated `_work` closure.

**M6. [Test]** `_make_neo4j_error()` helper uses `Neo4jError._basic_hydrate` (undocumented internal API). The docstring notes the risk but there's no `isinstance(ClientError)` guard INSIDE the factory itself — if Neo4j changes this API, the test silently constructs the wrong exception type. **Recommendation:** add `assert isinstance(exc, ClientError), f"expected ClientError but got {type(exc)}"` inside the factory to fail loudly on driver API changes.

**M7. [Test]** Warning-log assertion in `test_returns_empty_on_timeout_codes` checks kwargs but not the full log message template. A future refactor reordering log fields would pass the current assertion silently.

**M8. [Performance]** `unit_of_work(timeout=...)(_work)` closure rebind happens on every invocation. Overhead is microseconds (negligible vs. 30ms p50) but could be hoisted to `__init__` when `query_timeout` is configured. Not blocking. Performance Reviewer quantified as <1% of query latency.

**M9. [Devil's Advocate]** Default of 5.0s lacks rigorous justification — PR cites p95=150ms and p99=27.3s but does not show the distribution shape between them. If there's a cluster of legitimate queries at 3-8s, a 5s cap kills them. **Recommendation:** this is acceptable as an initial heuristic; plan to tune based on production timeout-counter data once the observability follow-up lands.

**M10. [Devil's Advocate]** Coupling to Neo4j 5.x-specific error code strings (`_NEO4J_TIMEOUT_CODES` tuple) — if Neo4j 6.x splits/merges/renames these codes, our catch will silently miss. **Mitigation already in place:** the architect documented the coupling in a comment, and the parametrized test covers both codes. Low-risk given we pin `neo4j==6.1.0`.

### LOW

**L1. [Security Audit]** `Neo4jConfig.query_timeout` has `gt=0` lower bound but no upper bound. A misconfigured `query_timeout=86400` would re-enable the unbounded behavior. Defense-in-depth only; requires explicit operator action to exploit. **Suggested fix:** `le=300` (5 minute ceiling) or add a recommended-max hint to the field description.

**L2. [Docs]** After merge, append to `docs/AI_CHANGELOG.md`: `- 2026-04-08: DYT-1948: Add configurable query timeout for Neo4j get_entity_neighborhoods`.

**L3. [Performance]** `logger.warning` on timeout path blocks the event loop synchronously (loguru default). At 1% timeout rate this is negligible; noted for awareness.

**L4. [Test]** `test_init_query_timeout_is_keyword_only` would benefit from an inline comment explaining *why* the kwarg is keyword-only (prevents positional-arg confusion with `database`).

**L5. [Devil's Advocate]** Per-method timeout soup risk — a single global field works today but may not scale if future tickets want different timeouts per read method. Track as known-tradeoff, revisit if it becomes an issue.

---

## Security Audit — PASS

Ran a dedicated pass against all 16 OWASP Top 10-style categories. Trust boundary analysis: this PR adds no new trust boundaries. The `query_timeout` config field is operator-controlled (env var / config file), not user-controlled.

| Category | Status |
|---|---|
| Injection | PASS (Cypher query unchanged; `depth` already clamped) |
| Broken Auth | N/A |
| Crypto/Secrets | PASS (log fields are non-sensitive) |
| Broken Access Control | N/A |
| Security Misconfiguration | PASS (secure-by-default) |
| XSS / Browser | N/A |
| SSRF | N/A |
| CSRF | N/A |
| Deserialization | N/A |
| File/Path | N/A |
| Open Redirects | N/A |
| Mass Assignment | N/A |
| Race Conditions | PASS (closure rebind is function-local) |
| ReDoS / Resource Exhaustion | **IMPROVES** (PR itself is a DoS fix) |
| Supply Chain | PASS (no new deps) |
| Boundary Validation | PASS (with L1 nit above) |

**Verdict: PASS.** The PR is a net security improvement — it mitigates the connection-pool-exhaustion DoS that caused HTTP 500s in production (per the ticket's incident report).

---

## Reviewer Verdicts

| Reviewer | Verdict |
|---|---|
| Correctness | **APPROVE** — no issues found; all edge cases verified |
| Performance | **APPROVE** — default 5s is sound, no retry amplification |
| Security (audit) | **APPROVE** (PASS) |
| Database | REQUEST_CHANGES (H2, M1) |
| Test | REQUEST_CHANGES (H3, M5, M6, M7) |
| Docs | REQUEST_CHANGES (M2, M3, M4) |
| Devil's Advocate | REQUEST_CHANGES (H1, M9, M10) |

## Overall Assessment

**NEEDS_DISCUSSION — ready to merge after addressing H1 / H2 as follow-ups**

The implementation is correct, the test coverage is strong, and there are no security or correctness blockers. The HIGH findings are **operational concerns**, not code bugs:

- **H1** (silent failure / observability gap) — mitigated by the existing `@trace` decorator; the dedicated counter is a legitimate follow-up rather than a blocker
- **H2** (server-side timeout interaction) — a documentation update, not a code change
- **H3** (integration test gap) — unit coverage of the exception branch is sufficient if the repo doesn't already use testcontainers for Neo4j

**Recommendation:** merge this PR and file these follow-ups in Linear:
1. **DYT-1948-followup-1:** Add Prometheus counter `neo4j_query_timeouts_total{method}` and a log-level metric. (addresses H1)
2. **DYT-1948-followup-2:** Document server-side `db.transaction.timeout` interaction in the field description. (addresses H2) — or do this in a tiny doc-only PR.
3. **DYT-1948-followup-3:** Extend timeout coverage to `Neo4jBackend.get_neighborhood*` methods if they show up in the p99 tail too. (addresses M1)
4. **DYT-1948-followup-4:** Document the new timeout in `CLAUDE.md` gotchas + `docs/architecture/storage-backends.md`. (addresses M2, M3, M4)

The MEDIUM test findings (M5-M7) are real but incremental — current tests still cover the observable behavior correctly; the improvements would harden against future driver API changes and log-format refactors. These can be addressed in a small follow-up test-quality PR.

---

_Review team spawned via `/team:review 293`. 7 independent reviewers + 1 dedicated security audit pass._

filippetrovic:
# Review-feedback fixup — status

Commit `9170b34` addresses 10 of the original review findings. The remaining 7 are deliberately deferred or won't-fix; rationale below.

## ✅ Fixed in this PR

| ID | Severity | Finding | Resolution |
|---|---|---|---|
| **H1** | HIGH | Silent failure / no observability counter | Added `trace_span("khora.neo4j.get_entity_neighborhoods.timeout")` with 6 structured attributes (`timeout_s`, `entity_count`, `depth`, `code`, `namespace_id`, `timeout_occurred`). Searchable by span name suffix in Logfire/OTEL. The deferred TODO is removed. |
| **H3** | HIGH | Integration test gap | New `tests/integration/test_dual_nodes_timeout_integration.py` exercises full composition path: `KHORA_STORAGE__GRAPH__QUERY_TIMEOUT` env var → `KhoraConfig` → `engine.py` `getattr` → `DualNodeManager` → `unit_of_work` hoist → `ClientError` catch → empty dict. Stubbed driver (no real Neo4j); see "Won't fix" below for the real-driver decision. |
| **M1** | MEDIUM | Scope mismatch with `Neo4jBackend` | `Neo4jConfig.query_timeout` field description now explicitly states the timeout applies ONLY to `DualNodeManager.get_entity_neighborhoods` and that `Neo4jBackend.get_neighborhood*` does NOT respect it (with a follow-up reference). |
| **M3** | MEDIUM | Env var documentation missing | Field description now documents the exact env var `KHORA_STORAGE__GRAPH__QUERY_TIMEOUT` with the double-underscore explanation (`env_nested_delimiter='__'`). Discovered via runtime probe — single-underscore variants all fail. |
| **M5** | MEDIUM | Test didn't actually exercise the wrapped function | `test_wraps_work_with_unit_of_work_when_timeout_set` rewritten with a spy that tags decorated functions with a sentinel + captures and invokes the function handed to `execute_read`. Also behaviour-locks the M8 hoist by asserting `unit_of_work` was called exactly once. |
| **M6** | MEDIUM | `_make_neo4j_error` factory fragility | `isinstance(ClientError)` + `exc.code == code` assertions added inside the helper itself. A future neo4j driver refactor of `_basic_hydrate` will fail loudly at test setup instead of silently passing. |
| **M7** | MEDIUM | Warning log assertions only checked kwargs | Tests now also assert the raw template string (`"{timeout}"`, `"{code}"`, `"{n}"`, `"{d}"` all present in `args[0]`) — guards against a refactor silently dropping structured fields or switching the log format. |
| **M8** | MEDIUM | `unit_of_work` factory called per-query | Hoisted to `DualNodeManager.__init__` as `self._timed_unit_of_work`. Verified `unit_of_work` is reusable per the neo4j driver source. |
| **L1** | LOW | No upper bound on `query_timeout` | `Field(gt=0, le=300, ...)` — values above 5 minutes are rejected as a sanity cap. Parametrized tests cover `301.0`, `1000.0`, `1e18` rejection plus a positive `[0.001, 300.0]` boundary test. |
| **L2** | LOW | Missing AI_CHANGELOG entry | Appended `- 2026-04-08: DYT-1948: Add Neo4j query timeout for get_entity_neighborhoods`. |

Test impact: **2497 passed, 5 skipped** (9 new tests added). Coverage 53.40%. CI green.

## ❌ Won't fix (with rationale)

| ID | Severity | Finding | Reason |
|---|---|---|---|
| **H2** | HIGH | Server-side `db.transaction.timeout` interaction needs documentation | The new field description does mention that "the Neo4j server terminates transactions exceeding this duration" but does NOT call out the specific Neo4j 4.2–5.2 precedence rule (server setting wins if lower than client). Deferred to a doc-only follow-up — adding more rationale to an already-long Pydantic field description hurts readability more than it helps. Will land in the same follow-up that updates `CLAUDE.md` and `docs/architecture/storage-backends.md`. |
| **M2** | MEDIUM | `CLAUDE.md` Gotchas section update | Out of scope per the original triage. Doc PR to follow. |
| **M4** | MEDIUM | `docs/architecture/storage-backends.md` Neo4j subsection | Out of scope per the original triage. Doc PR to follow. |
| **M9** | MEDIUM | 5s default justification weak | Accepted as an initial heuristic. Rigorous tuning requires production timeout-counter data, which the H1 telemetry now collects. Will revisit after observing real-world hit rates. |
| **M10** | MEDIUM | Coupling to Neo4j 5.x-specific error codes | Already mitigated: explicit tuple constant `_NEO4J_TIMEOUT_CODES` with a multi-line comment explaining the rationale, parametrized test covering both codes, runtime guard in the test factory that fails loudly if `Neo4jError._basic_hydrate` changes. The coupling is acceptable for our pinned `neo4j>=6.0` driver; a future Neo4j 7.x upgrade can revisit. |
| **L3** | LOW | `logger.warning` blocks the event loop on timeout path | Negligible at the expected 0.5–1% timeout rate. The synchronous blocking is microseconds and only fires on the cold path. Defer until production data shows it matters. |
| **L4** | LOW | `test_init_query_timeout_is_keyword_only` missing inline comment | Pure nit. The test name + assertion are self-explanatory. |
| **L5** | LOW | Per-method timeout soup risk | Tracked as a known design tradeoff. No action until a future ticket actually wants per-method tuning — at which point we'll refactor to a structured config (e.g., `read_timeouts: dict[str, float]`) instead of accumulating individual fields. |

## Follow-up tickets to file

1. **Doc PR** — addresses H2 (server-side interaction), M2 (CLAUDE.md gotcha), M4 (storage-backends.md subsection)
2. **Backend coverage** — wrap `Neo4jBackend.get_neighborhood` and `get_neighborhoods_batch` with the same timeout (M1 scope extension), if production data shows they're also in the p99 tail
3. **Default tuning** — once H1 telemetry has accumulated production timeout-frequency data, revisit the 5s default and the `le=300` cap with real numbers (addresses M9 and L1's tuning concern)

Ready for re-review and merge. Awaiting human approval.

filippetrovic:
**L3 follow-up filed:** [DYT-2050 — Add `enqueue=True` to loguru sinks to stop blocking async event loop](https://linear.app/deyta/issue/DYT-2050/khora-peras-add-enqueuetrue-to-loguru-sinks-to-stop-blocking-async)

On closer inspection, L3 is much bigger than a single call site. 631 `logger.*` calls across 79 async files in khora — all synchronous — and `src/khora/logging_config.py::setup_logging()` doesn't use `enqueue=True` on any of its three sinks. When khora is imported as a library (into Peras's async API server), every log call blocks the consumer's event loop.

DYT-2050 tracks the proper library-wide fix and the consumer-side audit (Peras, genesis, khora-benchmarks). Marking L3 as a known issue in this PR rather than won't-fix — the single-call-site fix wouldn't help anyway since the rest of khora is still synchronous.

msrdic (approved):

PR #373, opened 2026-04-14
DYT-2428: Wire external_id through MemoryLake API and all engines

filippetrovic:
## Summary
- Add `external_id: str | None = None` parameter to `MemoryLake.remember()` and forward it through the engine stack
- Update `MemoryEngineProtocol.remember()` signature with `external_id`
- Wire `external_id` through all 4 engines (vectorcypher, skeleton, graphrag, chronicle) in both `remember()` and batch paths
- Wire `external_id` through the shared ingest pipeline (`pipelines/flows/ingest.py`)
- `remember_batch()` reads `external_id` per-document from doc dicts — no signature change needed
- Add unit tests verifying round-trip for `remember()` and `remember_batch()` with and without `external_id`

Part of [ADR-050: Khora Document Identity Primitive](https://github.com/DeytaHQ/hegemon/blob/main/architecture/decisions/050-khora-document-identity-primitive.md). Depends on DYT-2427 (merged).

Linear: https://linear.app/deyta/issue/DYT-2428

## Test plan
- [x] `make format` passes (ruff format + ruff check)
- [x] `make test` passes (2707 passed, 5 skipped, coverage 53.72% ≥ 46% threshold)
- [x] `remember()` with `external_id="test-123"` stores value on Document
- [x] `remember_batch()` with mixed docs (some with external_id, some without) stores correctly per-doc
- [x] `remember()` without `external_id` works unchanged (backward compat)
- [x] All existing tests pass unchanged

### DIFF
--- a/src/khora/engines/chronicle/engine.py
+++ b/src/khora/engines/chronicle/engine.py
@@ -339,7 +339,6 @@ class ChronicleEngine:
         expertise: ExpertiseConfig | None = None,
         extraction_config_hash: str | None = None,
         chunk_strategy: ChunkStrategy | None = None,
-        external_id: str | None = None,
     ) -> RememberResult:
         """Store content in the memory engine.
 
@@ -359,7 +358,6 @@ class ChronicleEngine:
             expertise: Optional expertise config (ADR-022)
             extraction_config_hash: Optional hash for change detection
             chunk_strategy: Override chunking strategy for this call
-            external_id: Optional caller-supplied external identifier for the document.
 
         Returns:
             RememberResult with document_id and counts
@@ -406,7 +404,6 @@ class ChronicleEngine:
             content=content,
             metadata=doc_metadata,
             extraction_config_hash=extraction_config_hash,
-            external_id=external_id,
         )
         document = await storage.create_document(document)
         timings["document_create_ms"] = (time.perf_counter() - start) * 1000
@@ -1139,9 +1136,6 @@ class ChronicleEngine:
             }
             if extraction_config_hash is not None:
                 entry["extraction_config_hash"] = extraction_config_hash
-            ext_id = doc_data.get("external_id")
-            if ext_id is not None:
-                entry["external_id"] = ext_id
             doc_inputs.append(entry)
         timings["prepare_inputs_ms"] = (time.perf_counter() - start) * 1000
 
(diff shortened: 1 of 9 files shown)


filippetrovic:
## Consolidated Review — PR #373

**Reviewers:** Correctness, Test Coverage, Security, Performance, Devil's Advocate
**Overall:** APPROVE with minor suggestions

---

### MEDIUM

**M1. [Test]** `tests/unit/test_memory_lake.py` — Missing validation edge case tests. Tests verify happy-path pass-through but don't test `Document.__post_init__` rejection of blank strings or oversized (>512 char) `external_id`. The mocks bypass actual Document instantiation. Consider adding at least one test with `Document(external_id="")` to confirm ValueError propagation.

**M2. [Test]** `tests/unit/test_memory_lake.py:390` — Test assumes positional arg index (`call_args.args[0]`). Fragile if engine signature changes to keyword-only.

**M3. [Correctness]** `src/khora/storage/backends/sqlite.py:590-616` and `src/khora/storage/backends/surrealdb/relational.py:515-541` — `_row_to_document()` methods don't include `external_id` in the mapping. **Expected:** DYT-2429 covers SQLite/SurrealDB backends separately. `Document.external_id` defaults to `None` so retrieval works but won't round-trip `external_id` until DYT-2429 ships.

### LOW

**L1. [Docs]** Protocol `remember_batch()` docstring (`protocol.py:151-168`) doesn't mention `external_id` as an accepted key in document dicts.

**L2. [Docs]** API-layer docstrings don't mention the 512-char max or blank-string rejection from `Document.__post_init__`.

**L3. [Test]** No special character test for `external_id` (Unicode, quotes, SQL-like strings). Low risk since SQLAlchemy parameterizes queries, but good for confidence.

**L4. [Devil's Advocate]** Scope expanded beyond ticket spec (VectorCypher only → all 4 engines). This is defensible — prevents `TypeError` when `MemoryLake.remember()` passes `external_id` to non-VectorCypher engines. The alternative (VectorCypher only) would have introduced a runtime crash for other engine users.

### NOT AN ISSUE

- **Security:** No vulnerabilities found. Input validated at model layer (blank check + 512-char limit). SQLAlchemy parameterized queries prevent SQL injection. Namespace isolation maintained.
- **Performance:** Zero overhead. Write-only parameter, partial index on non-NULL values only, no new queries.
- **Backward compatibility:** Safe — `external_id: str | None = None` is kwarg-only with None default.

### Summary

Clean, surgical implementation. The all-engines scope was the right call (prevents TypeError). Tests adequately cover the pass-through contract. Minor gaps in validation edge-case tests and docstring completeness are non-blocking. **Approved.**

probablyrobot (approved):

PR #379, opened 2026-04-15
DYT-2497: Graceful degradation in VectorCypher retriever when Neo4j is unavailable

filippetrovic:
## Summary

- **Fix exception hierarchy bug**: The existing `except Neo4jError` fallback did not catch `ConnectionAcquisitionTimeoutError`, `ServiceUnavailable`, or `SessionExpired` because they inherit from `DriverError`, not `Neo4jError`. Replaced with a narrow tuple `_NEO4J_TRANSIENT_ERRORS` containing `ServiceUnavailable`, `ConnectionPoolError`, `SessionExpired`, `TransientError`.
- **Per-step graceful degradation**: Wrapped `_cypher_expand`, `_version_filter_entities`, and `_fetch_version_history` individually so each degrades independently. When `_cypher_expand` fails, the retriever preserves already-computed vector search results (entry entities + chunks from pgvector) instead of discarding them.
- **Skip graph-dependent steps on fallback**: When `graph_fallback=True`, `_fetch_chunks_from_entities` (which also uses Neo4j via DualNodeManager) is skipped to avoid a cascading failure.
- **Metadata flags for consumers**: Result metadata includes `graph_fallback: True` and `graph_error: "<ExceptionType>: <message>"` so downstream services (peras) can detect degraded results and surface a partial-result signal.
- **ClientError not swallowed**: Query bugs (`ClientError`) and other non-transient errors still propagate — only infra blips trigger degradation.

### Context

On 2026-04-15 07:53–07:55 UTC, a ~2min Neo4j outage caused every recall/ask request to 500 even though pgvector results completed in <1s. The timeout error (`ConnectionAcquisitionTimeoutError`) was never caught by the existing `except Neo4jError` because it's a `DriverError` subclass, not a `Neo4jError` subclass. This PR closes that gap.

## Test plan

- [x] `test_retrieve_graceful_degradation_on_neo4j_timeout` — `_cypher_expand` raises `ConnectionAcquisitionTimeoutError` → fallback path returns vector results with `graph_fallback=True`
- [x] `test_retrieve_does_not_swallow_client_error` — `ClientError` propagates (not caught)
- [x] `test_version_filter_graceful_degradation` — `ServiceUnavailable` from `_version_filter_entities` → filter skipped, retrieve succeeds
- [x] `test_version_history_graceful_degradation` — `ServiceUnavailable` from `_fetch_version_history` → version_history=None, retrieve succeeds
- [x] All 29 existing retriever tests pass (no regression)

### DIFF
--- a/src/khora/engines/vectorcypher/retriever.py
+++ b/src/khora/engines/vectorcypher/retriever.py
@@ -615,7 +615,7 @@ class VectorCypherRetriever:
             expanded_entities = {}
             entity_info_map = {}
             graph_fallback = True
-            graph_error_msg = f"{type(exc).__name__}: {str(exc)[:200]}"
+            graph_error_msg = type(exc).__name__
 
         # Step 4b: Bi-temporal version filtering
         # For EXPLICIT temporal queries with a parsed date, narrow entities to
@@ -934,7 +934,9 @@ class VectorCypherRetriever:
         entity_ids_str = [str(eid) for eid, _ in all_entity_scores]
 
         # Start relationship fetch immediately (doesn't need full Entity objects)
-        if self._dual_nodes is not None:
+        # Skip Neo4j relationship fetch when graph is unavailable to avoid
+        # blocking on a second timeout.
+        if self._dual_nodes is not None and not graph_fallback:
             rels_task = asyncio.create_task(
                 self._dual_nodes.get_relationships_between(
                     entity_ids_str,
@@ -1120,6 +1122,7 @@ class VectorCypherRetriever:
         # Update metadata to indicate fallback was used
         result.metadata["fallback_mode"] = "vector_only"
         result.metadata["graph_unavailable"] = True
+        result.metadata["graph_fallback"] = True
 
         return result
 
--- a/src/khora/engines/vectorcypher/retriever.py
+++ b/src/khora/engines/vectorcypher/retriever.py
@@ -24,7 +24,17 @@ from typing import TYPE_CHECKING, Any
 from uuid import UUID, uuid4
 
 from loguru import logger
-from neo4j.exceptions import Neo4jError
+from neo4j.exceptions import (
+    ServiceUnavailable,
+    SessionExpired,
+    TransientError,
+)
+
+try:
+    from neo4j.exceptions import ConnectionAcquisitionTimeoutError, ConnectionPoolError
+except ImportError:
+    ConnectionAcquisitionTimeoutError = ServiceUnavailable  # type: ignore[misc,assignment]
+    ConnectionPoolError = ServiceUnavailable  # type: ignore[misc,assignment]
 
 from khora.core.models import Chunk, ChunkMetadata, Entity, Relationship
 from khora.telemetry import trace_span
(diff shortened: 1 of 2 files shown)


filippetrovic:
## Consolidated Review — DYT-2497: Graceful Degradation in VectorCypher Retriever

**Reviewers**: Security, Performance, Correctness, Test, Devil's Advocate, Docs
**Security Audit**: PASS (no CRITICAL/HIGH)

---

### MEDIUM

- **M1 [Performance]** `retriever.py:937-1024` — When `graph_fallback=True`, the relationship fetch task (`rels_task`) still executes against Neo4j and may block for the full connection timeout before being caught by the existing `except Exception` at line 1020. This adds unnecessary latency on the fallback path. *Consider cancelling `rels_task` in the `_cypher_expand` exception handler.*

- **M2 [Security]** `retriever.py:618,1079` — `graph_error` metadata includes `str(exc)[:200]` which may contain Neo4j connection strings (`bolt://host:port`). Low risk since metadata is internal to the library and peras controls API exposure. *Consider sanitizing or using just the exception type name in consumer-facing metadata.*

- **M3 [Test]** `test_vectorcypher_retriever.py` — Only 2 of 4 transient error types tested (`ConnectionAcquisitionTimeoutError`, `ServiceUnavailable`). `SessionExpired` and `TransientError` are not exercised. *Consider parametrizing existing tests to cover all tuple members.*

- **M4 [Test]** `test_vectorcypher_retriever.py` — No log assertions. Warning logs are emitted at 3 fallback points but tests don't verify via `caplog`. *Nice-to-have for observability confidence.*

- **M5 [Correctness]** `retriever.py:632-642` — When `_version_filter_entities` fails, the retriever silently returns unfiltered entities without any metadata signal. Consumers can't distinguish "filter not applied due to Neo4j failure" from "filter succeeded." *Consider setting a `version_filter_skipped` flag or extending `graph_fallback` to cover this case.*

- **M6 [Correctness]** `retriever.py:1078` vs `retriever.py:1121-1122` — Two fallback paths use different metadata keys: inner fallback sets `graph_fallback=True`, outer fallback sets `fallback_mode="vector_only"` + `graph_unavailable=True`. *Consumers need to check multiple fields to detect degradation. Consider documenting the metadata contract or normalizing keys in a follow-up.*

### LOW

- **L1 [Performance]** `retriever.py:618` — Error message truncation at 200 chars may lose debugging context for long Neo4j errors. *Consider 500 chars or structured fields.*

- **L2 [Test]** `test_vectorcypher_retriever.py` — Missing test for `_fetch_chunks_from_entities` failure after successful `_cypher_expand` (outer fallback path). The path works (covered by outer `except _NEO4J_TRANSIENT_ERRORS`), but has no direct test. *Nice-to-have.*

- **L3 [Devil's Advocate]** `retriever.py:663-673` — `_fetch_chunks_from_entities` skip when `graph_fallback=True` is slightly pessimistic. If Neo4j recovered between `_cypher_expand` failure and this call, we'd miss valid chunk data. In practice, pool exhaustion affects the entire session uniformly, so this is the right trade-off.

### INFO

- **I1 [Security]** Exception hierarchy verified: `TransientError` subclasses are exclusively transient infrastructure errors (`DatabaseUnavailable`, `ForbiddenOnReadOnlyDatabase`, `NotALeader`). `ClientError` (query bugs), `AuthError`, `ConstraintError` are all on separate branches — correctly NOT caught.

- **I2 [Correctness]** Variable initialization is safe. `expanded_entities = {}` and `entity_info_map = {}` propagate correctly through all downstream `.get()` calls with safe defaults.

- **I3 [Correctness]** No placeholders, TODOs, FIXMEs, or stubs detected.

- **I4 [Performance]** Try/except overhead on the happy path is negligible (Python's exception machinery has near-zero cost when no exception is raised).

---

## Overall Assessment

**APPROVE** — The implementation is solid, well-structured, and fixes a real production issue (the existing `except Neo4jError` didn't even catch the timeout from the incident). The two-layer fallback provides defense in depth. The exception selection is well-reasoned.

**Must-fix before merge**: None.

**Should-fix (recommended)**: M1 (relationship fetch latency on fallback), M5 (version filter skip signal).

**Nice-to-have (follow-up)**: M2, M3, M4, M6, L1, L2.

msrdic (approved):

probablyrobot (approved):

PR #388, opened 2026-04-16
DYT-2626: Fix TypeError in Neo4jBackend.get_entity_relationships

filippetrovic:
## Summary

- One-line Cypher fix in `Neo4jBackend.get_entity_relationships` (`src/khora/storage/backends/neo4j.py`): change `RETURN r, ...` → `RETURN properties(r) as r, ...`, matching the pattern sibling `list_relationships` already uses. The raw `Relationship` value returned through `await result.data()` was being serialized to a `(start_node, rel_type, end_node)` 3-tuple, not a properties dict — so `_record_to_relationship` blew up on `rel["id"]` with `TypeError: tuple indices must be integers or slices, not str` on every non-empty result. After the fix, `data()` yields a plain dict (because `properties(r)` is a Cypher map), and `_record_to_relationship` is unchanged.
- Two new unit tests in `tests/unit/test_neo4j_backend.py` under `TestNeo4jBackendGetEntityRelationships`:
  - `test_returns_relationships_from_properties_dict` — exercises the non-empty path with the post-fix record shape and asserts `Relationship` objects come back with the right fields (ticket AC #1).
  - `test_raises_typeerror_on_raw_relationship_tuple` — regression-lock; feeds the pre-fix 3-tuple shape and asserts `TypeError` so we can't silently reintroduce the bug.

## Scope notes

- Verified ticket AC #3 via grep: no other Cypher queries in `neo4j.py` combine raw-relationship `RETURN r` with `await result.data()` + dict-indexing.
  - `get_relationship` uses `result.single()` + `record["r"]` — Record subscripting returns a raw `Relationship` (Mapping-compatible), unaffected.
  - `get_neighborhood` fallback uses `collect(DISTINCT r)` + `single()` + `_element_to_dict` — unaffected.
- **Follow-up flagged (out of scope per ticket):** `get_neighborhoods_batch` uses `collect(DISTINCT r)` + `await result.data()`, so the rels inside the list are 3-tuples and `_element_to_dict` falls through to `{"_raw": str(...)}` — degraded output (silent prop/id loss), not a TypeError. Worth a separate ticket; this PR is strictly the one-query semantic fix the ticket called out.

## Test plan

- [x] `uv run ruff format --check` — clean
- [x] `uv run ruff check` — clean
- [x] `uv run pytest tests/unit/test_neo4j_backend.py -v` — 15 passed (13 pre-existing + 2 new)
- [x] `uv run pytest --cov=src/khora --cov-fail-under=30` — 2771 passed, 5 skipped, coverage 54.22%
- [ ] Verified via failing trace `019d9516edfc2d44ab0e58af18bb13f4` namespace — **recommended next**: run peras recall against the same namespace locally and confirm the TypeError span is gone.

Closes DYT-2626.

🤖 Built with agent session `agent-ebd8d5`.

### DIFF
--- a/src/khora/storage/backends/neo4j.py
+++ b/src/khora/storage/backends/neo4j.py
@@ -1497,7 +1497,7 @@ class Neo4jBackend(GraphBackendBase):
         query = f"""
         MATCH {pattern}
         WHERE e.id = $entity_id
-        RETURN r, e.id as source_id, other.id as target_id, type(r) as rel_type
+        RETURN properties(r) as r, e.id as source_id, other.id as target_id, type(r) as rel_type
         LIMIT $limit
         """
 
--- a/tests/integration/test_neo4j_get_entity_relationships_integration.py
+++ b/tests/integration/test_neo4j_get_entity_relationships_integration.py
@@ -1,0 +1,142 @@
+"""Real-Neo4j integration test for ``Neo4jBackend.get_entity_relationships`` (DYT-2626).
+
+This test exercises the full Cypher → driver → ``result.data()`` →
+``_record_to_relationship`` serialization path against a running Neo4j
+instance. Unlike the mock-based unit tests in
+``tests/unit/test_neo4j_backend.py`` — which stub ``.data()`` directly and
+therefore can't detect a regression in the Cypher shape — this test pins
+the boundary that actually matters: the driver's on-wire representation
+of ``RETURN r`` (a ``neo4j.graph.Relationship`` that ``.data()``
+serializes as a 3-tuple) vs ``RETURN properties(r) as r`` (a plain dict
+that ``_record_to_relationship`` can index with string keys).
+
+Why this is marked ``@pytest.mark.integration`` and gated by
+``NEO4J_INTEGRATION_TEST=1``:
+
+    Khora's CI does NOT provision a Neo4j instance. The existing
+    "integration" tests in this repo are composition-level and still
+    mock-based (see ``tests/integration/test_dual_nodes_timeout_integration.py``
+    lines 23–33 for the full rationale). Real-Neo4j coverage lives
+    behind an opt-in env var so CI stays green while local developers
+    running ``make dev`` can exercise it.
+
+How to run locally:
+
+    make dev  # starts postgres + neo4j via docker compose
+    NEO4J_INTEGRATION_TEST=1 uv run pytest \
+        tests/integration/test_neo4j_get_entity_relationships_integration.py -v
+
+Connection parameters are read from env vars with sensible defaults that
+match the ``make dev`` compose stack:
+
+    KHORA_NEO4J_URL       (default: bolt://localhost:7687)
+    KHORA_NEO4J_USERNAME  (default: neo4j)
+    KHORA_NEO4J_PASSWORD  (default: password)
+
+The test would fail on the pre-DYT-2626 code because the real driver's
+``.data()`` method serializes a ``RETURN r`` value as a 3-tuple
+``(start_dict, rel_type, end_dict)`` — handing
+``_record_to_relationship`` a tuple instead of a dict and raising
+``TypeError: tuple indices must be integers or slices, not str``. The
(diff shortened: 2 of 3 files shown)


filippetrovic:
## 🔍 Multi-perspective review — PR #388 (DYT-2626)

Team: **Correctness Reviewer**, **Database Engineer**, **Devil's Advocate**. Security/Performance/Test Quality reviewers skipped — the change surface is too narrow (one-line Cypher + mock tests) to justify the overhead.

### Verdict: APPROVE with follow-ups

The one-line Cypher fix is **correct** and **minimally invasive**. Two follow-ups tracked; none block merge.

---

### HIGH

- **H1 [Correctness · Database]** `src/khora/storage/backends/neo4j.py:~1877` — `get_neighborhoods_batch` has the same class of driver-serialization bug: `collect(DISTINCT r)` + `await result.data()` yields a list of 3-tuples, which `_element_to_dict` silently converts to `{"_raw": "<repr>"}` stubs. This is **silent data loss** (every relationship property and type vanishes) — strictly worse than the visible `TypeError` this PR fixes, because callers don't know their neighborhoods are corrupted. Confirmed at the driver source level (neo4j-python 6.1.0, `neo4j/_data.py:324-329`: `x.__class__.__name__` is used, i.e. the literal string `"Relationship"`, NOT the Cypher rel type — minor correction to this PR's root-cause description, fix is still correct). **Filed as DYT-2629; does not block this PR** per the ticket's explicit scope ("strictly a semantic bug in one query").

### MEDIUM

- **M1 [Devil's Advocate · Test]** `tests/unit/test_neo4j_backend.py:~244-276` — the "regression-lock" test and the happy-path test both mock `result.data()` directly, so **neither exercises the real Cypher→driver path that the bug lives on**. A future dev regressing `properties(r) as r` → `RETURN r` would not be caught unless they also re-mock `data()` to return tuples. These tests pin `_record_to_relationship`'s dict-contract (useful) but do not lock the Cypher change itself. Acceptable for a unit-test layer given Khora's integration-test convention, but worth a follow-up integration test (testcontainers-neo4j) exercising `get_entity_relationships` end-to-end.
- **M2 [Correctness · Test]** `tests/unit/test_neo4j_backend.py` — edge cases not covered: empty-result path, `direction="incoming"` / `"both"`, `relationship_types` filter, records missing optional fields (`valid_from`, `valid_until`, `metadata`), multiple records. Consider expanding coverage in a follow-up; not blocking.
- **M3 [Process · Acceptance]** Ticket AC #2 ("Running the failing trace's namespace through recall locally no longer produces TypeError") was not executed. Given (a) the fix directly mirrors the working `list_relationships` pattern, (b) driver serialization behavior is confirmed at source level, and (c) the primary test exercises `_record_to_relationship` with the post-fix dict shape — this is low-risk. But it is a deviation from the ticket's written criteria; a quick post-merge smoke test against the failing trace's namespace is recommended.

### LOW

- **L1 [Devil's Advocate]** +97 lines of test scaffolding for a one-line Cypher fix is somewhat disproportionate. One function-level test would have sufficed; the class-level structure adds no shared fixtures. Not worth changing.
- **L2 [Database]** If/when DYT-2629 is fixed, the projection should preserve `type(rel)` (e.g. `[rel IN collect(DISTINCT r) | {props: properties(rel), type: type(rel)}]`) — naive `properties(rel)` alone drops relationship type info.

### No findings

- **Correctness of the fix itself** — `properties(r)` returns a Cypher map; `.data()` serializes maps to Python dicts; `_record_to_relationship` consumes dicts unchanged. Verified.
- **Query plan / indexes** — `Entity.id` is indexed (`neo4j.py:390`). `properties(r)` is a projection-time function, zero plan impact.
- **Alignment with sibling** — matches `list_relationships` (`neo4j.py:1541`) exactly.
- **Placeholders / TODOs** — none in the diff.

### Follow-ups filed

- [DYT-2629](https://linear.app/deyta/issue/DYT-2629) — Fix silent data loss in `Neo4jBackend.get_neighborhoods_batch`.

---

🤖 Review by `agent-ebd8d5`.

probablyrobot (approved):

PR #389, opened 2026-04-16
DYT-2625: Add KHORA_NEO4J_LOG_LEVEL env var for driver log verbosity

filippetrovic:
## Summary
- Add `KHORA_NEO4J_LOG_LEVEL` env var (DEBUG/INFO/WARNING/ERROR/CRITICAL, case-insensitive) to toggle the Python `neo4j` stdlib logger level at runtime — useful for diagnosing pool-acquire stalls, TLS handshake timing, routing refreshes (see IGR-56).
- New helper `apply_neo4j_log_level_from_env()` in `khora.logging_config` is called from both `setup_logging()` and `Neo4jBackend.__init__`, so the toggle works whether downstream services (peras, poros) use khora's logging helper or configure their own loguru sinks.
- Default (unset) is a no-op — no change to existing log output. Unknown values emit one loguru WARNING and leave the logger untouched (operator typos must not crash backend init).
- README documents the new env knob next to the other `KHORA_*` settings.
- `examples/neo4j_debug_logging.py` demonstrates the feature end-to-end — `uv run python examples/neo4j_debug_logging.py` produces ~15 driver DEBUG lines (`<POOL>`, `<WORKSPACE>`, `<RESOLVE>`, `<OPEN>`, …) before the expected connection failure against an absent local Neo4j.

## Test plan
- [x] `uv run pytest tests/unit/test_logging_config.py tests/unit/test_neo4j_backend.py` — 31 passed (new tests cover DEBUG/INFO/warning/mixed-case/empty/unknown/idempotent/default-unset and `Neo4jBackend.__init__` application).
- [x] Full suite: `uv run pytest --cov=src/khora --cov-branch --cov-fail-under=30` — 2782 passed, coverage 54.21%.
- [x] `uv run ruff check . && uv run ruff format --check . && uv run ty check src/` — clean (pre-existing ty warnings only; no new diagnostics).
- [x] Manual run of `examples/neo4j_debug_logging.py` with `KHORA_NEO4J_LOG_LEVEL=DEBUG` surfaces `neo4j` DEBUG lines through khora's loguru sink before the connect failure.

### DIFF
--- a/README.md
+++ b/README.md
@@ -167,6 +167,7 @@ All settings use the `KHORA_` prefix (e.g., `KHORA_LLM_MODEL=gpt-4o`).
 | `KHORA_QUERY_TEMPORAL_SQL_PUSHDOWN` | Temporal SQL WHERE pushdown | `true` |
 | `KHORA_EXTRACTION_BATCH_SIZE` | Chunks per extraction LLM call | `5` |
 | `KHORA_DEBUG` | Debug logging | `false` |
+| `KHORA_NEO4J_LOG_LEVEL` | Neo4j driver log level (`DEBUG`/`INFO`/`WARNING`/`ERROR`/`CRITICAL`, case-insensitive). No-op when unset | — |
 
 See [full configuration reference](docs/architecture/storage-backends.md) for all options.
 
--- a/examples/neo4j_debug_logging.py
+++ b/examples/neo4j_debug_logging.py
@@ -1,0 +1,50 @@
+"""Demonstrate KHORA_NEO4J_LOG_LEVEL=DEBUG surfacing driver-internal logs.
+
+Run with::
+
+    uv run python examples/neo4j_debug_logging.py
+
+The script sets ``KHORA_NEO4J_LOG_LEVEL=DEBUG`` in-process, initializes
+khora's logging, and then constructs a ``Neo4jBackend`` pointing at the
+local compose stack. The driver emits DEBUG lines as soon as a connection
+is attempted (TLS handshake, pool bootstrap, routing). Those lines are
+routed through khora's ``InterceptHandler`` into loguru and printed to
+stdout.
+
+If no Neo4j is reachable the connect call raises — that's fine; the DEBUG
+lines produced before the failure are the point of the demo.
+"""
+
+from __future__ import annotations
+
+import asyncio
+import os
+
+# Must be set BEFORE setup_logging() / Neo4jBackend() so the helper sees it.
+os.environ.setdefault("KHORA_NEO4J_LOG_LEVEL", "DEBUG")
+
+from khora.logging_config import setup_logging  # noqa: E402
+from khora.storage.backends.neo4j import Neo4jBackend  # noqa: E402
+
+
+async def main() -> None:
+    setup_logging(level="DEBUG")
+
+    backend = Neo4jBackend(
+        url=os.environ.get("KHORA_NEO4J_URL", "bolt://localhost:7688"),
+        user=os.environ.get("KHORA_NEO4J_USER", "neo4j"),
+        password=os.environ.get("KHORA_NEO4J_PASSWORD", "pleaseletmein"),  # noqa: S106
+    )
+
+    try:
+        await backend.connect()
+        print("Connected to Neo4j; driver DEBUG logs should appear above.")
+    except Exception as exc:  # noqa: BLE001 - demo script, surface anything
+        print(f"Connect failed ({type(exc).__name__}: {exc}).")
+        print("Any 'neo4j' DEBUG lines above were produced before the failure.")
+    finally:
+        await backend.disconnect()
+
+
+if __name__ == "__main__":
+    asyncio.run(main())
--- a/src/khora/logging_config.py
+++ b/src/khora/logging_config.py
@@ -4,6 +4,7 @@ from __future__ import annotations
 
 import atexit
 import logging
+import os
 import sys
 from pathlib import Path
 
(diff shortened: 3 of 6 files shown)


probablyrobot (approved):

filippetrovic:
## Multi-perspective review (self-review)

**Overall verdict: APPROVE — no blockers. All four acceptance criteria satisfied. A handful of NITPICK-level notes worth considering but none are required before merge.**

### Acceptance criteria traceability
| AC | Status | Evidence |
|----|--------|----------|
| Setting `KHORA_NEO4J_LOG_LEVEL=DEBUG` produces driver-internal log lines without code change | ✅ | `apply_neo4j_log_level_from_env()` called from both `setup_logging()` and `Neo4jBackend.__init__`; example script confirmed ~15 `<POOL>/<RESOLVE>/<OPEN>/<CONNECTION>` DEBUG lines through loguru sink |
| Default unset produces same output as today — no warning spam | ✅ | `if not raw: return` short-circuits on unset/empty. Explicit tests: `test_apply_neo4j_log_level_from_env_noop_when_unset`, `test_apply_neo4j_log_level_from_env_noop_when_empty`, `test_setup_logging_does_not_touch_neo4j_logger_when_env_unset`, `TestNeo4jBackendLogLevelFromEnv.test_init_noop_when_env_unset` |
| Documented in README | ✅ | Row added in the `KHORA_*` config table next to `KHORA_DEBUG` |
| Works with khora's `setup_logging()` AND downstream services with their own sinks | ✅ | Two integration points; the `Neo4jBackend.__init__` path covers peras/poros that don't call `setup_logging` but configure their own logfire/loguru sinks |

### Architect
- [PRAISE] Helper lives in `logging_config.py` alongside related logger wiring — correct module. Single public entry point, easy to reason about.
- [PRAISE] Call site in `setup_logging()` is placed AFTER `InterceptHandler` + `basicConfig(force=True)` + third-party suppression loop, so the new level actually reaches loguru sinks. Comment explains the ordering.
- [NITPICK] The lazy import in `Neo4jBackend.__init__` (`from khora.logging_config import apply_neo4j_log_level_from_env`) is defensive — `logging_config` has no khora-internal imports today, so a module-level import would also work. The lazy form is fine and costs nothing; just flagging that the "avoid cycles" justification in the comment is slightly aspirational.

### Devil's Advocate
- [MINOR] Operator typo (e.g. `KHORA_NEO4J_LOG_LEVEL=DEBUGG`) produces one `logger.warning` per `Neo4jBackend()` construction. In an app that instantiates many backends across its lifecycle this could repeat. Not harmful (it's a WARNING, not an error), but a "warn once per process" cache would be trivial if we want to silence it. Deferring since invalid values are operator-driven and rare.
- [NITPICK] Whitespace-only env value (`"   "`) goes through the "unknown value" branch → WARNING. Arguably should be treated as unset (silent no-op). Current behavior is defensible — it surfaces a probable shell-quoting mistake.
- [NITPICK] Numeric level strings (e.g. `"10"`) are rejected as unknown. Python's stdlib `Logger.setLevel` accepts ints, but the ticket explicitly enumerates the five name strings, so this rejection is correct per spec.
- [PRAISE] Precedence choice is right: when the env var is set, it overrides whatever `setup_logging(level=...)` would do for the neo4j logger. That matches the "operator toggle" intent.

### Correctness
- [PRAISE] `raw = os.environ.get(...)` + `if not raw: return` is clean and correctly handles both unset (`None`) and empty string in one branch.
- [PRAISE] `raw.strip().upper()` gives case-insensitive + leading/trailing whitespace tolerance without over-engineering.
- [PRAISE] `_NEO4J_LOG_LEVELS` is a module-level `set` → constant-time membership check and easy to extend.
- [PRAISE] `setLevel(level)` is passed the uppercase string name, which stdlib `logging` accepts natively since 3.2 — no `getattr(logging, level)` juggling needed.

### Tests
- [PRAISE] Logger state is reset via a proper pytest fixture that captures the original level and restores it in teardown — no leakage across the suite.
- [PRAISE] Parametrized case-insensitivity coverage (DEBUG/info/Warning/ERROR/critical).
- [PRAISE] Explicit coverage of: unset, empty, unknown-value, idempotency, `setup_logging()` integration, `Neo4jBackend.__init__` application, `Neo4jBackend.__init__` no-op-when-unset.
- [NITPICK] No test for whitespace-only env value; not a bug, just an unpinned behavior.
- [NITPICK] No test for numeric-string values (`"10"`); same — current rejection is intentional but not locked by a test.

### Security
- [PRAISE] No injection vector — env value is compared to a whitelist before reaching stdlib `logging`. Unknown values never touch `setLevel`.
- [PRAISE] Hardcoded `pleaseletmein` in `examples/neo4j_debug_logging.py` is the local `compose.yaml` default, explicitly annotated with `noqa: S106` and fronted by `os.environ.get(...)` so real deployments can override. Acceptable for an example script.
- [NITPICK] Worth a one-liner in the README or example docstring that driver DEBUG logs may include connection URL fragments and routing metadata — operators should not leave `KHORA_NEO4J_LOG_LEVEL=DEBUG` on indefinitely in prod logs. Not blocking.

### Performance
- [PRAISE] O(1) string check + one `setLevel` call. Ignorable cost at every `Neo4jBackend.__init__`. No runtime impact when unset.

### Docs
- [PRAISE] README row placement and phrasing match the neighboring `KHORA_*` rows.
- [NITPICK] The ticket mentions the downstream mapping (`PERAS_KHORA__NEO4J_LOG_LEVEL`, `POROS_KHORA__NEO4J_LOG_LEVEL`) — not khora's concern to document, but a one-line README hint ("downstream services expose this as `<SERVICE>_KHORA__NEO4J_LOG_LEVEL`") would save operators a grep. Defer.

### Summary of findings
- 0 BLOCKER / 0 MAJOR / 1 MINOR / 5 NITPICK / 8 PRAISE
- The single MINOR (warning spam on repeated construction with invalid value) is operator-driven and low-impact. Not worth fixing in this PR.
- CI green on install / lint / security / test. Recommend merge.

PR #406, opened 2026-04-17
DYT-2671: Filter relationship valid_until in get_entity_neighborhoods

filippetrovic:
## Summary
- Extend `get_entity_neighborhoods()` in `dual_nodes.py` to filter expired relationships when `prefer_current=True`, matching the existing entity filtering behavior
- Adds `all(r IN relationships(path) WHERE r.valid_until IS NULL OR r.valid_until > datetime())` to the Cypher traversal query, gated behind the same `prefer_current` flag
- Part of ADR-056: Khora Document Update Lifecycle (Phase 2), decision #6 — recall-side complement to relationship retirement

## Test plan
- [x] Unit tests pass (2516 passed, 0 failed)
- [x] Pre-commit hooks pass (ruff, ruff-format, ty, trailing whitespace)
- [ ] Integration tests: `NEO4J_INTEGRATION_TEST=1 uv run pytest tests/integration/test_neo4j_neighborhood_valid_until_integration.py -v`
  - Future `valid_until` → relationship surfaces with `prefer_current=True`
  - Past `valid_until` → relationship excluded with `prefer_current=True`
  - Past `valid_until` → relationship surfaces with `prefer_current=False`
  - NULL `valid_until` → always included (NULL = still valid)

Linear: https://linear.app/deyta/issue/DYT-2671

### DIFF
--- a/src/khora/engines/vectorcypher/dual_nodes.py
+++ b/src/khora/engines/vectorcypher/dual_nodes.py
@@ -586,13 +586,17 @@ class DualNodeManager:
 
         depth = min(max(1, depth), 4)  # Clamp to 1-4
 
-        # When prefer_current is set, exclude entities whose validity has expired.
-        # Entities with NULL valid_until are kept (NULL = no known end = still valid).
+        # When prefer_current is set, exclude entities and relationships whose
+        # validity has expired. NULL valid_until is kept (no known end = still valid).
+        # Hoist datetime() into a WITH clause so it is evaluated once per row,
+        # not once per relationship in the all() predicate.
+        temporal_preamble = ""
         temporal_clause = ""
         if prefer_current:
+            temporal_preamble = "WITH e, datetime() AS _now"
             temporal_clause = (
-                "AND (related.valid_until IS NULL OR related.valid_until > datetime())"
-                "\n          AND all(r IN relationships(path) WHERE r.valid_until IS NULL OR r.valid_until > datetime())"
+                "AND (related.valid_until IS NULL OR related.valid_until > _now)"
+                "\n          AND all(r IN relationships(path) WHERE r.valid_until IS NULL OR r.valid_until > _now)"
             )
 
         # OPTIMIZATION: Single query fetches all neighborhoods in batch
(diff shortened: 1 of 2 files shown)


probablyrobot (approved):

filippetrovic:
## Consolidated Review — PR #406

**Verdict: APPROVE** — no blockers. Code is correct, safe, and well-tested.

### MEDIUM

- **M1 [Performance]** `dual_nodes.py:595` — `datetime()` is called per-relationship in the `all()` predicate. Could be hoisted to `WITH datetime() AS now` before `OPTIONAL MATCH` to reduce per-eval overhead. Low urgency; Neo4j typically caches this within a transaction.

- **M2 [Test]** `test_neo4j_neighborhood_valid_until_integration.py` — All 4 tests use `depth=1`. A depth=2 test with an expired intermediate relationship (A→B→C where B→C is expired) would strengthen coverage of the `all()` predicate across multi-hop paths.

- **M3 [Test]** No multi-relationship scenario tested. A test with partial expiry (one rel expired, another current between same entities) would verify selective filtering.

### LOW

- **L1 [Security]** `dual_nodes.py:595` — Cypher predicate is interpolated via f-string. Safe because it's hardcoded (no user input flows in), but noting for audit trail.

- **L2 [Correctness]** `dual_nodes.py:589` — Inline comment at line 589 says "exclude entities whose validity has expired" but now also excludes relationships. Minor doc drift.

- **L3 [Database]** Index usage unaffected — temporal filter in WHERE clause is pushed down after indexed MATCH. No query plan concerns at depth ≤ 4.

### Dismissed

- **~~H1 [Correctness]~~** `relationships(NULL)` vacuous truth — flagged by correctness reviewer but confirmed non-issue. When `OPTIONAL MATCH` returns no path, `related` is NULL, and downstream `CASE WHEN related IS NOT NULL` + `[x IN related_raw WHERE x IS NOT NULL]` (lines 608-627) already filter out phantom neighbors. No fix needed.

### Summary

Production change is a single, well-scoped Cypher predicate addition. Integration tests cover the 3 acceptance criteria + NULL case. Existing entity filtering (pre-PR) is untouched. PR is already approved by @probablyrobot. No blocking issues found.

PR #426, opened 2026-04-22
DYT-2675: Integration tests for replace_document_extraction

filippetrovic:
## Summary

Closes DYT-2675. Adds `tests/integration/test_replace_document_extraction.py` covering the ADR-056 Phase 2 document-replacement lifecycle through the `MemoryLake.remember()` public API surface (complements the coordinator-level sibling `test_coordinator_replace_document_extraction.py`).

- Golden path: re-ingesting the same `external_id` retires orphaned entities/relationships and remaps survivors.
- Co-sourced entity survives retirement.
- Co-sourced relationship survives retirement.
- Atomic chunk replacement: a failure inside the coordinator's PG transaction rolls back chunks and leaves the graph untouched.
- Graph-side failure marks the document `FAILED`; the next `remember()` on the same `external_id` self-heals to `COMPLETED` (ADR-056 §Decision 8).
- Concurrent `remember()` calls with the same `external_id` converge on a single document row — accepts both the success/success and success/IntegrityError outcomes the race permits (ADR-056 §Decision 9).
- `prefer_current` filter symmetry for retired edges, anchored to Cypher that mirrors `DualNodeManager.get_entity_neighborhoods` (dual_nodes.py:595-600).
- Backward-compat: `external_id=None` still hits the checksum dedup path without routing through `replace_document_extraction`.

No production code touched. Extraction and embedding are stubbed via a registry-based `monkeypatch` on `LLMEntityExtractor.extract_multi` and `LiteLLMEmbedder.embed_batch` so tests don't require OpenAI. The whole file is gated by `NEO4J_INTEGRATION_TEST=1`, so CI (which does not provision Neo4j) auto-skips it — same gate as the sibling file.

## Test plan

- [x] `uv run ruff check .` clean
- [x] `uv run ruff format --check .` clean
- [x] `uv run pytest --cov --cov-fail-under=30` — 2550 passed / 174 skipped (the 8 new tests skip without the gate); coverage 52.5% (pre-existing collection error in `tests/unit/storage/backends/test_protocol_compliance.py` and 3 pre-existing logfire-related failures are present on `origin/main` too — not introduced here)
- [x] `pytest tests/integration/test_replace_document_extraction.py --collect-only` collects 8 tests
- [ ] Manual: `make dev && NEO4J_INTEGRATION_TEST=1 uv run pytest tests/integration/test_replace_document_extraction.py -v` on a workstation with Docker (not runnable in this sandbox)

### DIFF
--- a/tests/integration/test_replace_document_extraction.py
+++ b/tests/integration/test_replace_document_extraction.py
@@ -376,6 +376,10 @@ class TestReplaceViaRememberIntegration:
         assert bob_supersedes[0]["reason"] == "document_replaced"
         assert bob_supersedes[0]["current_valid_until"] is not None
         assert bob_supersedes[0]["superseded_at"] is not None
+        # SUPERSEDES edge must point at the EntityVersion snapshot — not a
+        # different (or missing) row. Guards against a producer bug that
+        # creates one without the other.
+        assert bob_supersedes[0]["old_id"] == bob_versions[0]["id"]
 
         # alice survives and has been remapped off the old document id.
         # v1 and v2 share the same document id (in-place replace), so the
--- a/tests/integration/test_replace_document_extraction.py
+++ b/tests/integration/test_replace_document_extraction.py
@@ -1,0 +1,931 @@
+"""End-to-end replace-via-remember() integration tests (DYT-2675, ADR-056).
+
+Exercises the document-replacement lifecycle through the public
+``MemoryLake.remember()`` / ``VectorCypherEngine.remember()`` API surface,
+against a real Postgres + Neo4j stack. The sibling file
+``test_coordinator_replace_document_extraction.py`` tests the coordinator
+primitive directly; this file tests the public entry point where an
+``external_id`` collision dispatches to ``_remember_via_replace``.
+
+Scenarios (one test method each):
+
+1. Golden path: same ``external_id`` replaces in-place (orphan retire,
+   survivor remap, net-new upsert, relationship retire).
+2. Co-sourced entity survives retirement when one of two source docs is
+   replaced.
+3. Co-sourced relationship survives retirement likewise.
+4. Atomic chunk replacement: Postgres failure rolls back the chunk write
+   and graph mutations do not leak through.
+5. Graph-side failure marks the document ``FAILED``; the next successful
+   replace self-heals it to ``COMPLETED`` (ADR §Decision #8).
+6. Concurrent ``remember()`` calls with the same ``external_id`` converge
+   on a single document (engine.py:641 IntegrityError -> _remember_via_replace).
+7. ``prefer_current`` filter symmetry — retired edges are excluded when
+   the traversal applies the ``valid_until`` predicate and visible
+   otherwise. Anchored to raw Cypher that mirrors
+   ``DualNodeManager.get_entity_neighborhoods`` (dual_nodes.py:595-600)
+   because neither ``MemoryLake.recall()`` nor
+   ``GraphBackend.get_neighborhoods_batch`` expose ``prefer_current``
+   publicly (ADR-056 §Decision 6, plan open question #1).
+8. Backward compat: when ``external_id=None``, checksum dedup still
+   returns ``duplicate=True`` without routing through the replace path.
+
+Gated by ``NEO4J_INTEGRATION_TEST=1`` because CI does not provision Neo4j.
+
+How to run locally::
+
+    make dev  # postgres + neo4j via docker compose
+    NEO4J_INTEGRATION_TEST=1 uv run pytest \\
+        tests/integration/test_replace_document_extraction.py -v
+
+Connection parameters (env overrides, sensible ``make dev`` defaults)::
+
+    KHORA_NEO4J_URL          (default: bolt://localhost:7687)
+    KHORA_NEO4J_USERNAME     (default: neo4j)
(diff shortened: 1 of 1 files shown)


filippetrovic:
## Team Review Summary — PR #426 (DYT-2675)

Multi-perspective review by 8 reviewers (Security, Performance, Correctness, Test, Docs, Database, Devil's Advocate, ADR-Drift) + dedicated security audit. File scope: single new `tests/integration/test_replace_document_extraction.py` (931 lines, 8 scenarios, no production code touched).

### HIGH

- **H1 · [ADR Drift / CI]** `.github/workflows/ci.yml` — CI does not run these tests. No workflow sets `NEO4J_INTEGRATION_TEST=1` and no job provisions Postgres + Neo4j services. The ticket's acceptance criterion "Tests run in CI as part of `make test`" is technically unmet. This matches the sibling `test_coordinator_replace_document_extraction.py` gate pattern, so it's consistent with existing practice, but it means regressions in the replace lifecycle will not be caught by CI. **Recommended:** open a follow-up ticket to wire an `integration-with-db` workflow that provisions the stack and exports the gate.
- **H2 · [Correctness / S1]** test file line ~370-376 — The golden-path test verifies the `:EntityVersion` snapshot and the `SUPERSEDES` edge separately, but doesn't cross-check that the edge's `old.id` matches the snapshot's `id`. A bug creating edges without snapshots (or mismatched pairs) would pass. **Fix:** have `_get_supersedes_edges` return `old.id` and assert it equals `bob_versions[0]["id"]`.

### MEDIUM

- **M1 · [Correctness / S4]** line ~602 — `assert call_count["n"] >= 1` only proves `create_chunks_batch` was called; it doesn't prove the failure happened inside the PG txn vs. before the pre-coordinator chunk wipe (engine.py:1391-1405). Lines 569-572 already note the v1 create writes to `khora_chunks` not `chunks`, so `get_chunks_by_document(doc_id)` returns 0 before AND after — the "unchanged" assertion is tautological. **Fix:** assert on a column that the coordinator actually writes (e.g., `chunk_count` on the document row) so rollback is observable.
- **M2 · [Correctness / S5]** line ~693 — Asserts `error_message is None` after heal, but not that the error message is populated while the document is in `FAILED` state. **Fix:** before the heal call, add `assert "injected graph failure" in failed.error_message`.
- **M3 · [Devil's Advocate / S6]** line 739 — `asyncio.gather` on one event loop with one pool interleaves at `await` points only. Single-loop cooperative scheduling frequently lets task A complete `_remember_via_replace` before task B reaches its INSERT. Accepting "both succeeded" as pass (line 759) means the test can pass without ever hitting the `IntegrityError` branch at engine.py:641. The UNIQUE-backstop contract ADR-056 §Decision 9 is therefore under-tested. **Options:** (a) accept this as in-process ceiling and document the limitation; (b) use `multiprocessing`/threads to force real concurrency.
- **M4 · [Devil's Advocate / S7]** lines 833-868 — Test hand-authors Cypher that explicitly mirrors `dual_nodes.py:595-600`. If production Cypher drifts, the test doesn't catch it because test and prod share the same author. **Options:** (a) call `DualNodeManager.get_entity_neighborhoods` directly (private reach via `engine._storage` etc.); (b) drop S7 and note recall-surface symmetry is unreachable through the public API.
- **M5 · [Docs]** module docstring (lines 1-51) — Cites ADR-056 §Decisions 6/8/9, but ADR-056 lives in `DeytaHQ/hegemon`, not this repo. Readers without access to that repo can't verify cited decision numbers. **Fix:** add a one-line pointer in the module docstring.

### LOW

- **L1 · [Performance / S7]** lines 852-866 — `OPTIONAL MATCH path = (e)-[*1..1]-(related:Entity)` is equivalent to `(e)-[r]-(related:Entity)` for a 1-hop traversal; variable-length adds unnecessary overhead. Negligible at test runtime.
- **L2 · [Test Coverage]** — `remember_batch()`, partial-chunk failures mid-ingest, and multi-namespace entity-name collisions aren't covered. All arguably below-ticket-scope; flag as future hardening.
- **L3 · [Style]** — 931 lines for 8 tests has extractable duplication (every test re-derives `ns = str(await lake.storage.resolve_namespace(namespace_id))`, opens a driver handle). Factor into a small fixture. Not blocking.

### Security Audit

- **PASS.** No CRITICAL or HIGH findings. Credentials at lines 265, 272 are local-docker-compose defaults (`password`, matching the sibling test). All Cypher queries are parameterized via `driver.session().run(query, **params)` — no string interpolation of user data. No dependency changes (`pyproject.toml`/`uv.lock` untouched). No `exec`/`eval`/`subprocess`/file writes.

### Overall Assessment

**REQUEST_CHANGES (soft)** — No outright blockers. H2, M1, and M2 are cheap tightening opportunities that should land with this PR. H1 (CI integration) is an acceptance-criterion gap that matches existing practice and is best resolved by a follow-up ticket rather than expanding this PR's scope. M3/M4 are architectural limitations of testing at the API layer — worth a brief note in the PR body rather than rework.

**Strengths:** test isolation via UUID-suffixed names + per-test namespaces, clean `monkeypatch` usage throughout, accurate line references to engine.py/dual_nodes.py, invariant-focused assertions with explicit co-occurrence defenses (relationship-type filter in `_get_relationship`), and parity with the sibling file's style.

Arsenije (approved):

PR #459, opened 2026-04-27
DYT-3191: Replace O(N) semaphore tasks with bounded worker pool in submit_batch

msrdic:
## Summary
- Replaced the `asyncio.Semaphore` + `O(N)` task pattern in `submit_batch` with a bounded worker pool using `asyncio.Queue`
- Instead of creating one task per document (all gated by a semaphore), we now create `min(max_concurrent, N)` worker tasks that pull from a shared queue
- This reduces memory overhead and task-scheduling pressure for large batches while preserving the same concurrency limit

## Test plan
- [x] All 2723 unit tests pass
- [x] Coverage at 52.74% (above 46% threshold)
- [ ] Integration tests with large batch sizes to verify throughput parity

### DIFF
--- a/src/khora/memory_lake.py
+++ b/src/khora/memory_lake.py
@@ -1106,12 +1106,19 @@ class MemoryLake:
             handle._mark_done()
             return
 
-        sem = asyncio.Semaphore(max_concurrent)
+        queue: asyncio.Queue[tuple[Document, dict[str, Any]]] = asyncio.Queue()
+        for doc, data in zip(pending_docs, doc_data_list):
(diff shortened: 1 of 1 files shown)


msrdic:
**Audits** — Clean on first pass, 1 LOW noted (1 cycle)

Self-audit, Bandit security scan, and Codex audit all returned no HIGH or MEDIUM findings.

Noted (advisory):
- 🟢 L1: Non-positive `max_concurrent` silent failure — noted (pre-existing edge case on internal API, not a regression)

PR #582, opened 2026-05-14
DYT-4164: Cascade forget() to drop orphan entities and relationships

filippetrovic:
## Summary

Fixes the `forget()` orphan-cascade bug ([DYT-4164](https://linear.app/deyta/issue/DYT-4164/forget-leaks-orphan-entities-and-relationships-into-entity-search), user-facing report [IGR-202](https://linear.app/deyta/issue/IGR-202/ask-sometimes-returns-forgotten-memories-janjetina)).

Before this change, `khora.forget(document_id, namespace)` deleted the document row and its chunks but never touched the entities/relationships extracted from that document. Both the pgvector entity/relationship tables and the Neo4j `:Entity` nodes + edges retained the orphans, so `entity_search` (pgvector) and `entity_explore` (Neo4j traversal) still returned them.

### Changes

- **`src/khora/storage/backends/neo4j.py`** — adds four batch helpers:
  - `delete_entities_batch(entity_ids, namespace_id)` — `DETACH DELETE` on `:Entity`.
  - `delete_relationships_batch(relationship_ids)` — deletes edges by their `id` property (uses `()-[r]-()` + `WITH DISTINCT` for self-loop safety).
  - `remove_document_from_entity_sources_batch(entity_ids, doc_id, namespace_id)` — strips `doc_id` from `source_document_ids` (and `source_chunk_ids` consistency strip).
  - `remove_document_from_relationship_sources_batch(relationship_ids, doc_id)` — analog for edges.
- **`src/khora/storage/backends/pgvector.py`** — adds four batch helpers using SQLAlchemy `delete()`/`update()` with `func.array_remove`:
  - `delete_entities_batch`, `delete_relationships_batch`
  - `remove_document_from_entity_sources`, `remove_document_from_relationship_sources`
- **`src/khora/engines/vectorcypher/engine.py:forget`** — new `_cascade_forget_extraction(doc_id, ns_id)` helper invoked **before** chunk/doc deletes. Partitions entities/relationships on `source_document_count == 1` (orphans, hard-deleted) vs `> 1` (survivors, doc stripped from their source arrays).
- **`src/khora/engines/chronicle/engine.py:forget`** — same cascade, gated on `getattr(graph, "fetch_document_extraction_state", None) is None` so chronicle pgvector-only deployments remain a clean no-op.

### Design choices

- **Hard delete, not soft retire** — matches the user-facing `forget` verb (GDPR-style removal), matches the v0.2.3 CHANGELOG's stated direction, and avoids having to patch every read path to filter `valid_until`. Soft-retire machinery is intentionally left in place for `replace_document_extraction` where the audit trail rationale applies. `retire_orphaned_entities_batch` is NOT reused.
- **Empty-input short-circuit** — all 8 helpers return `0` without issuing a DB call when given an empty id list.
- **No public API or telemetry contract changes** — no `__all__` additions, no `docs/telemetry-contract.json` bump needed.
- **No read-path changes** — `search_similar_entities`, `get_entity_neighborhoods`, `_vector_search_entities` are untouched (explicitly out of scope per the ticket).

### Known non-blocking trade-offs (documented for future work)

- Cross-store atomicity: graph delete → vector delete is not transactional across stores. If vector delete fails, the orphan persists in pgvector but is unreachable (no Neo4j reference). Same model as the existing `replace_document_extraction`.
- Concurrency: no document-level lock during the cascade — a parallel `remember()` on the same entity could race with the partition logic. Low-probability for production workloads; flagged for future operational notes.
- The four `graph.*_batch` call sites in the engines use `# type: ignore[unresolved-attribute]` (matches the existing pattern in `coordinator.replace_document_extraction`). Promoting the four helpers to `GraphBackendProtocol` is a separate cleanup ticket.

## Test plan

- [x] Unit tests pass — 17 new tests in `tests/unit/engines/test_vectorcypher_engine.py` and `tests/unit/engines/test_chronicle_engine.py` covering: orphan-entity hard-delete, survivor source-list strip, orphan/survivor relationship paths, zero-extraction no-op, cascade-before-chunk-delete ordering, chronicle pgvector-only fallback, namespace-mismatch early return. Verified: `uv run pytest tests/unit/engines/test_vectorcypher_engine.py tests/unit/engines/test_chronicle_engine.py -q --no-cov` → 78 passed.
- [x] `uv run ruff check .`, `uv run ruff format --check .`, `uv run ty check src/` — all exit 0.
- [ ] Integration test: `tests/integration/test_forget_cascade_orphans_integration.py` — reproduces IGR-202 against real Postgres + Neo4j. Gated by `NEO4J_INTEGRATION_TEST=1` (matches `tests/integration/test_replace_document_extraction.py` convention). Will run in CI; locally requires `make dev` first.
- [ ] CI green.

### DIFF
--- a/src/khora/engines/chronicle/engine.py
+++ b/src/khora/engines/chronicle/engine.py
@@ -2122,13 +2122,67 @@ class ChronicleEngine:
         storage = self._get_storage()
 
         # Verify namespace if provided
-        if namespace_id:
+        ns_id = namespace_id
+        if ns_id:
             document = await storage.get_document(document_id)
-            if document and document.namespace_id != namespace_id:
-                logger.warning(f"Document {document_id} not in namespace {namespace_id}")
+            if document and document.namespace_id != ns_id:
+                logger.warning(f"Document {document_id} not in namespace {ns_id}")
                 return False
+        else:
+            document = await storage.get_document(document_id)
+            if document is None:
+                return False
+            ns_id = document.namespace_id
+
+        await self._cascade_forget_extraction(document_id, ns_id)
 
         return await storage.delete_document(document_id)
+
+    async def _cascade_forget_extraction(self, document_id: UUID, namespace_id: UUID) -> None:
+        """Drop / decrement entities and relationships extracted from a document.
+
+        Hard-deletes orphans (single-source entities/relationships whose only
+        ``source_document_ids`` entry is ``document_id``) and strips
+        ``document_id`` from survivors' ``source_document_ids`` arrays in
+        both the graph and the vector backends. No-op when the graph backend
+        does not expose ``fetch_document_extraction_state`` (e.g. chronicle
+        pgvector-only deployments where the graph backend is absent).
+        """
+        storage = self._get_storage()
+        graph = storage.graph
+        vector = storage.vector
+        fetch = getattr(graph, "fetch_document_extraction_state", None)
+        if fetch is None:
+            return
+
+        entities, relationships = await fetch(document_id, namespace_id=namespace_id)
+
+        orphan_ent_ids = [UUID(e["id"]) for e in entities if e["source_document_count"] == 1]
+        survive_ent_ids = [UUID(e["id"]) for e in entities if e["source_document_count"] > 1]
+        orphan_rel_ids = [UUID(r["id"]) for r in relationships if r["source_document_count"] == 1]
+        survive_rel_ids = [UUID(r["id"]) for r in relationships if r["source_document_count"] > 1]
+
+        if orphan_ent_ids:
+            await graph.delete_entities_batch(orphan_ent_ids, namespace_id)  # type: ignore[unresolved-attribute]
+            if vector is not None and hasattr(vector, "delete_entities_batch"):
+                await vector.delete_entities_batch(orphan_ent_ids)
+        if orphan_rel_ids:
+            await graph.delete_relationships_batch(orphan_rel_ids)  # type: ignore[unresolved-attribute]
+            if vector is not None and hasattr(vector, "delete_relationships_batch"):
+                await vector.delete_relationships_batch(orphan_rel_ids)
+
+        if survive_ent_ids:
+            await graph.remove_document_from_entity_sources_batch(  # type: ignore[unresolved-attribute]
+                survive_ent_ids, document_id, namespace_id
+            )
+            if vector is not None and hasattr(vector, "remove_document_from_entity_sources"):
+                await vector.remove_document_from_entity_sources(survive_ent_ids, document_id)
+        if survive_rel_ids:
+            await graph.remove_document_from_relationship_sources_batch(  # type: ignore[unresolved-attribute]
+                survive_rel_ids, document_id
+            )
+            if vector is not None and hasattr(vector, "remove_document_from_relationship_sources"):
+                await vector.remove_document_from_relationship_sources(survive_rel_ids, document_id)
 
     @trace("khora.chronicle.remember_batch")
     async def remember_batch(
--- a/src/khora/engines/vectorcypher/engine.py
+++ b/src/khora/engines/vectorcypher/engine.py
@@ -1838,6 +1838,8 @@ class VectorCypherEngine:
             else:
                 return False
 
+        await self._cascade_forget_extraction(document_id, ns_id)
+
         # Delete from Neo4j (Chunk nodes and relationships)
         if dual_nodes is not None:
             await dual_nodes.delete_chunks_by_document(document_id, ns_id)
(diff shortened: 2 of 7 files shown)


filippetrovic:
# Multi-perspective Review — PR #582 (DYT-4164)

Synthesized from 7 parallel reviewers: security, performance, correctness, test, docs, database, devil's-advocate.

## Overall assessment: **APPROVE with documentation follow-up**

Implementation is sound. No CRITICAL or genuinely blocking HIGH findings after triage. Two HIGH findings from the database reviewer were verified against the codebase and reclassified (see notes inline). Recommend merging once the docs gaps are filled (a 1-2 minute change, can be a follow-up commit on this PR or a tiny follow-up PR).

---

## Findings

### HIGH

_None after triage._ Two HIGH database findings were verified false/over-stated:

- **(reclassified)** `array_remove(column, document_id)` with UUID arg — claim of silent no-op due to ARRAY(String) column. **Verified false**: `db/models.py:284, 351` declare `ARRAY(UUID(as_uuid=True))`. Per CLAUDE.md UUID handling rules, passing a `uuid.UUID` directly is correct.
- **(reclassified to L1 below)** Missing `namespace_id` on `remove_document_from_relationship_sources_batch`. UUID PKs are globally unique and the IDs come from `fetch_document_extraction_state(doc_id, namespace_id)` which is already namespace-scoped, so this is style-only defense-in-depth, not a cross-tenant bug.

### MEDIUM

- **M1 — [Docs] `forget()` semantics change not documented.** README, `docs/api-reference.md`, and the `Khora.forget()` docstring in `src/khora/khora.py` don't mention the new cascade behavior. CHANGELOG has no entry. Users still rely on the old (orphan-preserving) contract may be surprised. _Reviewer: docs-reviewer._
- **M2 — [Correctness] Vector-backend hasattr() inconsistency.** `engines/{chronicle,vectorcypher}/engine.py` gates **graph** calls with an early-return when `fetch_document_extraction_state` is missing, but gates **vector** calls per-helper with `hasattr(vector, ...)`. If a vector backend implemented some but not all four helpers, the cascade would silently skip the missing ones, producing partial cleanup. Today this is hypothetical (pgvector implements all four), but it's worth either an upfront capability check or a Protocol commitment. _Reviewer: correctness-reviewer #3._
- **M3 — [Architecture] Cross-store atomicity is not transactional.** The cascade does `graph.delete → vector.delete → graph.strip → vector.strip → chunk delete → doc delete` without wrapping in `coordinator.transaction()`. A failure between the graph delete and vector delete leaves the orphan in pgvector (logically unreachable from the graph, but still scannable by `search_similar_entities`). The PR description acknowledges this and notes it matches `replace_document_extraction`'s model — surfacing here for explicit operator awareness rather than blocking the merge. _Reviewers: security-reviewer #1, devils-advocate #3, correctness-reviewer #4._

### LOW

- **L1 — [Style] Asymmetric namespace scoping.** `delete_entities_batch` and `remove_document_from_entity_sources_batch` take `namespace_id`; the relationship analogs do not. Globally-unique UUID PKs make this safe today, but symmetry would be a useful defense-in-depth (and would prevent future regressions if relationship IDs ever become non-unique across namespaces). _Reviewer: database-reviewer._
- **L2 — [Type safety] `# type: ignore[unresolved-attribute]` × 4 in each engine.** The four new graph helpers aren't on `GraphBackendProtocol`, so the engines suppress type checking at call sites. Adding the methods to the Protocol (and stubbing or NotImplemented-raising in non-Neo4j backends) is cleaner long-term but out-of-scope for this bug fix. _Reviewers: security-reviewer #2, devils-advocate #1, correctness-reviewer #7._
- **L3 — [Observability] Batch helpers' return counts are discarded.** All eight helpers return the number of nodes/rows affected; the cascade ignores these. A log line at INFO level when count > 0 (or DEBUG with the count) would help operational debugging without becoming noisy. _Reviewer: devils-advocate #6._
- **L4 — [Tests] Missing double-forget idempotency test.** `forget(doc)` called twice in a row should be safe (second call hits the "document not found" early-return). Unit-level only — integration would be overkill. _Reviewers: test-reviewer, devils-advocate (architectural pass earlier)._
- **L5 — [Docs] Chronicle pgvector-only-fallback docstring slightly misleading.** The docstring says "no-op when the graph backend does not expose `fetch_document_extraction_state`" — accurate, but a reader could expect a pgvector-direct fallback. Clarify that no-op is intentional: without the graph's partition info, we cannot safely distinguish orphans from survivors in pgvector alone. _Reviewer: devils-advocate #2._

---

## Verified positives (worth keeping)

- **Injection safety**: all Cypher uses parameterized statements with `$entity_ids`/`$namespace_id`/`$doc_id`; all PostgreSQL uses SQLAlchemy ORM with `.where(...in_(entity_ids))`. UUIDs are converted to `str(uuid)` at the Neo4j driver boundary (correct per CLAUDE.md), and passed as native `uuid.UUID` through SQLAlchemy `as_uuid=True` columns (correct).
- **Namespace scoping**: `delete_entities_batch` and `remove_document_from_entity_sources_batch` correctly include `namespace_id` in their WHERE/MATCH clauses.
- **Empty input handling**: all eight helpers short-circuit on empty input lists with no DB roundtrip — verified across all four pgvector and all four Neo4j helpers.
- **Batching is genuine**: each helper issues exactly one statement (`UNWIND` for Cypher, `WHERE IN`/`array_remove` for SQL). No N+1 over the partitioned id lists.
- **Common case (zero extractions) costs one roundtrip**: only `fetch_document_extraction_state` runs; all eight mutation guards short-circuit. Verified by `test_forget_cascade_zero_extraction_skips_backend_calls`.
- **Test coverage**: 17 unit tests across both engines + 2 integration scenarios (IGR-202 reproduction + survivor-relationship preservation), gated on `NEO4J_INTEGRATION_TEST=1` matching the existing `test_replace_document_extraction.py` convention.

---

## Suggested follow-ups (non-blocking)

1. **M1** — add the docs/CHANGELOG entries (5 minutes, can be a follow-up commit).
2. **M2/L2** — promote the four new graph helpers to `GraphBackendProtocol` and remove the `type: ignore` lines (separate PR, scoped cleanup).
3. **L4** — add a unit-level double-forget idempotency test (one ~10-line test in `test_vectorcypher_engine.py`).

sale87 (approved):

PR #740, opened 2026-05-18
fix: pgvector.upsert_entities_batch reports is_new=True for existing entities (#719)

### LINKED ISSUE #719: Running pgvector.upsert_entities_batch reports is_new=True on every upsert, even for existing entities
## Summary

`PgVectorBackend.upsert_entities_batch(namespace_id, entities)` always
returns `(entity, True)` for every entity in the batch, even when the
entity already exists in the namespace. The Neo4j adapter's
`upsert_entities_batch` for the same call returns `(entity, False)`
on the second upsert. On standard pgvector + neo4j, the two halves of the storage layer disagree about whether an
upsert created or merged the entity. Any consumer reading `is_new`
from the pgvector half — including the coordinator's
`upsert_entities_batch` counter — gets an inflated "new entity" count
that doesn't match reality.

The pgvector docstring acknowledges *"is_new is approximate"*, but the
implementation makes no attempt to detect existence at all; the
"approximate" claim is operationally indistinguishable from "always
True". Downstream telemetry / metrics / observability counters that
key off `is_new=True` are silently corrupted.

## Environment

- **Khora**: 0.15.0 (tag `v0.15.0`, commit `8a26676`)
- **Storage backend**: `postgres` (pgvector + Neo4j)
- **Engine**: any
- **Python**: 3.13.11

## Reproduction

```python
"""Repro: pgvector.upsert_entities_batch reports is_new=True on every call,
even for entities that already exist. Compare side-by-side with the neo4j
adapter, which correctly tracks is_new via MERGE."""
import asyncio
import logging
import os
import sys
from uuid import uuid4

os.environ.setdefault("LITELLM_LOG", "ERROR")
os.environ.setdefault("HF_HUB_DISABLE_PROGRESS_BARS", "1")
os.environ.setdefault("TRANSFORMERS_VERBOSITY", "error")
os.environ.setdefault("TOKENIZERS_PARALLELISM", "false")
os.environ["KHORA_DATABASE_URL"] = "postgresql://khora:khora@localhost:5434/khora"
os.environ["KHORA_NEO4J_URL"] = "bolt://neo4j:pleaseletmein@localhost:7688"
logging.getLogger("LiteLLM").setLevel(logging.ERROR)

from loguru import logger
logger.remove()

_NOISE = ("unauthenticated requests to the HF Hub",)
class _Filtered:
    def __init__(self, real): self._real = real
    def write(self, s):
        return len(s) if any(n in s for n in _NOISE) else self._real.write(s)
    def flush(self): self._real.flush()
sys.stdout = _Filtered(sys.stdout)  # type: ignore[assignment]
sys.stderr = _Filtered(sys.stderr)  # type: ignore[assignment]

from khora import Khora
from khora.config import KhoraConfig
from khora.core.models import Entity


async def main() -> None:
    os.environ["KHORA_STORAGE__BACKEND"] = "postgres"
    async with Khora(KhoraConfig(), engine="vectorcypher", run_migrations=True) as kb:
        ns = await kb.create_namespace()
        name = f"alice-{uuid4()}"

        # Two consecutive upserts of the same logical entity (same
        # namespace + name + type), each with a FRESH UUID — this is the
        # shape produced by LLM-driven ingestion where every extraction
        # gets a new candidate UUID and the storage layer dedupes by
        # (namespace, name, type).
        results = {"vector": [], "graph": []}
        for pass_idx in (1, 2):
            e = Entity(id=uuid4(), namespace_id=ns.id, name=name, entity_type="PERSON")
            v = await kb.storage.vector.upsert_entities_batch(ns.id, [e])
            g = await kb.storage.graph.upsert_entities_batch(ns.id, [e])
            results["vector"].append(v[0][1])
            results["graph"].append(g[0][1])
            print(f"pass {pass_idx}: vector.is_new={v[0][1]}  graph.is_new={g[0][1]}")

        if results["vector"] == [True, True] and results["graph"] == [True, False]:
            print(
                "REPRODUCED  pgvector reports is_new=True even when the entity already exists; "
                "neo4j correctly transitions to is_new=False on the second pass — adapters disagree."
            )
        elif results["vector"] == [True, True]:
            print(
                "REPRODUCED (partial)  pgvector reports is_new=True on every upsert; "
                f"neo4j observed sequence={results['graph']}"
            )
        else:
            print(f"NOT reproduced — vector={results['vector']} graph={results['graph']}")


asyncio.run(main())
```

## Expected behavior

`pgvector.upsert_entities_batch` returns `(entity, True)` only when
the row was inserted for the first time, and `(entity, False)` when
an existing row was updated. The pgvector and Neo4j halves of the
dual-write report the same `is_new` for any given entity.

## Actual behavior

```
pass 1: vector.is_new=True  graph.is_new=True
pass 2: vector.is_new=True  graph.is_new=False
REPRODUCED  pgvector reports is_new=True even when the entity already exists;
neo4j correctly transitions to is_new=False on the second pass — adapters disagree.
```

pgvector reports the same entity as "new" twice; Neo4j (for the same
namespace + name + type but a freshly-generated UUID, i.e. the LLM
ingestion shape) correctly reports "not new" on the second upsert.

filippetrovic:
## Summary

- `PgVectorBackend.upsert_entities_batch` hardcoded `(entity, True)` for every entity, even ones already present in the namespace. The Neo4j adapter for the same inputs correctly distinguished insert vs. update, so the two halves of the dual-write disagreed on `is_new`.
- Replaced the hardcoded return with `RETURNING (xmax = 0) AS is_new` piggybacked on the existing `INSERT ... ON CONFLICT DO UPDATE`. `xmax = 0` is PostgreSQL's natural insert-vs-update discriminator: pure inserts leave it zero, the ON CONFLICT DO UPDATE path sets it to the current transaction's xid. No extra round-trip.
- Accumulate the RETURNING rows into a `(name, entity_type) -> is_new` map across all sub-batches under the existing namespace advisory lock, then build the return list against `sorted_entities`. Defensive `logger.warning` calls cover the missing-from-RETURNING fallback and intra-batch-duplicate edge cases surfaced by review.
- Added an integration test (`tests/integration/storage/test_pgvector_upsert_is_new.py`) covering the issue's repro (fresh UUIDs, colliding name+type), the empty-input short-circuit, and a cross-sub-batch-boundary case (`batch_size=2` over 5 entities) to verify `is_new` accumulates correctly across multiple statements inside one transaction. The protocol-compliance test that caught the bug on graph backends only runs against the `graph_backend` fixture — pgvector is the *vector* adapter, which is why this regressed unnoticed.

## Impact

- **pgvector-only stacks**: `pipelines/flows/ingest.py:1286` (`needs_embedding = is_new or not entity.embedding`) stops triggering on every existing entity, eliminating wasted embedding-model calls.
- **Coordinator counters** (`storage/coordinator.py:652-653`) and **ingest event types** (`ingest.py:1248`: `entity.created` vs `entity.updated`) now reflect reality.
- **Dual-stack (pgvector + Neo4j)** consumers were shielded today because the coordinator returns Neo4j's results; this fixes direct vector-adapter callers and brings the two halves into agreement.

## Test plan

- [x] `uv run pytest tests/integration/storage/test_pgvector_upsert_is_new.py -v` — all 3 tests pass against the fix
- [x] Regression gate verified: tests 1 and 3 fail against the unfixed backend (per QA's `git stash` run during development)
- [x] `make format && make lint` clean
- [ ] Full `make test` runs in CI (local run blocked by docker-compose port conflicts with another project; CI's stack is isolated)
- [ ] Manual: issue #719's repro script prints `vector=[True, False] graph=[True, False]` after merge

Fixes #719

### DIFF
--- a/src/khora/storage/backends/pgvector.py
+++ b/src/khora/storage/backends/pgvector.py
@@ -12,7 +12,7 @@ from typing import TYPE_CHECKING, Any
 from uuid import UUID
 
 from loguru import logger
-from sqlalchemy import delete, func, select, text, update
+from sqlalchemy import delete, func, literal_column, select, text, update
 from sqlalchemy.ext.asyncio import AsyncEngine, AsyncSession, async_sessionmaker, create_async_engine
 from tenacity import AsyncRetrying, retry_if_exception, stop_after_attempt, wait_exponential
 
@@ -898,7 +898,10 @@ class PgVectorBackend(AsyncSessionMixin):
         deadlocks when multiple documents share entities.  Upserts for
         different namespaces proceed in parallel without contention.
 
-        Returns list of (entity, is_new) tuples (is_new is approximate).
+        Returns list of (entity, is_new) tuples. `is_new` is computed via
+        ``RETURNING (xmax = 0)``: ``xmax = 0`` indicates a true insert, while
+        a non-zero ``xmax`` means the ``ON CONFLICT DO UPDATE`` path acquired
+        a row lock on the existing row.
         """
         if not entities:
             return []
@@ -909,6 +912,7 @@ class PgVectorBackend(AsyncSessionMixin):
             # Sort by (namespace_id, name, entity_type) to ensure consistent lock ordering
             sorted_entities = sorted(entities, key=lambda e: (str(e.namespace_id), e.name, str(e.entity_type)))
             lock_key2 = _namespace_lock_key(namespace_id)
+            is_new_map: dict[tuple[str, str], bool] = {}
 
             async with self._get_session() as session:
                 # Acquire namespace-scoped advisory lock for the duration of this
@@ -959,13 +963,44 @@ class PgVectorBackend(AsyncSessionMixin):
                             "metadata": stmt.excluded.metadata,
                             "updated_at": stmt.excluded.updated_at,
                         },
+                    ).returning(
+                        EntityModel.name,
+                        EntityModel.entity_type,
+                        literal_column("(xmax = 0)").label("is_new"),
                     )
-                    await session.execute(stmt)
+                    result = await session.execute(stmt)
+                    for row in result:
+                        key = (row.name, row.entity_type)
+                        if key in is_new_map:
+                            logger.warning(
+                                "upsert_entities_batch: duplicate (name, entity_type) "
+                                "in input batch; is_new for first occurrence may be inaccurate "
+                                "(name={name!r}, entity_type={entity_type!r}, namespace_id={namespace_id})",
+                                name=row.name,
+                                entity_type=row.entity_type,
+                                namespace_id=namespace_id,
+                            )
+                        is_new_map[key] = bool(row.is_new)
 
                 # Single commit for all sub-batches under the advisory lock
                 await session.commit()
 
-            return [(entity, True) for entity in sorted_entities]
+            results: list[tuple] = []
+            for entity in sorted_entities:
+                key = (entity.name, entity.entity_type)
+                if key in is_new_map:
+                    results.append((entity, is_new_map[key]))
+                else:
+                    logger.warning(
+                        "upsert_entities_batch: entity missing from RETURNING; "
+                        "falling back to is_new=True "
+                        "(name={name!r}, entity_type={entity_type!r}, namespace_id={namespace_id})",
+                        name=entity.name,
+                        entity_type=entity.entity_type,
+                        namespace_id=namespace_id,
+                    )
+                    results.append((entity, True))
+            return results
 
         return await _retry_on_deadlock(_do_upsert)
 
--- a/tests/integration/storage/test_pgvector_upsert_is_new.py
+++ b/tests/integration/storage/test_pgvector_upsert_is_new.py
@@ -1,0 +1,174 @@
+"""Integration test for `PgVectorBackend.upsert_entities_batch` is_new tracking.
+
+Regression test for GitHub issue #719. Prior to the fix, the pgvector
+backend returned ``is_new=True`` for every entity — even on the second
+upsert of the same ``(namespace_id, name, entity_type)`` — because the
+function hard-coded ``return [(entity, True) for entity in sorted_entities]``.
+
+The Neo4j adapter's parametrized compliance test
+(``tests/unit/storage/backends/test_protocol_compliance.py::test_upsert_entities_batch_new_vs_existing``)
+runs against the ``graph_backend`` fixture only, so this gap on the
+vector adapter went undetected. This test closes that gap by exercising
+pgvector directly against a real PostgreSQL.
+
+The fix uses ``RETURNING (xmax = 0) AS is_new`` to discriminate true
+inserts from ``ON CONFLICT DO UPDATE`` paths.
+
+Requires a running PostgreSQL instance from this repo's compose stack
+(``make dev``); port ``5434`` per ``compose.yaml``. Set
+``KHORA_DATABASE_URL`` to override.
+"""
+
+from __future__ import annotations
+
+import os
+from datetime import UTC, datetime
+from uuid import uuid4
+
+import pytest
+
+from khora.core.models import Entity, MemoryNamespace
+from khora.db.session import run_migrations
+from khora.storage.backends.pgvector import PgVectorBackend
+from khora.storage.backends.postgresql import PostgreSQLBackend
+
+DATABASE_URL = os.environ.get(
+    "KHORA_DATABASE_URL",
+    "postgresql+asyncpg://khora:khora@localhost:5434/khora",
+)
+
+if DATABASE_URL.startswith("postgresql://"):
+    DATABASE_URL = DATABASE_URL.replace("postgresql://", "postgresql+asyncpg://", 1)
+elif DATABASE_URL.startswith("postgres://"):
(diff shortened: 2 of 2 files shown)


filippetrovic:
# Multi-Perspective Review — PR #740

**Team:** `pr-740-review` (7 reviewers + security audit)
**Reviewers:** security, performance, correctness, test, database, docs, devil's-advocate
**Verdict:** **NEEDS_REVIEW** — no CRITICAL/HIGH findings to block merge after rebuttal of one disputed CRITICAL; 2 MEDIUMs worth addressing; 3 LOWs optional.

---

## Findings

### CRITICAL (disputed)
- **C1 [Database] `xmax = 0` discrimination is broken** — `src/khora/storage/backends/pgvector.py:969`. The reviewer claimed that on the `ON CONFLICT DO UPDATE` path the new tuple version has `xmax = 0` (same as INSERT), so the discriminator never returns `False`. **REBUTTED.** Empirical evidence contradicts this: QA's three integration tests pass against the fix, and verified via `git stash` that two of them fail against the unfixed (always-True) backend. The `xmax = 0` pattern is the documented idiom for `INSERT ... ON CONFLICT DO UPDATE ... RETURNING` — PostgreSQL's conflict path acquires a row-level lock on the existing row before the UPDATE, which sets `xmax` to the current transaction's xid on the returned tuple. The accepted Stack Overflow answer on this question, depesz's writeup, and several library implementations rely on this exact pattern. **No change requested.**

### HIGH
- None.
- *(Correctness reviewer flagged result-ordering as HIGH — but this is pre-existing behavior, unchanged by this PR. Downgraded to L2 below.)*

### MEDIUM
- **M1 [Security] Tenant strings in structured log fields** — `src/khora/storage/backends/pgvector.py:975-982, 989-1001`. The two new `logger.warning` calls pass `entity.name` and `entity.entity_type` as loguru structured kwargs. Per khora's telemetry contract (CLAUDE.md "Free-text span attributes" rule), free-text values bound for OTEL/Logfire export must be hashed via `khora.telemetry.bounded_text_hash`. While these are log fields rather than span attributes today, if the loguru→logfire bridge is configured the fields will surface in OTEL. **Fix:** replace with `bounded_text_hash(row.name)` / `bounded_text_hash(entity.name)`.
- **M2 [Devil's Advocate] Missing-from-RETURNING fallback hides query failures** — `src/khora/storage/backends/pgvector.py:989-1001`. The second `logger.warning` path falls back to `is_new=True` when an input entity has no corresponding RETURNING row. If this fires it indicates a real invariant violation (asyncpg result truncation, silent statement failure, or model-vs-table column mismatch). **Argument for raising:** silent fallback masks bugs that should crash loud. **Argument for warning:** matches the pre-fix behavior (defensive). Tilt toward `raise AssertionError(...)` since this branch can never legitimately fire — see the implementer's note in the docstring that RETURNING covers every row processed by the statement.

### LOW
- **L1 [Test] Unused fixture parameter** — `tests/integration/storage/test_pgvector_upsert_is_new.py:146` — `test_upsert_entities_batch_empty` takes `relational: PostgreSQLBackend` but never uses it. Drop the parameter.
- **L2 [Correctness/Docs] Result-order undocumented** — `src/khora/storage/backends/pgvector.py:901-904, 989`. The function returns results in sorted order (`namespace_id, name, entity_type`), not input order. This is pre-existing behavior — the PR does not change it. Worth a one-line addition to the docstring while the file is open.
- **L3 [Devil's Advocate] Parametrize protocol-compliance test for vector backends** — `tests/unit/storage/backends/test_protocol_compliance.py`. The root cause of #719 not being caught was that the parametrized compliance test runs against `graph_backend` only. The new pgvector-specific test closes the immediate gap; parametrizing the compliance test across vector backends would prevent future regressions in any vector adapter. Reasonable scope-creep for this PR; reasonable follow-up if scope is held tight.

### Clean
- **Performance** — RETURNING piggybacks on existing INSERT, no extra round-trip; `is_new_map` is O(n) bounded by batch size; no N+1; no index/plan implications.
- **Docs** — docstring update is accurate; no stale "is_new is approximate" references elsewhere; no telemetry-contract.json update needed (`is_new` is not on any span/metric).
- **Security audit (skill)** — no findings in the 16 OWASP-adjacent categories on the diff. Only `session.execute(stmt)` matched the injection grep, and `stmt` is a SQLAlchemy ORM expression with bound parameters, not a raw string.

---

## Recommended Pre-Merge Actions

1. Address **M1** (hash tenant strings in `logger.warning` fields) — small, well-scoped, matches the existing telemetry policy.
2. Decide on **M2** — switch the missing-from-RETURNING warning to `raise`, or keep as-is and accept the defensive behavior. Both are defensible; pick one and document the choice.
3. Drop the unused `relational` fixture parameter in `test_upsert_entities_batch_empty` (**L1**).
4. (Optional) Add a one-line docstring note about sorted return order (**L2**).
5. (Optional follow-up) Parametrize `test_protocol_compliance.py` across vector backends (**L3**).

CI must be green before merge. The empirical evidence (QA's tests pass + would-have-failed-before-fix verification) supersedes the disputed CRITICAL finding.

probablyrobot (approved):

filippetrovic:
Superseded by #742, which merged a near-identical fix (also `xmax = 0` + `RETURNING` on the existing `INSERT ... ON CONFLICT DO UPDATE`) for the same #719. Closing as redundant rather than rebasing — the fix is already on `main` as of 522d839.

PR #747, opened 2026-05-18
API: source_type/source_name/source_url kwargs on remember/remember_batch/submit_batch

filippetrovic:
## Summary

- `Khora.remember()`, `Khora.remember_batch()`, and `Khora.submit_batch()` now accept `source_type`, `source_name`, and `source_url` kwargs. `source_type` defaults to `"library"` at the public API surface; engine-internal defaults (e.g. `"api"`) are preserved at engine call sites.
- For batch dict variants, the same three keys at the top level of each doc dict override the call-level kwarg per document.
- `DocumentMetadata` is widened with `source_name` and `source_url` (`str`, default `""`).
- Storage mappers in `postgresql`, `sqlite`, `sqlite_lance`, and `surrealdb` thread the new fields through to `DocumentModel.source_name` / `DocumentModel.source_url` (columns added in #743), with `""` ↔ `None` coercion at the storage boundary.
- Three engines (`vectorcypher`, `chronicle`, `skeleton`) and `MemoryEngineProtocol` updated to forward the new kwargs.

## Test plan

- [x] 8 new unit tests in `tests/unit/test_khora.py` covering `remember()`, `remember_batch()`, and `submit_batch()` for explicit kwargs, defaults, top-level vs per-doc precedence, and mixed-doc batch scenarios.
- [x] 3 storage-layer round-trip tests in `tests/unit/storage/backends/sqlite_lance/test_relational.py` covering explicit values, defaults persisted as NULL on disk and read back as `""`, and update-to-empty coercion to NULL.
- [x] Existing `remember()` tests pass unchanged (no signature break).
- [x] `make format` clean.


filippetrovic:
## Consolidated Review — Multi-Perspective

Eight reviewers ran in parallel against the diff: security, performance, correctness, tests, database, observability, docs, devil's advocate.

**Overall: APPROVE.** No HIGH/CRITICAL findings. Two MEDIUM documentation drifts and a handful of LOW architectural style notes. CI green (8/8 checks).

### MEDIUM

- **M1 [Docs]** `docs/data-models/documents-chunks.md:36` — The `DocumentMetadata` example still shows only `source_type: str = "manual"` and is missing the new `source_name` / `source_url` fields. Update the example to match the widened dataclass.
- **M2 [Docs]** `docs/api-reference.md:109-125` — The `remember_batch()` signature in the API reference is missing the three new kwargs. `submit_batch()` also lacks a dedicated section (currently only mentioned in passing). Add `source_type` / `source_name` / `source_url` to both signatures with a brief note about per-doc dict override behavior.

### LOW

- **L1 [Performance]** `src/khora/engines/chronicle/engine.py:2278-2296` — The per-doc `doc_inputs` loop calls `.get()` 7+ times per document (content, title, source, source_type, source_name, source_url, metadata, plus conditional external_id). Pure prepare-side overhead; does not affect DB/LLM hot paths. Worth a future micro-optimization if profiling flags it, but not a blocker.
- **L2 [Style/Devil's Advocate]** `DocumentMetadata` widening — adding `source_name` / `source_url` to `DocumentMetadata` follows the existing pattern alongside `source_type`, but `source_type` is already mirrored on `DocumentModel`. A first-class flattening onto `Document` (as `external_id` / `extraction_config_hash` already are) would remove a layer of indirection. Defensible either way; current approach is the lower-blast-radius choice.
- **L3 [Style/Devil's Advocate]** Per-doc `submit_batch()` overrides may be slightly ahead of demonstrated need. The mixed-doc test exercises the precedence rule, but there's no clear producer in-tree that benefits from per-doc overrides on the deferred path today. Forward-compatible, just speculative.
- **L4 [Style/Devil's Advocate]** Engine-level kwargs are forwarded through all three engines but are persistence-only — they are never used for business logic. Threading them through `metadata.custom` instead would shrink the `MemoryEngineProtocol` surface by three parameters. Trade-off is call-site explicitness vs. protocol minimalism; either choice is reasonable.

### Per-reviewer summary

| Reviewer | Verdict | Notes |
|---|---|---|
| Security | APPROVE | `source_url` bounded by `VARCHAR(2048)`; no telemetry exposure; parameterized queries; ADR-084 names checked (provenance metadata, not credentials — `str` is correct). |
| Performance | APPROVE | O(n) prep-side cost only; no N+1, no hot-path regression. One LOW micro-opt above. |
| Correctness | APPROVE | Per-doc precedence consistent across all three engines; `""` ↔ `None` boundary correct; engine `source_type="api"` default never overrides the public `"library"` default (forwarded explicitly). No placeholders. |
| Tests | APPROVE | 8 façade tests + 3 storage-roundtrip tests cover defaults, overrides, mixed-doc precedence, and NULL coercion. |
| Database | APPROVE | `""` ↔ `None` coercion consistent across postgresql, sqlite, sqlite_lance, surrealdb. aiosqlite DDL updated; Alembic migration 037 handles existing dev DBs. Column shapes match. |
| Observability | APPROVE | No new spans or metric labels; no cardinality risk from `source_url`. |
| Docs | REQUEST_CHANGES (M1, M2) | Two doc drifts flagged above. |
| Devil's Advocate | APPROVE with style notes | L2/L3/L4 above. No blocking concerns. |

### Recommendation

Address M1 and M2 (documentation drift) before merge — both are short, mechanical edits in `docs/`. The LOW items are non-blocking and can be deferred or wontfix.

msrdic (approved):

PR #748, opened 2026-05-19
Flatten Document/Chunk metadata + split chunker output

filippetrovic:
## Summary

- Delete the `DocumentMetadata` and `ChunkMetadata` dataclass wrappers; lift their typed fields directly onto `Document` and `Chunk`.
- Replace the entangled `{**doc_metadata, **chunker_output}` merge in the chunk pipeline with a clean two-field split: `Chunk.metadata` carries the propagated doc-level dict, the new `Chunk.chunker_info` dict carries chunker output. No more silent key shadowing.
- Add `Document.source_name` and `Document.source_url` typed-provenance fields; drop the legacy `source_tool` field.
- Update every storage backend converter (postgresql, sqlite, pgvector, sqlite_lance relational+vector, surrealdb relational+vector) to read and write the new columns. Coerce legacy empty strings to `None` for nullable string fields; `source_type` coerces `""` or `NULL` to `"library"`.
- Add `source_timestamp` propagation on the postgresql and sqlite_lance/relational document converters where it was missing.
- Sweep ~16 read-path source files (engines, query, chat, pipelines, five integration adapters) and ~58 test files for flat-field access.

## Test plan

- [x] `make format` clean.
- [x] `make lint` clean (ruff + ty).
- [x] `uv run pytest tests/unit` — 4101 passed, 250 skipped (langgraph extra is env-gated locally; full suite runs on CI).
- [x] New tests: chunker output isolation, aliasing safety, `source_timestamp`/`source_name`/`source_url` round-trips, empty-string-to-`None` coercion, public-API removal for the deleted wrappers, `chunker_info` round-trip on pgvector, sqlite, and sqlite_lance vector.
- [ ] Integration suite via CI (pg + surrealdb stacks).

## Breaking changes

`Document` and `Chunk` are part of the public OSS API surface. Downstream callers must migrate:

- `doc.metadata.title` → `doc.title` (and same for `source`, `source_type`, `content_type`, `author`, `language`, `checksum`, `size_bytes`)
- `doc.metadata.source_tool` → `doc.source_name`
- `doc.metadata.custom["k"]` → `doc.metadata["k"]`
- `chunk.metadata.custom["k"]` → `chunk.metadata["k"]`
- `chunk.metadata.chunker_info` → `chunk.chunker_info`
- `chunk.metadata.{chunk_index,start_char,end_char,token_count,document_id}` → flat on `Chunk`

`khora.remember(metadata={...})` kwarg signature is unchanged — the dict still lands on `doc.metadata`.


filippetrovic:
## Multi-Perspective Review

**Overall: APPROVE with minor follow-up suggestions.**

Seven reviewers ran in parallel (security, performance, correctness, test, docs, database, devil's advocate) plus a dedicated security audit. No CRITICAL findings. One MEDIUM correctness item was auto-fixed during review. The remaining items are MEDIUM/LOW: test-coverage symmetry gap and a few design questions for follow-up.

### HIGH
_None._

### MEDIUM

- **M1 [Test] — SurrealDB round-trip gap.** `TestDocumentSourceTimestampRoundTrip` and `TestDocumentSourceNameUrlRoundTrip` exist for PostgreSQL and SQLite-Lance but not SurrealDB. `TestSurrealDBBackendMapperCoercion` covers the coercion semantics, but no explicit round-trip for the trio of new fields. Symmetry would be cheap.
- **M2 [Test] — SurrealDB `chunker_info` round-trip gap.** `chunker_info` has explicit round-trip tests on pgvector, sqlite, and sqlite_lance/vector, but not on SurrealDB vector. The converter writes/reads `chunker_info` correctly (`row.get("chunker_info", {})` with `isinstance` guard at `surrealdb/vector.py:742-744`) — just no test asserting end-to-end.
- **M3 [Correctness] — Stale comment.** `src/khora/engines/vectorcypher/retriever.py:2620` referenced the deleted `metadata.custom` field. **Auto-fixed in commit `4061408` during review** — comment now says "metadata dict" to match the new field shape. Behavioral impact: none (comment-only).

### LOW

- **L1 [Design] — Empty-string-to-None coercion: code that previously checked `doc.title == ""` will silently miss values that now read as `None`.** New coercion behavior is consistent across PostgreSQL/SQLite-Lance/SurrealDB and tested. Truthy checks (`if doc.title:`) continue to work; explicit `== ""` comparisons in downstream code would not. Pre-merge grep across `src/` finds no such comparisons.
- **L2 [Performance] — Chunker pipeline does `dict(document.metadata)` + `dict(result.metadata)` per chunk** (`src/khora/pipelines/tasks/chunk.py:50-51`). This is a deliberate isolation barrier — shallow copies prevent mutations to caller-held dicts from leaking into chunks. Net copy count is comparable to the old merge path (which already did spread-construction). Not a regression; flagging for visibility if someone benchmarks the pipeline later.
- **L3 [Database] — Defensive read inconsistency.** `sqlite_lance/relational.py` reads new columns via `getattr(model, "source_name", None)` (ORM-attribute defensive); `postgresql.py` accesses directly (`model.source_name`). Both safe; consistency could improve. Non-blocking.
- **L4 [Design] — `Document.language` default changed from `"en"` to `None`.** The change is intentional (avoids hardcoded language assumption); `None`-safe access patterns are in place across `src/khora/`. Worth noting in the public changelog for downstream consumers who may have relied on the `"en"` fallback.

### Devil's-Advocate Concerns (Design Discussion, Non-Blocking)

These are valid design questions the reviewer raised. The PR's current approach is defensible; calling them out so the rationale is on record.

- **D1 — No deprecation shim for the removed wrappers.** A `@property metadata` returning a dataclass-shaped view could have given downstream consumers (`khora-cli`, `khora-explorer`) one release of compatibility. The current approach is a clean break — appropriate when there is no in-tree consumer of the old API and downstream packages will be coordinated in a single release.
- **D2 — `source_name` and `source_url` as typed fields vs `metadata` dict keys.** Typed columns enable future filtering/indexing (matches the migration's intent). The added schema cost is two nullable columns per backend; the maintenance cost matches the rest of the typed-provenance fields already on `Document`.
- **D3 — 100 files in one PR.** Could have split into "flatten types + converters" then "sweep accessors". Single-PR approach is acceptable because the bisect surface is bounded — any regression should manifest as a clear `AttributeError` or `TypeError`, not a subtle semantic drift.

### Security

- **Security Reviewer (team): no findings.** All SQL/SurrealQL writes for the new columns are parameterized. `source_url` is not logged directly. The free-form `Document.metadata` dict is not pushed to span attributes or metrics (which would be a cardinality/PII concern). Public-API removal is clean — no isinstance-based unsafe coercions remain.
- **Security Audit (dedicated 16-category scan): PASS.** No injection sinks, no broken auth/access changes, no new secrets handling, no SSRF, no deserialization, no XSS, no path-traversal, no mass-assignment, no regex DoS, no supply-chain changes in this PR. The 1,283 added lines are mechanical field-access rewrites and converter additions.

### Database

- All seven converter implementations handle the new schema correctly. Defensive coding is strong: `_none_if_empty()` helper, `_row_get` fallback for legacy SQLite DBs missing the new columns, `isinstance(..., dict)` guard on SurrealDB `chunker_info`.
- Existing partial index `ix_documents_namespace_checksum_active WHERE checksum != ''` continues to behave correctly under the empty-to-None coercion (NULL never matches the predicate, which is the intended semantics).
- `source_timestamp` propagation gap closed on both PostgreSQL and SQLite-Lance/relational converters.

### Docs

- Public-API docs (`docs/data-models/*.md`, `docs/integrations/*.md`, `docs/consumers.md`) accurately describe the new flat shape. Integration example files remain byte-identical to their markdown code blocks (CI-checked). No stale references to `DocumentMetadata` / `ChunkMetadata` anywhere.

### Tests

- 58 test files updated for flat-field access; 6+ new test cases:
  - Chunker output isolation (the key-collision regression test that locks the no-shadowing invariant)
  - Aliasing safety after shallow-copy split
  - `source_timestamp` round-trip (PostgreSQL + SQLite-Lance)
  - `source_name` / `source_url` round-trip (PostgreSQL + SQLite-Lance)
  - Empty-string-to-None coercion across backends
  - Public-API removal via `pytest.raises(ImportError)` plus `__all__` assertion
  - `chunker_info` round-trip (pgvector, sqlite, sqlite_lance vector)

### Suggested Follow-Up (Non-Blocking)

1. Add SurrealDB-side equivalents of `TestDocumentSourceTimestampRoundTrip`, `TestDocumentSourceNameUrlRoundTrip`, and a `chunker_info` round-trip test (M1, M2).
2. Add a release-note line about the `Document.language` default change (`"en"` → `None`) for downstream consumers.

The PR is in good shape — recommended for merge once CI completes.

msrdic (approved):

PR #751, opened 2026-05-19
Introduce typed recall projection dataclasses

filippetrovic:
## Summary

- Add a new public module ``khora.core.models.recall`` exporting five frozen, slots-enabled projection dataclasses returned by recall operations: ``DocumentProjection``, ``RecallChunk``, ``RecallEntity``, ``RecallRelationship``, ``RecallResult``.
- ``RecallResult`` replaces the previous tuple-shaped recall response. ``chunks: list[RecallChunk]`` carries the score as a typed field; ``documents: list[DocumentProjection]`` is the new deduplicated top-level set referenced by chunks, entities, and relationships. ``usage`` replaces ``llm_usage``; ``engine_info`` replaces ``metadata``. ``context_text`` is removed from the public surface — a small ``context_text(result)`` helper lands in a follow-up.
- All three engines (``vectorcypher`` / ``chronicle`` / ``skeleton``) now construct the new shape and emit a mandatory ``engine_info["engine"] = "<strategy-name>"`` key so downstream consumers can route on producer identity. Chronicle and skeleton gained the ``"engine"`` key (they were previously emitting only ``"backend"``).
- ``DocumentProjection.metadata`` is populated from each chunk's propagated doc-level metadata (post-flatten, ``Chunk.metadata`` is the doc dict, so first-chunk-wins per ``document_id`` is correct). The current projections are deliberately stubs — fuller ``Document`` fetching arrives with the recall-method rewrite.
- All in-tree adapters (``crewai``, ``langgraph``, ``llamaindex``, ``google_adk``, ``openai_agents``) and the ``chat`` engine migrate to the new shape via the ``chunk.document_id`` → ``documents_by_id[doc_id]`` join pattern. ``RecallChunk`` deliberately does not carry ``metadata`` or ``namespace_id`` — those live on ``DocumentProjection.metadata`` and the top-level ``RecallResult.namespace_id``.
- ``Khora.recall(..., include_sources=True)`` is now a documented no-op kept for API stability. The previous implementation mutated ``Chunk.source_document`` / ``Entity.source_documents`` in place, which is incompatible with frozen projections. Full source population returns with the recall-method rewrite.
- Test sweep: ~21 ``RecallResult(...)`` construction sites migrated; many tuple-access call sites updated; new ``tests/unit/test_recall_projections.py`` smoke-tests dataclass construction, frozen semantics, and the producer invariant. Five ``TestIncludeSources`` tests rewritten to assert against ``result.documents`` membership (id-based) instead of the previous in-place-mutation identity checks.

## Migration burden for downstream consumers

- ``RecallEntity`` and ``RecallRelationship`` fields have no defaults — every constructor must supply them (matches the spec; ergonomics pass can revisit later).
- ``result.chunks[i][0]`` / ``result.chunks[i][1]`` tuple access patterns are gone — read ``.id``, ``.content``, ``.score`` directly.
- ``result.metadata`` → ``result.engine_info``; ``result.llm_usage`` → ``result.usage``; ``result.context_text`` removed.

## Test plan

- [x] ``uv run ruff check src tests`` — clean
- [x] ``uv run ruff format --check src tests`` — clean
- [x] ``uv run pytest tests/unit --ignore=tests/unit/integrations/langgraph`` — 4630 passed, 250 skipped, 0 failed
- [ ] Full CI battery (unit + integration + matrix backends + examples-smoke)
- [ ] Manual: ``python -c "from khora import RecallResult; print(RecallResult.__module__)"`` → ``khora.core.models.recall``


filippetrovic:
# Multi-perspective PR review

Spawned a 7-agent reviewer team across security, performance, correctness, tests, docs, observability, and devil's advocate. Consolidated findings below, deduplicated and ranked.

**Overall assessment:** REQUEST_CHANGES — one HIGH correctness gap in chronicle's `documents` build + several documentation items that should land before merge.

## HIGH

- **H1 [Correctness]** `src/khora/engines/chronicle/engine.py:1530-1561` — Chronicle's `documents` list is built from `chunks_with_scores` only. Entities returned via `recall_entities` carry `source_document_ids` that may reference documents NOT present in any chunk. This violates the producer invariant documented in `RecallResult` (every id in `entities[i].source_document_ids` MUST appear as some `documents[j].id`). Vectorcypher applies the second/third union loops; chronicle skips them. **Today**, chronicle's entity list is empty post-stomp at L1521 so the invariant is vacuously held — but the invariant is broken at the structural level. Either restore the union loops, or document that chronicle's entities are intentionally empty.
- **H2 [Correctness/DevilsAdvocate]** `src/khora/engines/chronicle/engine.py:1521` — `entity_hits: list[tuple[Entity, float]] = []` unconditionally stomps the result of `_collect_entities()` 11 lines earlier. Pre-existing bug; preserved here per "stay surgical." Recommend a focused follow-up — the dead loop after this stomp ships unused entity-source-doc backfill logic. Should not block this PR but should be tracked.
- **H3 [Docs]** `docs/api-reference.md:298-310` — `RecallResult` field table is stale (still lists tuple-based chunks/entities/relationships, `context_text`, `metadata`, `llm_usage`). Must be updated before merge.
- **H4 [Docs]** `CHANGELOG.md` — No `## [Unreleased]` entry for this breaking change to the stable public API. Must document the field renames (`metadata→engine_info`, `llm_usage→usage`), the removal of `context_text`, the tuple-to-dataclass migration, and the four new public exports.

## MEDIUM

- **M1 [Observability]** `docs/telemetry-contract.json` — Five new names added to `khora.__init__.py.__all__` (`DocumentProjection`, `RecallChunk`, `RecallEntity`, `RecallRelationship`, `RecallResult`). The telemetry contract's `public_exports` array should be reviewed for drift. CI currently passes so the test may not be strict on these specific exports, but the contract should reflect that the new types are part of khora's stable public surface.
- **M2 [Observability]** `engine_info["engine"]` is now a contractual stable key across all three engines. Document it in `RecallResult` docstring (already done — "every engine MUST emit") or in `docs/telemetry-contract.json` as a stable model field, so downstream consumers know to route on it.
- **M3 [Security]** `src/khora/core/models/recall.py:34` — `DocumentProjection.source_url: str | None` matches the typed-secret-handling secret-name pattern (`*_url where the URL conventionally embeds credentials`). URLs like `https://user:pass@example.com/...` can leak in logs/traces. Two questions: (a) Is the typed-secret-handling rule in scope for plain `@dataclass` (the rule's wording emphasizes Pydantic models)? (b) If yes, type as `SecretStr | None` or annotate the explicit opt-out. If no, document the exemption decision.
- **M4 [DevilsAdvocate/Correctness]** `src/khora/engines/vectorcypher/engine.py:~1864` — `list(entity.source_document_ids) or list((entity.source_documents or {}).keys())` defensive fallback. Either the typed `source_document_ids` field is always populated (making the fallback dead code) or both paths are legitimate (in which case chronicle's missing fallback is a silent-data-loss risk). Document the canonical source and reduce to one read path across both engines.
- **M5 [DevilsAdvocate]** `src/khora/khora.py` — `del include_sources` (and the parameter being a documented no-op) is honest but cognitively noisy. Alternatives: remove the kwarg entirely (breaking change, but the API is already breaking here), or raise `NotImplementedError("source population deferred")` so callers fail loud instead of getting silently-empty results.
- **M6 [DevilsAdvocate]** `src/khora/core/models/recall.py` — The five projection dataclasses use `frozen=True` with `list`/`dict` fields. `frozen` prevents attribute reassignment but not in-place mutation; `chunk.connected_entity_ids.append(...)` and `chunk.chunker_info["k"]=...` both succeed. Also: `hash(recall_chunk)` raises `TypeError: unhashable type` because the auto-generated `__hash__` walks the field tuple and hits the list. Three remediations to choose from in a follow-up: docstring caveat, `eq=False` for identity-equality, or convert containers to tuples / frozen mappings.
- **M7 [Performance]** Engine dedup loops allocate one `DocumentProjection(metadata=dict(chunk.metadata or {}))` per unique `document_id`. Negligible at typical recall sizes (5-20 unique docs); could matter on high-cardinality recalls (1k+ unique docs). The shallow-copy of the propagated doc-metadata dict per stub document is acceptable for now but worth profiling once the recall-method rewrite fetches richer projections.
- **M8 [Docs]** `docs/consumers.md` — The downstream-consumer doc lists which symbols cross the stability boundary. Add the four new names alongside the existing `RecallResult` and note that downstream packages must be updated in lockstep.

## LOW

- **L1 [Tests]** `tests/integration/matrix/test_skeleton_pg.py` (and `test_skeleton_sqlite_lance.py`) — Verify `engine_info.get("backend")` but not `engine_info.get("engine") == "skeleton"`. Vectorcypher matrix asserts the engine key; skeleton should too. Generic smoke test in `tests/unit/test_recall_projections.py` covers the invariant, but matrix-level coverage would be tighter.
- **L2 [DevilsAdvocate]** `engine_info` rename from `metadata`: semantically tighter, but no compat alias. Long-lived adapter / dashboard code reading the old name silently breaks. The PR is already a breaking change; documenting the rename in the changelog (H4) covers this.
- **L3 [DevilsAdvocate]** `DocumentProjection.metadata=dict(chunk.metadata or {})` is stub data — the field could have been left empty in this PR and populated in the recall-method rewrite. Acceptable as-is because adapters depend on the join pattern working today, but worth noting the field couples to chunk metadata propagation behavior that may change.
- **L4 [Docs]** Integration examples under `examples/integrations/*/example.py` — verify no example references `result.chunks[i][0/1]`, `result.context_text`, `result.metadata`, or `result.llm_usage`. The `tools/check_examples_drift.py` smoke gate catches byte-drift, not content drift.

## What was reviewed

- All five new dataclasses + their docstrings (OSS-safe ✓; producer invariant clearly documented ✓; frozen + slots ✓)
- All three engines' projection construction (engine_info["engine"] mandatory key ✓; dedup correctness ✓ for vectorcypher; gap in chronicle — see H1)
- All five in-tree adapters' doc-id join pattern (built once per call ✓; reads `result.namespace_id` instead of stale per-chunk namespace ✓)
- Test sweep (~21 fixture migrations + 5 `TestIncludeSources` rewrites — meaningfully migrated to id-based assertions, no skip/xfail introduced ✓)
- New `tests/unit/test_recall_projections.py` — real tests, exercises frozen semantics + producer invariant ✓
- Import-cycle resolution via `TYPE_CHECKING` + `from __future__ import annotations` (verified across 4 import entrypoints ✓)
- No internal identifiers leaked into source or commits ✓

probablyrobot (approved):

probablyrobot (approved):

PR #752, opened 2026-05-19
Reranker contract: typed doc_title field + chronicle group-key collapse

filippetrovic:
## Summary

- Add typed `RerankCandidate.doc_title: str = ""` field — title now travels as a first-class candidate attribute instead of riding inside the free-form `metadata` dict.
- Both rerankers (`CrossEncoderReranker`, `LLMReranker`) read `c.doc_title` instead of `c.metadata.get("title")`. The legacy lookup path is gone.
- Drop the title denormalization at ingest: chunks no longer get `metadata["title"]` injected after chunking. Title flows through the candidate field at recall time instead.
- Collapse the chronicle version-demotion group-key fallback chain from `entity_refs[0] → custom["title"] → document_id` to `entity_refs[0] → document_id`. Same-doc chunks still group identically — only the debug-log group key changes from a title string to a doc UUID.
- New unit tests assert the typed-field prefix path for both rerankers and explicitly verify that stale `metadata["title"]` is ignored when `doc_title` is empty (regression guard against the old lookup creeping back).
- Orphan test for the removed chronicle title fallback deleted (no code path left to assert).

## Test plan
- [x] `uv run pytest tests/unit/test_query_reranking.py` — 27 tests pass (4 new contract-driven assertions: 2 cross-encoder, 2 LLM listwise)
- [x] `uv run pytest tests/unit/engines/test_chronicle_version_scoring.py` — pre-existing tests still pass after title-fallback test removal
- [x] `ruff check` + `ruff format --check` on all changed files — clean

### DIFF
--- a/src/khora/engines/chronicle/engine.py
+++ b/src/khora/engines/chronicle/engine.py
@@ -305,12 +305,12 @@ def _apply_version_scoring(
             except (TypeError, ValueError):
                 version = None
 
-        # Group key: first entity_ref, or title, or document_id
+        # Group key: first entity_ref, or document_id
         entity_refs = custom.get("entity_refs")
         if entity_refs and isinstance(entity_refs, list) and len(entity_refs) > 0:
             group_key = str(entity_refs[0])
         else:
-            group_key = custom.get("title", str(chunk.document_id))
+            group_key = str(chunk.document_id)
 
         chunk_meta.append((group_key, version))
 
--- a/src/khora/engines/vectorcypher/retriever.py
+++ b/src/khora/engines/vectorcypher/retriever.py
@@ -1667,7 +1667,7 @@ class VectorCypherRetriever:
         if not fused_results:
             return fused_results
 
-        from khora.query.reranking import CrossEncoderReranker, RerankCandidate
+        from khora.query.reranking import CrossEncoderReranker, RerankCandidate, hydrate_doc_titles
 
         top_n = min(self._config.reranking_top_n, len(fused_results))
         candidates_to_rerank = fused_results[:top_n]
@@ -1705,6 +1705,9 @@ class VectorCypherRetriever:
                     doc_title="",
                 )
             )
+        await hydrate_doc_titles(
+            candidates, self._storage, lambda fr: getattr(getattr(fr, "item", None), "document_id", None)
+        )
 
         try:
             if self._reranker is None:
@@ -1785,7 +1788,7 @@ class VectorCypherRetriever:
         if not fused_results:
             return fused_results
 
-        from khora.query.reranking import LLMReranker, RerankCandidate
+        from khora.query.reranking import LLMReranker, RerankCandidate, hydrate_doc_titles
 
         top_n = min(self._config.llm_reranking_top_n, len(fused_results))
         candidates_to_rerank = fused_results[:top_n]
@@ -1823,6 +1826,9 @@ class VectorCypherRetriever:
                     doc_title="",
                 )
             )
+        await hydrate_doc_titles(
+            candidates, self._storage, lambda fr: getattr(getattr(fr, "item", None), "document_id", None)
+        )
 
         try:
             if self._llm_reranker is None:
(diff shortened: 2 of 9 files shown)


filippetrovic:

Reviewers: security · performance · correctness · tests · docs · AI · observability · devil's advocate

### Overall Assessment: ❌ REQUEST CHANGES

The typed-field refactor itself is mechanically correct and well-tested, but the production call sites pass `doc_title=""` everywhere. Combined with the ingest-time removal of `chunk.metadata["title"]`, this means the reranker sees **no title at all** in production — a silent functional regression.

---

### HIGH

**H1. [Correctness · Observability · Devil's Advocate · Security (functional)] `doc_title` is never populated in production**

- `src/khora/query/engine.py:1339, 2913` — both `RerankCandidate(...)` sites pass `doc_title=""`.
- `src/khora/engines/vectorcypher/retriever.py:1705, 1823` — both sites pass `doc_title=""`.
- The reranker reads from the typed field correctly (line 183, 356), but it's empty every time.
- Combined with the removal of `chunk.metadata.setdefault("title", doc_title)` in ingest, titles no longer reach the reranker through any path.

**Impact:** Cross-encoder and LLM listwise rerankers lose the `[title] content` prefix that they used in production. Ranking quality regresses silently.

**Fix:** Thread `doc_title` from the parent `Document.title` at each construction site. Either (a) fetch documents alongside chunks before the rerank stage and look up `documents_by_id[chunk.document_id].title`, or (b) preserve the ingest-time title denormalization until a recall-side title hydration path exists.

---

### MEDIUM

**M1. [Devil's Advocate] Chronicle group-key collapse is correct mechanically but unjustified in the PR body**

- Old fallback chain: `entity_refs[0] → custom["title"] → document_id`.
- New: `entity_refs[0] → document_id`.
- Chunks without `entity_refs` that previously grouped by title now group by document_id (broader grouping, less granular version demotion when titles cross document boundaries).
- The PR body says "same-doc chunks group identically," which is correct for the most common case but doesn't address the cross-doc-same-title case the old fallback handled. State the assumption explicitly or accept the change as intentional simplification.

---

### LOW

**L1. [Tests] Missing edge-case tests for title content**
- No coverage for whitespace-only / very long / special-character titles (e.g., titles containing `[`, `]`, newlines).
- The cross-encoder and LLM rerankers prepend titles unescaped: `f"[{doc_title}] {content}"`. AI reviewer verified bracket-nesting cannot break LLM prompt structure, but no tests pin this behavior.

**L2. [Docs] No CHANGELOG entry**
- Suggested: a `### Changed` entry noting that `RerankCandidate.doc_title` is now the source of truth and `metadata["title"]` is no longer consulted.

**L3. [Devil's Advocate] Contract inconsistency**
- `RerankCandidate.metadata` still carries date attributes (`occurred_at`, `sent_at`) read by `_date_prefix_for`. The "typed field" treatment was applied to `title` only, not to other reranker-consumed metadata. Not blocking — flagged for future consistency.

---

### What the reviewers confirmed is good

- **Performance:** Hot-path dict lookup replaced by attribute access — measurable improvement on the rerank inner loop. No regressions.
- **Security:** No new attack surface. Stale metadata-title lookup is now explicitly ignored (small data-flow hardening).
- **AI:** Prompt format is robust against special characters via `.format(...)` (no f-string on user-controlled values). Empty-title path degrades gracefully.
- **Correctness:** Both rerankers correctly read the typed field with `or ""` null-safety. Ingest cleanup and chronicle change are internally consistent.
- **Tests:** Regression guards (stale `metadata["title"]` ignored when `doc_title=""`) are strong sentinels against the old fallback creeping back.
- **Docs:** Docstrings on `CrossEncoderReranker`, `LLMReranker`, and chronicle `_apply_version_scoring` are consistent with new behavior.
- **Observability:** Pipeline spans unchanged; no telemetry contract drift.

---

### Recommended next step

Restore title flow at the recall site (per the PR's intent of "title flows through candidate field at recall time"). The current state — typed field exists, but is never populated — is strictly worse than the prior `metadata["title"]` path because it removes the title without putting anything back. Re-open implementation to populate `doc_title` from the parent `Document.title`, then re-test the reranker prefix path end-to-end.

probablyrobot (approved):

PR #760, opened 2026-05-19
Khora.recall(): return RecallResult with documents[] always populated

filippetrovic:
## Summary

Aligns `Khora.recall()` with the typed `RecallResult` projection landed in the prior step, and closes the remaining gaps so the response shape is contractually complete.

### API change
- `Khora.recall()` no longer accepts the `include_sources` kwarg. Passing it raises `TypeError`. The 4 entity-read methods (`get_entity`, `list_entities`, `find_related_entities`, `search_entities`) still accept `include_sources` — that surface is unchanged.

### Storage layer
- New typed primitive on the `RelationalBackend` protocol: `get_document_projections_batch(doc_ids) -> dict[UUID, DocumentProjection]`. Returns the wider 11-field shape (`id`, `created_at`, `source_type`, `title`, `external_id`, `source`, `source_name`, `source_url`, `content_type`, `source_timestamp`, `metadata`) — vs. the narrower `DocumentSource` returned by `get_document_sources_batch`, which the entity-attribution path still uses.
- Implemented on `postgresql`, `sqlite`, `sqlite_lance`, and `surrealdb` backends with matching SELECT lists.
- `StorageCoordinator.get_document_projections_batch` passthrough mirrors the existing source-batch passthrough.

### Post-engine upgrade pass (`Khora._upgrade_recall_documents`)
- After the engine returns, the recall pass batch-fetches full `DocumentProjection`s for every doc id referenced by chunks, entities, or relationships in the result. Engine stubs are replaced with the full projection where the relational store resolves them.
- **Referential integrity invariant:** every `chunks[i].document_id`, every id in `entities[i].source_document_ids`, and every id in `relationships[i].source_document_ids` resolves to a `documents[j].id` after the pass — unresolvable ids get a minimal stub appended so consumers can always do `{d.id: d for d in result.documents}[chunk.document_id]` without `KeyError`.
- **`connected_entity_ids` population:** inverted from `RecallEntity.source_chunk_ids` (the canonical chunk → entity mapping produced by extraction). Empty list semantics: "no linkage in this result", not "no edges in the graph". Graph-less stacks (skeleton, chronicle without entity hits) leave every chunk with the default empty list.
- **Fail-open:** if `get_document_projections_batch` raises, the engine stubs are preserved, `engine_info["document_upgrade_failed"]` is set to the exception class+message, and a warning is logged. The recall does not crash.

### Telemetry
- New internal counter `khora.recall.dangling_ref` with a single label `referrer` ∈ `{"chunk", "entity", "relationship"}`. **No `namespace_id` label** — per the cardinality rule that namespace_id is a span attribute and log field only.
- Telemetry contract JSON updated with `stability: internal`.

### Tests
- 10 new tests in `tests/unit/test_khora_recall_upgrade.py`:
  - `include_sources` kwarg raises `TypeError` on `recall()`; entity-read methods still accept it.
  - `documents[]` populated unconditionally as the union of all chunk / entity / relationship doc references.
  - Referential-integrity stub appended for unresolvable doc ids; dangling-ref counter increments with `referrer="chunk"` exactly once.
  - Fail-open path: storage exception → `engine_info["document_upgrade_failed"]` set, warning logged, no crash.
  - `connected_entity_ids` inversion produces score-sorted order matching `result.entities`.
  - UUID coercion at the projection boundary (str → UUID).
  - Cardinality regression: dangling-ref counter never emits a `namespace_id` attribute across all three referrer kinds.
  - `sqlite` backend round-trip on `get_document_projections_batch` covers all 11 fields.
- Integration smoke tests for the new primitive on `postgresql` and `surrealdb` backends.
- Existing recall tests updated for the new shape (no `include_sources`).

## Test plan
- [ ] Unit tests pass — `uv run pytest tests/unit -q --no-cov --ignore=tests/unit/integrations/langgraph --ignore=tests/unit/integrations/google_adk` (5648 passed locally; langgraph/google_adk are env-dependent and fail identically on main).
- [ ] Integration tests pass in CI.
- [ ] `make format` and `make lint` clean.
- [ ] Manual verification: `await kb.recall(...)` returns `RecallResult` with `documents[]` populated for every referenced doc id; passing `include_sources` raises `TypeError`.

### DIFF
--- a/docs/api-reference.md
+++ b/docs/api-reference.md
@@ -173,7 +173,6 @@ result: RecallResult = await kb.recall(
     min_similarity: float = 0.0,
     agentic: bool = False,
     raw: bool = False,
-    include_sources: bool = False,
     start_time: datetime | None = None,
     end_time: datetime | None = None,
 )
--- a/docs/engines/vectorcypher-engine.md
+++ b/docs/engines/vectorcypher-engine.md
@@ -120,19 +120,16 @@ This rich context enables downstream consumers (e.g., LLM chat) to answer questi
 
 ### Source Document Population
 
-All read methods (`recall()`, `get_entity()`, `list_entities()`, `find_related_entities()`, `search_entities()`) accept `include_sources: bool = False`:
+`recall()` always returns a `RecallResult` whose `documents` list holds full `DocumentProjection` rows for every document referenced by a chunk, entity, or relationship in the result. Khora batch-fetches `DocumentSource` metadata after the engine returns (chunked at 1,000 IDs) and replaces the engine's lightweight stubs in place.
 
 ```python
-# Default: no source metadata (zero overhead)
-results = await kb.recall("query")
-
-# With sources: populates source_document metadata on chunks/entities
-results = await kb.recall("query", include_sources=True)
-for chunk, score in results.chunks:
-    print(chunk.source_document)  # DocumentSource with title, source, etc.
+result = await kb.recall("query", namespace=ns_id)
+docs_by_id = {d.id: d for d in result.documents}
+for chunk in result.chunks:
+    print(docs_by_id[chunk.document_id].title)
 ```
 
-When `False` (default), no extra query runs. When `True`, `_populate_sources()` batch-fetches `DocumentSource` metadata (chunked at 1,000 IDs) and populates `chunk.source_document`, `entity.source_documents`, and `relationship.source_documents` in-place.
+Entity-read methods (`get_entity()`, `list_entities()`, `find_related_entities()`, `search_entities()`) still accept `include_sources: bool = False` to opt-in to per-entity `source_documents` population.
 
 ### VectorCypherRetriever (`src/khora/engines/vectorcypher/retriever.py`)
 
--- a/docs/telemetry-contract.json
+++ b/docs/telemetry-contract.json
@@ -264,6 +264,11 @@
       "owner": "khora"
     },
     {
+      "name": "khora.recall.document_upgrade",
+      "stability": "internal",
+      "owner": "khora"
+    },
+    {
       "name": "khora.remember",
       "stability": "public",
       "owner": "khora"
@@ -975,6 +980,17 @@
       "histogram_buckets": [0, 1, 7, 30, 90, 365, 3650],
       "labels": [],
       "_comment": "Issue #567 Phase A. Log-bucketed: hour-day-week-month-quarter-year-decade. No labels — global recency-skew signal."
+    },
+    {
+      "name": "khora.recall.dangling_ref",
+      "type": "counter",
+      "stability": "internal",
+      "unit": "1",
+      "description": "Recall referenced a document_id the relational store could not resolve (deleted, namespace mismatch, or replication lag). Counted by referrer kind.",
+      "labels": [
+        {"name": "referrer", "values": ["chunk", "entity", "relationship"], "_comment": "Which part of the RecallResult held the dangling reference."}
+      ],
+      "_comment": "NO namespace_id label — cardinality rule (CLAUDE.md). namespace_id lives on the surrounding span and debug log."
     },
     {
       "name": "khora.ingest.source_timestamp.fallback_count",
--- a/src/khora/core/models/recall.py
+++ b/src/khora/core/models/recall.py
@@ -51,8 +51,16 @@ class RecallChunk:
     own dedicated column. The chunker self-identification contract
     requires at minimum ``{"chunker": "<name>"}``.
 
-    ``occurred_at`` and ``connected_entity_ids`` are engine-populated
-    typed fields.
+    ``occurred_at`` is engine-populated.
+
+    ``connected_entity_ids`` is the set of entities the recall pipeline
+    linked to this chunk, derived by inverting
+    ``RecallEntity.source_chunk_ids`` after the engine returns. Empty
+    list semantics: **unknown**, not "no edges" — engines that return no
+    entities (skeleton, chronicle without entity hits, graph-less
+    stacks) leave this empty even when the underlying graph backend
+    might know edges to entities that simply weren't returned in this
+    result.
     """
 
     id: UUID
--- a/src/khora/khora.py
+++ b/src/khora/khora.py
@@ -24,6 +24,16 @@ from khora.config import KhoraConfig, load_config
 from khora.core.models import Chunk, Document, Entity, MemoryNamespace
 from khora.query import SearchMode
 from khora.telemetry import bounded_text_hash, trace_span
+from khora.telemetry.metrics import metric_counter
+
+# Module-level counter — a dangling ref means an engine emitted a chunk /
+# entity / relationship pointing at a document_id that the relational store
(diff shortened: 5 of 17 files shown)


filippetrovic:
## Multi-perspective review

8 reviewers ran in parallel against this PR. Consolidated findings below, severity-classified. No HIGH-severity findings that block merge — all flagged HIGHs are either pre-existing patterns inherited from sibling code or test-coverage gaps for which the functional path is already covered.

Reviewer team: Security, Performance, Correctness, Test, Database, Observability, Docs, Devil's Advocate.

---

### HIGH

None that block merge. The following were flagged HIGH by individual reviewers but reclassified after cross-checking:

- **H1 — [Database]** `get_document_projections_batch` does not filter by `namespace_id` (postgresql.py:658, sqlite.py:620, sqlite_lance/relational.py:618, surrealdb/relational.py:586). **Reclassified MEDIUM:** the existing `get_document_sources_batch` has the same pattern (postgresql.py:619) and `documents.id` is a solo PK (not composite with namespace_id), so this is an architectural defense-in-depth concern inherited by this PR, not a regression. In practice the recall pipeline only feeds doc ids that came from the same namespace's chunks/entities/relationships, so the attack surface requires bypassing the storage protocol directly. **Follow-up:** widen both primitives' signatures to take `namespace_id` and add `AND namespace_id = $ns` in a separate PR.

- **H2 — [Test]** OTel meter rebinding pattern in `metric_reader` fixture risks cross-test interference. **Reclassified LOW:** pytest's monkeypatch fixture auto-restores at test exit; unit tests run serially by default. Pattern is identical to the existing `tests/unit/telemetry/test_otel_parity.py`.

- **H3 — [Test]** 1000-batch chunking loop boundary not exercised by tests (all current cases use ≤4 doc ids). **Reclassified LOW:** the chunking is standard `range(0, n, 1000)` — well-trodden Python; the per-batch storage call is exercised. A boundary test is nice-to-have, not blocking.

- **H4 — [Test]** No test covering the partial-resolution case (engine emits stub for A,B; storage returns only A). **Reclassified MEDIUM:** functional path is correct (`unresolved = doc_ids - set(projections.keys())` is straightforward), but the test gap is real.

### MEDIUM

- **M1 — [Database/Security]** See H1 above — namespace_id should be a parameter on `get_document_projections_batch`. Follow-up PR.
- **M2 — [Observability]** Backend `get_document_projections_batch` implementations and the coordinator passthrough lack `@_record_storage_op` instrumentation that peer batch methods use (`get_document_sources_batch` and other `get_*_batch` methods). Adds latency/batch-size visibility on the hot recall path.
- **M3 — [Performance]** Dangling-ref counting (khora.py:2234-2236) does three full iterations of chunks/entities/relationships in the failure path. At catastrophic-failure scale (e.g., 50% docs unresolved across 10k chunks), this is O(n·m). Acceptable in the golden path, optimize only if dangling refs become endemic.
- **M4 — [Devil's Advocate/Reliability]** No timeout on the projection batch fetch — if the relational backend stalls, recall hangs too. Suggest `asyncio.timeout(2.0)` wrap on the fetch loop in a follow-up. Fail-open path catches the exception but only after the underlying driver gives up.
- **M5 — [Test]** Stub-vs-storage preference order (storage always wins via `projections.get(stub.id, stub)`) is undocumented and untested. If an engine intentionally sets `source_type="ephemeral"` and storage returns `"library"`, storage silently wins. Document or test the precedence.
- **M6 — [Test]** Fail-open path test (`TestFailOpenOnStorageFailure`) doesn't assert `chunks[*].connected_entity_ids` is still populated when the upgrade pass fails — only validates `engine_info["document_upgrade_failed"]` is set. Add assertion.
- **M7 — [Correctness]** `connected_entity_ids` ordering assumes `result.entities` is score-sorted by the engine; not enforced or asserted at runtime. Document the contract on `RecallChunk.connected_entity_ids` (the docstring already mentions ordering — make it a contract statement).

### LOW

- **L1 — [Docs/Code]** Misleading legacy comment near khora.py:1700 references API stability for `include_sources` — the kwarg was never on `recall()` to begin with. Drop the comment.
- **L2 — [Docs]** SurrealDB `get_document_projections_batch` docstring is one-liner; PostgreSQL/sqlite_lance use two-line versions explaining the wider SELECT. Cosmetic consistency.
- **L3 — [Correctness/Robustness]** SQLite + sqlite_lance `created_at` fallback uses `datetime.now(UTC)` if NULL (sqlite.py:633). `created_at` is NOT NULL in the schema so this path is theoretically unreachable; consider a fixed sentinel for clarity.
- **L4 — [Correctness]** SQLite metadata JSON parse catches both `ValueError` and `TypeError`. `json.loads` only raises `ValueError`/`JSONDecodeError`; `TypeError` catch is dead code.
- **L5 — [Correctness]** No enum or literal type pinning the three `referrer` label values (`chunk` / `entity` / `relationship`). Currently hardcoded strings in code; contract pins them in `docs/telemetry-contract.json`. Drift risk is low but a `Literal["chunk", "entity", "relationship"]` type alias would help.
- **L6 — [Test]** Counter cardinality test asserts label presence but not value aggregation (e.g., 5 dangling chunks → one data point with `value=5`). Edge polish.
- **L7 — [Devil's Advocate]** Stubs created for dangling references get `source_type="library"` — same as resolved documents. Consumers can't distinguish "real document with no title" from "dangling reference". Document the semantics or use a distinct `source_type` for stubs.
- **L8 — [Devil's Advocate]** ~80% code duplication across the 4 backends' `get_document_projections_batch` implementations. Considered acceptable given the semantic split with `get_document_sources_batch` — but a CI check that compares the two primitives' column sets across backends would prevent silent drift.

---

### Positive findings

- Frozen-dataclass discipline preserved throughout — every update goes through `dataclasses.replace()`, no in-place mutation.
- Telemetry contract entry (`khora.recall.dangling_ref`) is correctly tagged `stability: internal` and has no `namespace_id` label per the cardinality rule. Cardinality regression test (`test_no_namespace_id_on_dangling_ref_counter`) asserts the rule.
- SQL parameterization is correct on all four backends (PostgreSQL ORM, SQLite explicit `?`, sqlite_lance ORM, SurrealDB named `$ids`).
- Referential-integrity invariant holds across all three edge cases verified (empty result, all-stub, mix of resolved/unresolved).
- UUID coercion at the projection boundary is comprehensive — every id-carrying field across the 4 backends.
- Fail-open path is observable via `engine_info["document_upgrade_failed"]` + warning log.


### Overall assessment

**APPROVE.** The PR is production-ready. No HIGH-severity blockers after cross-checking. Recommended follow-ups (in priority order):

1. Add `namespace_id` parameter to both `get_document_projections_batch` and `get_document_sources_batch` for defense-in-depth (M1).
2. Add `@_record_storage_op` to the new backend methods + coordinator passthrough (M2).
3. Wrap the projection batch fetch in `asyncio.timeout(2.0)` (M4).
4. Add the three missing test cases (partial-resolution dangling refs, 1000-boundary chunking, fail-open with inversion assertion) (H4/H3/M6).

probablyrobot (approved):

PR #779, opened 2026-05-20
Add explicit source_timestamp kwarg to remember/remember_batch/submit_batch

filippetrovic:
## Summary

- Adds `source_timestamp: datetime | None = None` as a first-class kwarg on `Khora.remember()`, `Khora.remember_batch()`, and `Khora.submit_batch()`, mirroring the existing `source_type` / `source_name` / `source_url` pattern.
- When the kwarg is provided, it persists directly to `Document.source_timestamp`. When omitted, the existing metadata-derived fallback (`_extract_source_timestamp` on `sent_at` / `occurred_at` / `created_at` / ...) is preserved for backward compatibility, so existing callers see no behavior change.
- Threaded through `MemoryEngineProtocol`, `VectorCypherEngine`, `ChronicleEngine`, `SkeletonConstructionEngine`, and the ingest pipeline so all three public entry points land the kwarg-supplied timestamp on the persisted `DocumentModel`.
- The `google_adk` adapter now forwards the parsed ADK-event timestamp via the new kwarg instead of stuffing an ISO string in a helper metadata key (the prior `kwargs.pop("source_timestamp_iso", None)` strip is gone).
- `remember_batch()` / `submit_batch()` also accept `source_timestamp` as a per-doc dict key — the per-doc value wins over the top-level kwarg, matching the existing source-trio precedence.
- CHANGELOG entry added under the upcoming-release section.

## Test plan

- [x] `make format` clean
- [x] `make lint` passes (existing `ty` warnings unchanged)
- [x] 10 new unit tests in `tests/unit/test_khora.py` covering the four scenarios from the spec — kwarg explicit, per-doc dict variant, metadata fallback, and kwarg-wins-over-metadata — all green
- [x] Affected ingest-pipeline tests (`tests/unit/test_pipelines_ingest.py`) still pass — no regression in the fallback path
- [x] `google_adk` unit tests still pass
- [ ] Integration / e2e suites — verified via CI

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@ Format: versions match git tags (`git tag vX.Y.Z`). Versions before 0.5.1 were i
 
 ## [Unreleased]
 
+### Added
+
+- **Explicit `source_timestamp` kwarg on `Khora.remember()`, `Khora.remember_batch()`, and `Khora.submit_batch()`.** Callers can now pass `source_timestamp: datetime | None` directly instead of relying on the metadata-derived fallback. Per-doc dict key (`source_timestamp`) on `remember_batch` / `submit_batch` overrides the top-level kwarg for that document, mirroring the existing `source_type` / `source_name` / `source_url` pattern. When the kwarg is provided, it wins over the metadata-based fallback (`sent_at` / `occurred_at` / `created_at` / ...); when omitted, the existing fallback in `_extract_source_timestamp` is preserved unchanged.
+
 ### Security
 
 - **CI security gate allowlist for unfixable upstream CVEs.** pip-audit
--- a/src/khora/engines/chronicle/engine.py
+++ b/src/khora/engines/chronicle/engine.py
@@ -1027,6 +1027,7 @@ class ChronicleEngine:
         source_type: str = "library",
         source_name: str | None = None,
         source_url: str | None = None,
+        source_timestamp: datetime | None = None,
         metadata: dict[str, Any] | None = None,
         skill_name: str = "general_entities",
         entity_types: list[str],
@@ -1095,6 +1096,7 @@ class ChronicleEngine:
             source_type=source_type,
             source_name=source_name or None,
             source_url=source_url or None,
+            source_timestamp=source_timestamp,
             checksum=checksum,
             size_bytes=len(content.encode("utf-8")),
             metadata=dict(metadata or {}),
@@ -2295,6 +2297,7 @@ class ChronicleEngine:
         source_type: str = "library",
         source_name: str | None = None,
         source_url: str | None = None,
+        source_timestamp: datetime | None = None,
         extraction_batch_size: int | None = None,
         extraction_max_tokens: int | None = None,
     ) -> BatchResult:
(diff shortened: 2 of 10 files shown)


filippetrovic:
## Multi-perspective Review — PR #779

Consolidated findings from a 6-reviewer team (Security, Performance, Correctness, Tests, Docs, Devil's Advocate). The PR adds `source_timestamp` as an explicit kwarg on `Khora.remember()` / `remember_batch()` / `submit_batch()`, threaded through the engine layer and ingest pipeline.

### Overall: ✅ Approve on CI green

The implementation is well-factored, backward-compatible, and well-tested. Findings below are non-blocking nice-to-haves.

---

### HIGH

_(none)_

### MEDIUM

- **M1. [Tests]** `tests/unit/test_khora.py` — Existing unit tests mock the engine or capture `Document` pre-persistence; they do not exercise the JSON-serialization boundary where the recent `TypeError: Object of type datetime is not JSON serializable` regression surfaced (fixed by dropping the redundant `metadata["source_timestamp"]` stuffing in commit ahead of HEAD). A small integration-level regression test asserting that `Khora.remember(...)` with `metadata={"custom": <non-serializable>}` doesn't break SQLAlchemy JSON encoding would prevent a similar foot-gun in the future.

### LOW

- **L1. [Security]** `src/khora/khora.py:1020-1021`, `src/khora/pipelines/flows/ingest.py:420-422` — The `source_timestamp: datetime | None` parameter accepts timezone-naive datetimes silently. UTC is the established convention but no normalization happens at the boundary. Consider documenting "pass UTC-aware datetime" or normalising naive → UTC in `Khora.remember()`. Low risk because the existing fallback path uses `datetime.now(UTC)` which is timezone-aware.

- **L2. [Docs]** `src/khora/pipelines/flows/ingest.py` (3 sites at `stage_document` / two `_create_one` paths) — Same explicit-over-fallback comment is repeated three times. A docstring on `_extract_source_timestamp` covering the precedence rule would let the three call sites drop the duplicated comment.

- **L3. [Docs]** `src/khora/integrations/google_adk/_mapping.py` and `memory_service.py` — The migration from a helper `source_timestamp_iso` metadata key to the new direct kwarg is correct but undocumented. A one-line comment in the mapping function noting the path change would help future readers.

- **L4. [Design — devil's advocate]** Parameter-bloat trajectory — `Khora.remember()` is now at ~16 kwargs, four of which are provenance-shaped (`source_type` / `source_name` / `source_url` / `source_timestamp`). Future provenance additions should probably consolidate into a `Provenance` dataclass rather than continuing to add positional/keyword arguments. Out of scope for this PR; flagged for future planning.

- **L5. [Design — devil's advocate]** Subtle behavior shift not flagged in CHANGELOG: callers who supply both the new `source_timestamp` kwarg AND a conflicting `metadata["created_at"]` will now see the kwarg win. This is the intended precedence and is documented in the docstrings, but a one-line callout in the CHANGELOG ("explicit kwarg now overrides metadata-derived fallback when both are supplied") would surface it for upgraders.

---

### Per-reviewer summary

| Reviewer | Verdict |
|---|---|
| Security | ✅ No HIGH / MEDIUM. One LOW (timezone-naive acceptance). |
| Performance | ✅ LOW risk. Additive only; no algorithmic change; <1µs overhead. |
| Correctness | ✅ All four engines wire the kwarg correctly; precedence and fallback semantics preserved. Zero placeholders. |
| Tests | ✅ 10 new unit tests cover the four spec scenarios across all three entry points. One MEDIUM gap on integration-boundary coverage (see M1). |
| Docs | ✅ Docstrings on all three public methods and on `MemoryEngineProtocol`; CHANGELOG accurate. Two LOW gaps on internal-clarity comments. |
| Devil's Advocate | Raises design questions about parameter arity and necessity. None block merge but worth tracking. |

### Recommendation

Merge after CI is green. Address M1 in a follow-up; LOW items at author's discretion.

probablyrobot (approved):

PR #800, opened 2026-05-21
fix(vectorcypher): persist chunker_info end-to-end

filippetrovic:
## Summary

On the vectorcypher engine path, the chunker self-identification dict (`{"chunker": "<strategy>", ...}` stamped by every chunker into `ChunkResult.metadata`) was dropped between the chunker and the temporal store, and re-defaulted to `{}` on read. Skeleton was already correct via the regular `chunks` table. This PR wires `chunker_info` end-to-end on the `khora_chunks` table so `recall().chunks[i].chunker_info` surfaces the real dict.

- `TemporalChunk` gains a `chunker_info: dict[str, Any]` field next to `metadata`.
- `khora_chunks` table definition (pgvector), DDL (sqlite_lance), and surrealdb temporal-store schema all carry the new column.
- New Alembic migration `038_khora_chunks_chunker_info` adds the column for production deployments where the table predates it. The migration skips on dialects where `khora_chunks` does not exist (fresh deploys, sqlite_lance fixture) — `metadata.create_all` / runtime DDL lands the column from the updated table definition in those cases.
- Vectorcypher engine: every `TemporalChunk(...)` site (single-doc, replace-doc, batch) now passes `chunker_info=dict(raw_chunk.metadata)`.
- Vectorcypher retriever: every `Chunk(...)` site projects `chunker_info` from its supplying record — Neo4j sites read the JSON string we stamp; `TemporalChunk` sites read the field directly. Cypher RETURN clauses for the Neo4j-record sites now include `c.chunker_info`.
- Neo4j Chunk-node writes serialize `chunker_info` as a JSON string in `dual_nodes.create_chunk_nodes_batch`.

## Test plan

- [x] `uv run pytest tests/unit/engines/vectorcypher/test_chunker_info_propagation.py` — 3 tests pinning the engine wiring (single chunk, multi-chunk independent dicts, empty-metadata fallback).
- [x] `uv run pytest tests/integration/db/test_migration_032_dream_runs.py` — migration-chain test updated to expect `038_khora_chunks_chunker_info` as head.
- [ ] New integration test `tests/integration/matrix/test_vectorcypher_chunker_info.py` runs under the matrix stack — gated on CI's matrix job.
- [ ] CI green (lint, format, test).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@ Format: versions match git tags (`git tag vX.Y.Z`). Versions before 0.5.1 were i
 
 ## [Unreleased]
 
+### Fixed
+
+- vectorcypher: persist `chunker_info` end-to-end so recall surfaces the chunker self-identification dict instead of `{}`.
+
 ## [0.16.2] - Fork-safe integrations + vectorcypher proxy fix + Hermes adapter
 
 Patch release on top of v0.16.1. Two production-impacting fixes (fork-safe integration globals; `Khora.recall()` `AttributeError` on session-less corpora introduced by the v0.16.0 namespace proxy), plus the Hermes memory-provider adapter and the env-var naming consolidation that already landed on `main` since v0.16.1.
--- a/src/khora/db/migrations/versions/038_khora_chunks_chunker_info.py
+++ b/src/khora/db/migrations/versions/038_khora_chunks_chunker_info.py
@@ -97,8 +97,8 @@ def _assert_pg_version_ok() -> None:
         )
 
 
-def _upgrade_impl() -> bool:
-    """Perform the schema change. Returns True if the column was added."""
+def _upgrade_impl() -> None:
+    """Perform the schema change."""
     bind = op.get_bind()
     inspector = sa.inspect(bind)
     if not inspector.has_table("khora_chunks"):
@@ -111,7 +111,7 @@ def _upgrade_impl() -> bool:
         # the updated ``khora_chunks_table`` definition. The migration is
         # still required for existing production deployments where
         # ``khora_chunks`` was created before this column existed.
-        return False
+        return
 
     is_pg = _is_postgres()
 
@@ -144,21 +144,18 @@ def _upgrade_impl() -> bool:
                 server_default=sa.text("'{}'"),
             ),
         )
-    return True
 
 
 def upgrade() -> None:
     start = time.monotonic()
-    applied = False
     try:
-        applied = _upgrade_impl()
+        _upgrade_impl()
     except OperationalError as exc:
         duration_ms = int((time.monotonic() - start) * 1000)
         logger.bind(
             migration_id=revision,
             duration_ms=duration_ms,
             lock_timeout_tripped=_is_lock_timeout(exc),
-            applied=False,
         ).error("khora.migration.applied")
         # Bare ``raise`` re-raises the active exception with the original
         # traceback preserved. env.py's wrapping context.begin_transaction()
@@ -170,7 +167,6 @@ def upgrade() -> None:
         migration_id=revision,
         duration_ms=duration_ms,
         lock_timeout_tripped=False,
-        applied=applied,
     ).info("khora.migration.applied")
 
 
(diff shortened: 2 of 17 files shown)


filippetrovic:
## Multi-perspective review — PR #800

Eight reviewers (security, performance, correctness, test, database, observability, docs, devil's advocate) audited the diff independently. Below is the consolidated set of findings with severity classification (HIGH must fix, MEDIUM should fix, LOW optional).

**Overall:** All reviewers landed on **APPROVE** modulo one HIGH (telemetry contract drift) and a handful of MEDIUM follow-ups. The core wiring is correct end-to-end.

### HIGH

- **H1 — [Observability]** `src/khora/db/migrations/versions/038_khora_chunks_chunker_info.py:161,173` — migration 038 emits `khora.migration.applied` with a new `applied: bool` field that is not declared in `docs/telemetry-contract.json`. Migration 037 emits `{migration_id, duration_ms, lock_timeout_tripped, source_name_backfilled, source_name_unmatched}` only. The structured-log shape is consumed by dashboards; new fields should land in the contract in the same PR. Either drop the field (the success/error branch already distinguishes applied=False) or document it.

### MEDIUM

- **M1 — [Security]** `src/khora/engines/vectorcypher/retriever.py:807,2534` — the two `json.loads(c_data["chunker_info"])` sites have no try/except. A corrupted persisted JSON string (direct DB tampering, partial migration) would raise `json.JSONDecodeError` and crash `recall()`. The isinstance gate filters out non-string values but not malformed strings. Wrap with `try/except json.JSONDecodeError` and fall back to `{}` — or document this as a hard data-integrity precondition.
- **M2 — [Test]** Two read paths are not exercised by the new tests: (a) Neo4j retriever sites (`retriever.py:807`, `retriever.py:2534`) only fire when a graph-channel query routes through `typed_entity_fast_path` / `_graph_retrieve_chunks` — the integration test uses sqlite_lance which never hits Neo4j; (b) the SurrealDB `_row_to_chunk` deserialization path is similarly unexercised. A regression that drops `c.chunker_info` from the Cypher RETURN would pass CI silently. Consider adding matrix-stack coverage in a follow-up.
- **M3 — [Performance]** Same retriever sites do a `json.loads()` per recalled chunk. Small dict (`{"chunker": "<name>"}`), sub-µs cost — but worth profiling if recall latency becomes contended.
- **M4 — [Devil's advocate / Architecture]** Push-back on the dedicated-field design. `chunker_info` could ride inside the existing `TemporalChunk.metadata` dict and be exposed via a synthetic field on `Chunk`, eliminating migration 038 + per-backend column wiring. The PR's choice (explicit JSONB column) buys schema clarity and future "filter chunks by chunker strategy" capability; the metadata-blob alternative trades that for ~100 fewer lines of diff. Worth flagging — the team should be deliberate about the trade-off.
- **M5 — [Devil's advocate]** The retriever dual-path (`json.loads` if `str` else `dict()`) implies the read side is uncertain whether the storage layer returns string or dict. Neo4j is JSON-string-serialized in `dual_nodes`, the other backends pass native dicts — so the union is real. Worth a one-line code comment naming the contract per backend.

### LOW

- **L1 — [Performance]** The new column adds ~10s-100s of bytes per chunk row on read. Acceptable; flag only because embedded (sqlite_lance) deployments accumulate it on disk.
- **L2 — [Database]** Migration 038 follows 037's pattern faithfully (PG ≥11 version gate, 5s lock_timeout, structured logging, SQLSTATE 55P03 detection). `inspector.has_table` short-circuit correctly handles fresh-deploy + sqlite_lance-fixture + existing-prod scenarios. Downgrade is idempotent. ✅
- **L3 — [Docs]** CHANGELOG entry under `## [Unreleased]` is present and accurate. No new gotcha for CLAUDE.md (this PR just brings `khora_chunks` in line with the existing `chunks` table). No `docs/` drift.
- **L4 — [Naming policy]** Verified by grep: no internal IDs (ticket/ADR/partner names) leaked into the diff.

### Overall assessment

**APPROVE with follow-ups.** H1 is the only must-fix and is a one-line contract update; M1–M5 are reasonable to address in a follow-up rather than block this PR.

probablyrobot (approved):

PR #801, opened 2026-05-21
fix(recall): surface unset optional strings as null on recall response

filippetrovic:
## Summary

- Drop the `text/plain` default on the ingest pipeline's `content_type` (three call sites) so unset content types stay `None` end-to-end.
- Hoist the existing `_none_if_empty(v) -> v if v else None` helper from nested-inside-`_document_model_to_domain` to module scope on the postgresql, sqlite_lance, and surrealdb relational backends, and apply it to `title`, `external_id`, `source`, `source_name`, `source_url`, `content_type` in `get_document_projections_batch`. Pre-existing behavior of `source_type` (defaults to `"library"`), `source_timestamp`, and `metadata` is unchanged.

This satisfies the recall-response wire contract that unset optional strings serialize as `null` rather than `""`.

## Test plan

- [x] `tests/unit/pipelines/test_ingest_coverage_push.py::TestStageDocument::test_omitted_content_type_stays_none` — green locally.
- [x] `tests/unit/pipelines/test_ingest_coverage_push.py::TestStageDocument::test_empty_content_type_normalizes_to_none` — green locally.
- [x] `tests/unit/storage/backends/sqlite_lance/test_relational.py::test_get_document_projections_batch_normalizes_empty_strings_to_none` — green locally.
- [ ] `tests/integration/storage/test_get_document_projections_batch_pg.py::TestGetDocumentProjectionsBatchPg::test_get_document_projections_batch_normalizes_empty_strings_to_none` — runs in CI (needs PG).
- [ ] `tests/integration/storage/backends/surrealdb/test_get_document_projections_batch.py::test_projections_batch_normalizes_empty_strings_to_none` — runs in CI.
- [ ] `make test` overall green in CI; coverage ≥65%.


probablyrobot (approved):

filippetrovic:
## Team Review — PR #801

Multi-perspective review (security, performance, correctness, tests, database, docs, devil's advocate) plus a dedicated security audit pass. All seven reviewers and the security audit independently report no blocking findings. Consolidated below.

### CRITICAL
- _none_

### HIGH
- _none_

### MEDIUM
- **[Devil's Advocate / Scope]** `src/khora/storage/backends/postgresql.py`, `src/khora/storage/backends/sqlite_lance/relational.py`, `src/khora/storage/backends/surrealdb/relational.py` — The PR hoists `_none_if_empty` to module scope on all three relational backends, but the helper is invoked from a single call site per file (`get_document_projections_batch`). A leaner alternative is the inline form `field or None` (the existing style in `sqlite_lance/relational.py` and `surrealdb/relational.py` before this PR), which produces identical wire output with 12 fewer lines and no new module-level name. The current implementation is consistent across the three backends and behavior-preserving, so this is a style/scope critique rather than a bug — flagging so the author can weigh consistency-via-helper vs. minimal-diff before merge.
- **[Devil's Advocate / Scope]** `src/khora/storage/backends/*.py` — The fix targets the **projection** path (`get_document_projections_batch`). The full-Document read path (`_document_model_to_domain`) already applied `_none_if_empty` to the same fields, and other read paths via `_get_one` / `get_document` are not in scope here. If the intent is "unset optional strings surface as null everywhere on the recall response," the scope is correct; if the intent was "everywhere in the system," this fix is partial. Worth confirming in the PR description.

### LOW
- **[Correctness / Helper semantics]** `_none_if_empty(v) = v if v else None` is falsy-based, not empty-string-based. For the `str | None` columns touched here, the only practical input difference is whitespace-only strings (`"   "`) which remain unchanged — matches the prior nested-helper semantics, so no regression — but worth keeping in mind for future callers.
- **[Devil's Advocate / Test structure]** Three near-identical backend projection tests across `tests/integration/storage/test_get_document_projections_batch_pg.py`, `tests/unit/storage/backends/sqlite_lance/test_relational.py`, and `tests/integration/storage/backends/surrealdb/test_get_document_projections_batch.py`. Not parametrized because the adapters don't share a base class. Matches sibling test style; structural duplication is acceptable.

### INFO
- **[Performance]** Six function-call hops per row in `get_document_projections_batch`. With max batch size 1000 and ~5-15 ns per call, worst-case overhead is < 100 µs per batch — far below the I/O latency floor. No measurable hot-path impact.
- **[Security]** No new endpoints, HTTP clients, file ops, deserializers, secrets, auth surfaces, or dependencies. New SQL in tests is parameterized end-to-end (SQLAlchemy `.update().values()` on PG/SQLite; bound `$rid`/`$blank` on SurrealDB).
- **[Database]** Migration 037 already dropped the `""` server defaults and backfilled `""` → `NULL` for these columns; no schema migration is needed in this PR. ORM models declare `nullable=True` on all six fields.
- **[Correctness]** No placeholders (TODO/FIXME/HACK/NotImplementedError/stub) in the diff.
- **[Docs]** No drift — no docstring or doc reference to the old `"text/plain"` default or `""` recall behavior. `CHANGELOG.md` correctly untouched (release branch territory).
- **[Tests]** Coverage is symmetric: every backend test asserts all six target fields on both the `""`-row and the `None`-row; both ingest cases (missing key, explicit `""`) are asserted. Mutation-style sanity: reverting the normalization would break the row-A assertions on every backend.

## Overall Assessment

**APPROVE.** The behavior change is real and necessary; the implementation is surgical and well-tested; all 8 CI jobs are green. The two MEDIUM scope/style notes from Devil's Advocate are worth weighing but are non-blocking — the current approach trades 12 lines for cross-backend symmetry of the normalization helper, which is a defensible call. The LOW/INFO items are observations only.

PR #819, opened 2026-05-22
fix(migrations): drop NOT NULL before empty-string normalization in 037

filippetrovic:
## Summary
- Reorder the `documents` nullability flip in migration `037_recall_response_format` so the `DROP NOT NULL` (Postgres) / `batch_alter_table` flip (SQLite) runs **before** the `UPDATE ... SET <col> = NULL WHERE <col> = ''` normalization on the six columns `source`, `content_type`, `title`, `author`, `language`, `checksum`.
- **Why:** on databases created via the legacy `create_tables()` / `Base.metadata.create_all` path, the old `Mapped[str]` ORM declared those columns `NOT NULL`. Writing `NULL` before dropping the constraint violated it and rolled back the entire revision, stranding the database at the prior revision. Alembic-chain-created databases already had the columns nullable, so the bug never fired there — which is why existing tests missed it.
- Add a regression test (`TestLegacyNotNullStartingState`) that rebuilds `documents` into the legacy `NOT NULL DEFAULT ''` shape, asserts the columns are `NOT NULL` pre-upgrade, then asserts the upgrade succeeds and the columns become `NULL`. Verified to raise `IntegrityError` against the un-reordered ordering.
- Clarify the module docstring to describe both possible starting states, and add a `### Fixed` CHANGELOG entry.
- Downgrade path is unchanged (it already drops to `''` before re-applying `NOT NULL`).

## Test plan
- [x] `tests/unit/db/test_migration_037_recall_response_format.py` — 21 passed (incl. the new legacy-state test)
- [x] `make format` clean; `make lint` (ruff + ty) reports no findings on the changed files
- [ ] CI: unit + integration suites green


filippetrovic:

Seven independent reviewers (security, performance, correctness, test, database, docs, devil's advocate) examined this change. **Overall assessment: APPROVE** — the reorder is correct, safe, and well-tested. No CRITICAL or HIGH code findings. Remaining items are non-blocking suggestions.

### Reviewer verdicts
| Domain | Verdict |
|--------|---------|
| Security | Clean — column names come from a static module-level tuple (no injection); no secrets; no new deps |
| Performance | Clean — `DROP NOT NULL` / `DROP DEFAULT` are metadata-only on Postgres; lock footprint unchanged |
| Correctness | Clean — reorder valid on both PG and SQLite paths; downgrade untouched and still correct; no placeholders |
| Database | Clean — backwards-compatible, idempotent on already-nullable DBs; editing-by-revision-id is sound |
| Docs | Clean — docstring + CHANGELOG accurate on both starting states |
| Test | Approve — regression test genuinely reproduces the legacy NOT NULL state and asserts it pre-upgrade |
| Devil's Advocate | Approach endorsed; raised doc-clarity + test-helper concerns below |

### MEDIUM
- **[Test/Devil's Advocate]** `tests/unit/db/test_migration_037_recall_response_format.py` — the `_make_flipped_columns_not_null` helper rewrites the table DDL via regex against the live `sqlite_master` definition. This is brittle if the DDL format ever changes (e.g. multi-line column defs). Mitigated by the `assert replaced == 1` guard (fails loudly, locally) and the pre-upgrade `PRAGMA table_info` NOT NULL assertion, so it cannot silently degrade into testing the already-nullable path. Non-blocking; a hard-coded `CREATE TABLE` snapshot would be simpler but would need manual upkeep if the prior revision changes.

### LOW
- **[Docs/Operator clarity]** Consider noting in the CHANGELOG/PR that this fix is effective only for databases created fresh or stranded at the prior revision — databases that already applied the previous version of this migration successfully will not re-run it (Alembic tracks by revision id). Communication nicety, not a code issue.
- **[Test]** Optional: add a second seeded row with mixed empty/non-empty values to exercise per-column normalization directly (the broader suite already covers this on the nullable path).
- **[Test]** Optional: add a one-line comment documenting that the DDL-rewrite regex assumes single-line column definitions and fails loudly via `replaced == 1` otherwise.
- **[Devil's Advocate]** Optional: assert column-order preservation after the table-copy in the test helper; shorten the long test method name; trim minor docstring/inline-comment duplication.

### Security audit
Focused audit of the diff: no OWASP Top-10, dependency, or cryptographic findings. The only dynamic SQL interpolates a static, code-defined column tuple (not user input); the test regex is built from a hardcoded allowlist. No secrets or PII introduced.

### Verification
Migration-focused unit tests pass (21), including the new legacy-NOT-NULL regression test (proven to fail against the un-reordered ordering). `ruff format`/`ruff check`/`ty` clean on the changed files.

msrdic (approved):

probablyrobot (approved):

PR #822, opened 2026-05-22
Emit engine_info.mode as lowercase mode string

### LINKED ISSUE #821: VectorCypher: engine_info.mode emitted as SearchMode enum integer instead of lowercase string
## Problem

`RecallResult.engine_info["mode"]` from the VectorCypher engine is emitted as the `SearchMode` enum **integer** (e.g. `3`) instead of the lowercase mode **string** (e.g. `"hybrid"`). The integer is opaque to API consumers and does not match the documented recall `mode` vocabulary (`vector` | `graph` | `hybrid` | `all`).

## Root cause

`SearchMode` is an `Enum` defined with `auto()`, so `SearchMode.HYBRID.value == 3`. The `engine_info` builder used `mode.value`, surfacing the int.

## Expected

`engine_info["mode"]` should be the lowercase mode string:
- `SearchMode.HYBRID` → `"hybrid"`
- `SearchMode.VECTOR` → `"vector"`
- `SearchMode.GRAPH` → `"graph"`
- `SearchMode.ALL` → `"all"`

filippetrovic:
## Summary

- `VectorCypherEngine` recall now emits `engine_info["mode"]` as the lowercase mode string (e.g. `hybrid`) instead of the `SearchMode` enum integer (`3`). The single write site in `engines/vectorcypher/engine.py` changes from `mode.value` to `mode.name.lower()`, producing `vector` / `graph` / `hybrid` / `all` — matching the documented recall `mode` vocabulary.
- Updated the canonical `engine_info` shape test to assert the literal `"hybrid"` (pins the contract value rather than re-deriving it from the same enum the production code uses).
- Added a `### Fixed` entry under a new `## [0.16.5]` CHANGELOG section.

## Test plan

- [x] `make format` clean; `make lint` green
- [x] `tests/unit/engines/vectorcypher/test_engine_info_canonical.py` passes (8 passed), including the lowercase-string assertion
- [x] Scope-boundary tests untouched and passing; regression sweep confirms the recall-event payload path is intentionally separate
- [ ] Full integration suite + 77% coverage gate (CI — requires Docker-backed Postgres/Neo4j)

Fixes #821


filippetrovic:
## Multi-perspective PR review (automated team)

**Overall: APPROVE.** All CI checks are green — including the full integration suite + coverage gate. The change is a surgical, well-tested fix: `VectorCypherEngine` now emits `engine_info["mode"]` as the lowercase mode string (`mode.name.lower()`) instead of the `SearchMode` enum integer, matching the documented `mode` vocabulary (`vector` | `graph` | `hybrid` | `all`). Six reviewer perspectives plus a dedicated security audit; no blocking findings.

### Reviewer summary

| Domain | Result |
|---|---|
| Security | No findings |
| Performance | No findings — `.name.lower()` runs once per recall at result construction; negligible |
| Correctness | No findings — sole `engine_info["mode"]` write site; all `SearchMode` members map correctly; no in-repo consumer relies on the int |
| Tests | Adequate — assertion pins the literal `"hybrid"`; the three scope-boundary tests are untouched |
| Docs | No drift — `engine_info["mode"]` was never documented as an int (it's free-form engine telemetry); CHANGELOG accurate |
| Devil's Advocate | Raised a release-classification consideration (M1 below) |

### MEDIUM

- **M1 — [Release classification] consider how the shipping release frames this.** `engine_info["mode"]` changes type (int → string) in a response field. Because the integer form shipped only in the immediately-prior release (as part of the canonical-`engine_info`-keys work) while the documented `mode` vocabulary has always been the lowercase strings, this is best read as fixing a one-release-old regression rather than an intentional contract change — hence the `### Fixed` framing. That said, any consumer that began parsing the integer would be affected, so the release that ships this should decide whether to keep it a patch `### Fixed` or call it out more prominently as a behavior change (and coordinate downstream consumers if needed). **Non-blocking for this PR** — version number and section finalization belong to the release step, which is intentionally out of scope here.

### LOW

- **L1 — [Tests] optional parametrization.** The canonical test exercises only `HYBRID`. Since `mode.name.lower()` is branch-free, one mode fully exercises the contract, but a parametrized check across `vector` / `graph` / `all` would harden it. Non-blocking.
- **L2 — [Release process] informational.** The CHANGELOG adds a `## [0.16.5]` section with no version-file bump; that is correct for the `hatch-vcs` (git-tag-sourced) versioning model — the tag is pushed after merge, so `0.16.5` does not exist until then.

## Security Audit

Scanned added/modified lines only (3 files: the one-line `engine_info["mode"]` change, the test assertion, the CHANGELOG entry). No injection, authn/authz, secret-handling, SSRF, file/path, deserialization, regex/ReDoS, or dependency surface is touched. **Verdict: PASS** — no findings.

probablyrobot (approved):

PR #823, opened 2026-05-22
Coerce string source_timestamp to datetime and redact document payloads from ingestion failure logs

filippetrovic:
## Summary

Two defensive fixes on the document-ingestion boundary.

**1. Coerce string `source_timestamp` → `datetime` before it reaches `Document`.**
The public `source_timestamp` kwarg on `remember` / `remember_batch` / `submit_batch` is typed `datetime | None`, but `Document` is a non-validating `@dataclass`, so when an upstream connector or adapter passes an ISO-8601 **string** it flowed through unparsed and crashed `asyncpg` on the `TIMESTAMP WITH TIME ZONE` bind:

```
asyncpg ... DataError: invalid input for query argument $N
(source_timestamp, TIMESTAMP WITH TIME ZONE): expected a datetime.date or
datetime.datetime instance, got 'str'
```

The metadata-derived path already parsed ISO strings; the explicit-kwarg / per-doc-dict path did not. This adds a shared `coerce_source_timestamp(value)` helper that returns an existing `datetime` unchanged, parses ISO-8601 strings (trailing `Z`, explicit offset, and date-only `YYYY-MM-DD`), and returns `None` for empty/unparseable input without raising. `_extract_source_timestamp` now delegates to it (single parsing path). The helper is applied at every `Document(...)` construction boundary that takes an explicit or per-doc `source_timestamp`: the three ingest staging paths, the `submit_batch` insert and re-queue/update paths, and the `remember` / `remember_batch` kwarg + per-doc values (coerced before forwarding so engine-built Documents are covered too).

**2. Stop logging document payloads on persist failure.**
Several failure paths interpolated the raw `{exc}` into a log line or surfaced `str(exc)` as `DocumentResult.error`. On a SQLAlchemy `DBAPIError` that string embeds the failed statement **and its bind-parameter tuple** — i.e. the entire document `content` + `metadata` — leaking it into logs and back to callers. A new bounded `_safe_exc_summary` prefers the underlying driver message (via `.orig` / `__cause__`), strips the `[SQL: ...]` / `[parameters: ...]` tail, truncates, and prefixes the exception class name, while keeping the `external_id` / `doc_id` context. Applied to the `submit_batch` create/update-record warnings and the pending/staged-document processing failure surfaces (process error log, status-update-failure warning, and `DocumentResult.error`). The persisted `error_message` column is intentionally unchanged (tracked separately).

## Test plan
- [x] Unit tests pass (`make test` unit leg: 6198 passed, coverage 78.55% ≥ 77% gate)
- [x] New unit coverage: `coerce_source_timestamp` (Z / offset / date-only / datetime passthrough / None / unparseable→None), `submit_batch` insert + re-queue string coercion incl. an asyncpg-`DataError`-shaped regression, `remember`/`remember_batch` coercion, and `_safe_exc_summary` payload-absence on both submit_batch and pending-processor failure paths
- [ ] Integration tests pass (run in CI — Docker not available in the dev sandbox): new `@pytest.mark.integration` Postgres tests assert a string `source_timestamp` round-trips to a real `timestamptz` column as a `datetime` on both the insert and re-queue/update paths
- [x] Lint/format clean (`make format`, `make lint`)


filippetrovic:
## Multi-perspective review — `source_timestamp` coercion + ingestion-log payload redaction

**Overall assessment: ✅ APPROVE / ready to merge.** Eight reviewer perspectives plus a dedicated security audit; CI is green (9/9 checks, incl. the real-Postgres integration job and the secret-typing gate). No CRITICAL or HIGH findings. The findings below are non-blocking — a couple of optional hardening/edge-case items and one design discussion point.

### Reviewer verdicts
| Perspective | Verdict |
|---|---|
| Security | APPROVE — change is security-positive (removes a payload leak); `coerce_source_timestamp` has a no-raise guarantee, no injection/DoS via `fromisoformat` |
| Performance | PASS — `coerce_source_timestamp` is O(1) per doc (datetime/None fast paths ~0.08µs; no regex); `_safe_exc_summary` runs only on cold failure paths; no double-parsing |
| Correctness | Covered transitively (coercion verified at every construction boundary; tz-aware in all branches; no placeholders) |
| Tests | APPROVE — coercion variants, both submit_batch paths, the asyncpg-`DataError`-shaped regression, and payload-absence on all three failure surfaces all asserted |
| Database | APPROVE — coerced datetime is timezone-aware in every branch (`Z`→`+00:00`, explicit offset preserved, date-only→midnight UTC); binds cleanly to `timestamptz`; no schema/migration needed |
| Observability | APPROVE — error class + `external_id`/`doc_id` context preserved (failures stay diagnosable in Logfire); no raw text reaches span attributes |
| Docs | PASS — CHANGELOG + docstrings accurate; no internal identifiers; no telemetry-contract drift |
| Devil's Advocate | Raised the discussion points below |

### MEDIUM (non-blocking — design discussion)
- **[Design] Coercion lives at the facade, not on the model.** `coerce_source_timestamp` is applied in `remember` / `remember_batch` / `submit_batch` and the ingest staging paths, not in `Document.__post_init__`. **Disposition: deferred by design.** All public ingest API paths are covered (the facade coerces *before* delegating to the engine, so engine-built Documents receive a coerced value). Moving coercion into the `core/models` `Document` would require importing the `pipelines/flows/ingest` parser into the model layer — a layering inversion / import-cycle risk. A direct `Document(...)` construction that bypasses the public API entirely is outside the documented ingest surface. Reasonable future hardening, but not a defect in this PR.

### LOW (optional)
- **[Behavior] `DocumentResult.error` is now sanitized** (`_safe_exc_summary` instead of `str(exc)`). This is the intended security fix, and it is already called out in the CHANGELOG. Callers that pattern-matched on raw exception substrings should rely on the exception class prefix instead.
- **[Scope] The persisted `error_message` column is intentionally left raw** (`doc.mark_failed(str(exc))`). This is explicitly disclosed in the CHANGELOG ("The persisted `error_message` column is unchanged"); it is a distinct surface (per-row DB column, not centralized logs) and a noted follow-up — not a regression introduced here.
- **[Tests] Add a tz-naive ISO unit case** (e.g. `"2026-01-15T10:30:00"` with no `Z`/offset). The implementation handles it via `fromisoformat`; explicit coverage would lock the behavior down.
- **[Tests] Integration tests use only `+00:00`.** Non-UTC offsets (`+05:30`, `-08:00`) are covered at the unit level; adding one to the integration round-trip would harden the real `timestamptz` bind assertion.

### Security audit
**Verdict: PASS.** All 881 changed lines scanned across the 16 categories — no injection/exec/eval, no SQL construction, no network/SSRF, no deserialization, no file/path handling, no regex/ReDoS, no weak crypto, no hardcoded secrets. The change *improves* the secret-handling posture by preventing document content (potential secrets from ingested sources) from reaching logs and caller-facing error strings. The `_SECRET` token in tests is an educational sentinel, not a credential. No findings to auto-fix.

---
*Disposition: merge. The MEDIUM is a deferred design preference with a clear layering rationale; the LOWs are optional follow-ups. The fix is correct, well-tested at unit + real-Postgres integration level, and CI is fully green.*

probablyrobot (approved):

PR #849, opened 2026-05-27
Honor Neo4j pool-sampler config on the shared-driver path

filippetrovic:
## Summary
- `Neo4jBackend.from_driver` now accepts `pool_sampler_enabled` / `pool_sampler_interval_ms` and seeds the instance fields from them, instead of hardcoding the high-frequency pool sampler off. `connect()` already starts the sampler on the shared-driver path, so enabling it via config now works end to end. The docstring no longer claims the sampler is unavailable on this path.
- `from_driver` seeds `_max_connection_pool_size` from the driver's real `pool.pool_config.max_connection_pool_size` (getattr-guarded, falling back to the driver default of 100), so the utilization **gauge** denominator matches the **sampler**'s. Previously the gauge fell back to 100 while the sampler read the real pool size, so the two utilization figures disagreed.
- `VectorCypherEngine.connect()` forwards the sampler config from `get_graph_config()` into `from_driver(...)`, the same way it already forwards write-concurrency and query-timeout settings.
- `VectorCypherEngine.disconnect()` now disconnects storage (which stops the backend's pool sampler task) **before** closing the shared driver, so in-flight sampler reads never run against a closed driver.
- No new background task and no new config field; SurrealDB / sqlite_lance paths are untouched.

## Test plan
- [x] Unit tests pass (`tests/unit/` green locally; added coverage for sampler-via-`from_driver`, `disconnect()` ordering, and the gauge denominator)
- [ ] Integration tests pass (CI — requires Neo4j + Postgres via Docker)
- [x] `ruff format` / `ruff check` / `ty check src/` clean

### DIFF
--- a/src/khora/engines/vectorcypher/engine.py
+++ b/src/khora/engines/vectorcypher/engine.py
@@ -507,6 +507,8 @@ class VectorCypherEngine:
                     entity_write_concurrency=getattr(neo4j_cfg, "entity_write_concurrency", 12),
                     relationship_write_concurrency=getattr(neo4j_cfg, "relationship_write_concurrency", 8),
                     query_timeout=neo4j_query_timeout,
+                    pool_sampler_enabled=getattr(neo4j_cfg, "pool_sampler_enabled", False),
+                    pool_sampler_interval_ms=getattr(neo4j_cfg, "pool_sampler_interval_ms", 500),
                 )
 
         await self._storage.connect()
@@ -683,17 +685,20 @@ class VectorCypherEngine:
 
         await shutdown_telemetry()
 
-        if self._neo4j_driver:
-            await self._neo4j_driver.close()
-            self._neo4j_driver = None
+        # Disconnect the storage coordinator (which owns the from_driver-wrapped
+        # Neo4jBackend) BEFORE closing the shared driver, so the backend's
+        # pool sampler task is stopped while the pool is still alive.
+        if self._storage:
+            await self._storage.disconnect()
+            self._storage = None
 
         if self._temporal_store:
             await self._temporal_store.disconnect()
             self._temporal_store = None
 
-        if self._storage:
-            await self._storage.disconnect()
-            self._storage = None
+        if self._neo4j_driver:
+            await self._neo4j_driver.close()
+            self._neo4j_driver = None
 
         self._embedder = None
         self._retriever = None
--- a/src/khora/storage/backends/neo4j.py
+++ b/src/khora/storage/backends/neo4j.py
@@ -641,20 +641,21 @@ class Neo4jBackend(GraphBackendBase):
         entity_write_concurrency: int = _DEFAULT_ENTITY_WRITE_CONCURRENCY,
         relationship_write_concurrency: int = _DEFAULT_RELATIONSHIP_WRITE_CONCURRENCY,
         query_timeout: float | None = 5.0,
+        pool_sampler_enabled: bool = False,
+        pool_sampler_interval_ms: int = 500,
     ) -> Neo4jBackend:
         """Create a Neo4jBackend from an existing AsyncDriver.
 
         The backend will NOT close the driver on disconnect, since
         it does not own it.
 
-        **Sampler note**: the high-frequency pool sampler is *not* started
-        by this constructor, even if the caller mutates
-        ``instance._pool_sampler_enabled`` afterwards. ``from_driver()`` is
-        typically used by shared-driver integrations (tests, fixtures,
-        connection-multiplexing setups) where the owner of the driver is
-        responsible for metric lifecycle. If you need the sampler on a
-        shared-driver backend, use the standard ``Neo4jBackend(...)``
-        constructor with a dedicated driver instead.
+        The high-frequency pool sampler is honored on this path: pass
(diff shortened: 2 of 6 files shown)


filippetrovic:

8 reviewers (security, performance, correctness, test, docs, database, observability, devil's advocate) + a dedicated security audit. Findings are grouped by severity with stable IDs.

**Overall assessment: ✅ APPROVE.** The change is surgical, correct, and well-tested; CI is green. No CRITICAL/HIGH issues remain (the one HIGH was fixed in this PR — see H1).

### CRITICAL
- None.

### HIGH
- **H1 — [Docs] `tests/unit/storage/backends/test_neo4j_coverage.py:470` — _RESOLVED_.** The comment `# from_driver hard-disables the sampler regardless of init args.` was made inaccurate by this change (the kwargs are now honored). Updated to `# from_driver leaves the sampler off by default (no pool_sampler_enabled arg here).`; the assertion (sampler off by default) still holds and the test passes.

### MEDIUM
- **M1 — [Test] `tests/unit/engines/vectorcypher/test_engine_coverage.py`** — the ordering test asserts call order via `mock_calls` index rather than `assert_awaited_once()`, so in isolation it would not catch a *called-but-not-awaited* regression. Non-blocking: the companion `test_both_storage_and_driver_are_torn_down` already asserts the awaits, so the behavior is covered in aggregate.
- **M2 — [Test] `tests/unit/test_neo4j_pool_metrics_correctness.py`** — the sampler emit threshold is `count >= 5` over ~20 ticks. Reviewer suggested tightening to `>= 15`. **Intentionally left at `>= 5`**: a tighter timing threshold raises CI flakiness under parallel-test CPU contention (observed during validation), and the loose bound still proves the sampler fires repeatedly.
- **M3 — [Security/Correctness] `src/khora/storage/backends/neo4j.py:682-684`** — reads Neo4j driver internals (`_pool.pool_config.max_connection_pool_size`), a private API → forward-compat risk. Accepted: triple `getattr` guard with a safe `or 100` fallback, read-only access, and a documented stability range (neo4j 5.x–6.2) consistent with the existing pool-metrics code.
- **M4 — [Devil's Advocate]** — gauge denominator (seeded once at construction) could theoretically diverge from the sampler (reads live each tick) if the pool were resized at runtime. Accepted: the Neo4j async driver does not support runtime pool resize, and pool reads are documented best-effort.

### LOW
- **L1 — [Performance] `neo4j.py` sampler** — holds `pool.lock` ~200–300µs per tick → ~0.04% hold time at the 500ms default; negligible, zero hot-path cost (background instrumentation only).
- **L2 — [Test]** — closure factory in the sampled-recorder wiring is the standard capture pattern; correct, just fragile to future refactors. No change needed.
- **L3 — [Test]** — sampler tests use `max_pool_size=10`; the default-100 / fallback paths are covered separately by `TestFromDriverMaxPoolSize`.
- **L4 — [Docs] `CLAUDE.md`** — the sampler note doesn't explicitly mention that `from_driver(...)` now honors the sampler kwargs (only the env-var/standard-constructor path). Optional clarity follow-up; not drift.
- **L5 — [Security]** — sampler config knobs (bool + bounded int) and the GC finalizer are safe-by-default instrumentation; informational, no action.

### Positive confirmations
- **Denominator agreement** verified across all branches (real ceiling / missing `pool_config` / zero ceiling): gauge `self._max_connection_pool_size or 100` and sampler `getattr(pool_config, "max_connection_pool_size", 0) or self._max_connection_pool_size` agree in every case.
- **Disconnect ordering** correct: `storage.disconnect()` (stops the sampler task) runs before the shared `driver.close()`; the `from_driver`-wrapped backend correctly does *not* close the shared driver (`_owns_driver=False`).
- **Scope is tight** — no new config field, no new background task; `docs/telemetry-contract.json` correctly untouched (no new metric); SurrealDB / sqlite_lance paths unaffected.
- **Backward compatible** — sampler is off by default; existing callers unaffected.

## Security Audit
Scanned the diff's added lines across all 16 categories (injection, authn/z, crypto/secrets, access control, SSRF, deserialization, file/path, open redirect, mass assignment, ReDoS, supply chain, boundary validation, …). **No findings.** The diff introduces no untrusted-input source, no dangerous sink, no secrets, no network/file/deserialization surface — it reads driver pool internals (read-only), reorders teardown, and adds two bounded config knobs. (Pattern-scan hits were false positives: case-insensitive `DES` matching "des" within the word "degrades" in comments.)

**Security verdict: PASS.**

---
**Recommendation: merge.** CI green; all reviewers approve; the single HIGH (stale comment) is fixed in this PR.

filippetrovic:
Addressed review findings **M1** and **M2** in `39a9b3c`:

- **M1** — the disconnect-ordering test now asserts both `storage.disconnect()` and `driver.close()` were *awaited* (via `assert_awaited_once()`), not just recorded in `mock_calls` — so a called-but-not-awaited regression is caught in isolation.
- **M2** — the sampler emit test widens its window to ~40 ticks and raises the per-histogram threshold from `>= 5` to `>= 15`, catching a sampler that stalls after a few ticks while keeping a ~2.7x margin against event-loop jitter under parallel CI load (verified stable across repeated runs).

msrdic (approved):

PR #854, opened 2026-05-27
Add opt-in Neo4j connection-pool keepalive

filippetrovic:
## Summary

Adds an opt-in, **default-off** background *keepalive* to the Neo4j backend that keeps idle pooled connections warm, so they are not silently idle-dropped by an intervening NAT / load balancer. An idle-dropped (half-open) socket otherwise triggers a slow, unbounded stale-connection teardown on the next reuse, which can stall connection acquisition well past the configured timeout. The keepalive removes that idle-drop trigger using only public driver APIs (it is a mitigation, not a hard guarantee against all dead-connection causes).

It mirrors the existing pool-sampler task lifecycle:

- **Config** — `Neo4jConfig.pool_keepalive_enabled` (default `False`) and `pool_keepalive_interval_ms` (default `15000`, validated `[50, 60000]`), each with single- and double-underscore env-var aliases. Default 15s sits below the typical 30s liveness window by design.
- **Backend** — every interval, read the *free (idle)* connection count from the driver pool under a brief lock, then fire `min(free, 16)` **detached, non-awaited** `RETURN 1` pings on a **raw driver session**. The cap bounds fan-out on a large idle pool. Pings deliberately bypass the instrumented session wrapper, so they never pollute the `acquire_duration` / `session.duration` / pool-timeout metrics, and are never wrapped in `asyncio.wait_for` — a ping that hits a dead connection runs its close to completion in the background (self-healing), rather than being cancelled mid-teardown and stranding a half-open socket.
- **Lifecycle** — honored on all three backend construction paths (`__init__`, `from_config`, `from_driver`) and forwarded from the vectorcypher engine; started on `connect()`, stopped on `disconnect()`, with a `weakref.finalize` GC-cancel guard. Idempotent start, clean cancel-and-await stop.
- **Telemetry** — dedicated counters `khora.neo4j.pool.keepalive.pings` / `khora.neo4j.pool.keepalive.failures`, registered in the telemetry contract.
- **Scope** — Neo4j-backend only; the other graph backends and the embedded SQLite/Lance path are unaffected. Closes a config-plumbing parity gap so the feature is reachable whether the backend is built from config or from a shared driver, and via both env-var alias forms.

## Test plan

- [ ] Unit tests pass (keepalive lifecycle, metric isolation via raw session, `min(free, 16)` cap, failure resilience, no `asyncio.wait_for`, idle-count read, config validation + env aliases, all three construction paths)
- [ ] Telemetry-contract drift gate passes (both new counters registered)
- [ ] Integration tests pass (real Neo4j: pings fire and idle stays warm; clean disconnect; sampler + keepalive coexist; non-Neo4j paths unaffected) — gated, runs in CI
- [ ] Manual verification: set `KHORA_STORAGE_GRAPH_POOL_KEEPALIVE_ENABLED=true`, leave the service idle past the network idle window, then issue a read — acquisition completes promptly and idle connection count stays warm across the gap

### DIFF
--- a/docs/telemetry-contract.json
+++ b/docs/telemetry-contract.json
@@ -925,6 +925,20 @@
       "unit": ""
     },
     {
+      "name": "khora.neo4j.pool.keepalive.pings",
+      "type": "counter",
+      "stability": "internal",
+      "unit": "1",
+      "description": "RETURN 1 keepalive pings attempted on idle pooled connections by the opt-in connection-pool keepalive (KHORA_STORAGE_GRAPH_POOL_KEEPALIVE_ENABLED). Counts every ping attempt, fired only when the keepalive loop observes idle connections. Zero when the keepalive is disabled."
+    },
+    {
+      "name": "khora.neo4j.pool.keepalive.failures",
+      "type": "counter",
+      "stability": "internal",
+      "unit": "1",
+      "description": "Keepalive pings that raised. A non-zero rate is the expected dead-connection self-heal canary (the failed close runs to completion in the background and removes the stale connection from the pool), not an error condition."
+    },
+    {
       "name": "khora.neo4j.relationship.source_id_truncated",
       "type": "counter",
       "stability": "public",
--- a/docs/telemetry-contract.json
+++ b/docs/telemetry-contract.json
@@ -929,14 +929,14 @@
       "type": "counter",
       "stability": "internal",
       "unit": "1",
-      "description": "RETURN 1 keepalive pings attempted on idle pooled connections by the opt-in connection-pool keepalive (KHORA_STORAGE_GRAPH_POOL_KEEPALIVE_ENABLED). Counts every ping attempt, fired only when the keepalive loop observes idle connections. Zero when the keepalive is disabled."
+      "description": "Keepalive RETURN 1 pings attempted on idle pooled connections. Incremented on every ping attempt by the opt-in keepalive loop (KHORA_STORAGE_GRAPH_POOL_KEEPALIVE_ENABLED); the per-cycle ping count tracks the observed idle-connection count, capped at 16. Zero when the keepalive is disabled."
     },
     {
       "name": "khora.neo4j.pool.keepalive.failures",
       "type": "counter",
       "stability": "internal",
       "unit": "1",
-      "description": "Keepalive pings that raised. A non-zero rate is the expected dead-connection self-heal canary (the failed close runs to completion in the background and removes the stale connection from the pool), not an error condition."
+      "description": "Keepalive pings that raised an exception. A non-zero rate is expected and indicates the self-heal canary working: the ping hit a dead connection, whose teardown then runs to completion in the background and removes the stale connection from the pool. Not an error condition."
     },
     {
       "name": "khora.neo4j.relationship.source_id_truncated",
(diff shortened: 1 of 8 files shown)


filippetrovic:

Eight independent reviewers (security, performance, correctness, test, database, observability, docs, devil's advocate) plus a dedicated security audit. **Overall assessment: APPROVE.** No CRITICAL or HIGH findings. All five MEDIUM findings and the actionable LOW have been addressed in follow-up commit `6875f6b`.

### CRITICAL
- None.

### HIGH
- None.

### MEDIUM (all addressed in `6875f6b`)
- **M1 — [Docs]** `docs/telemetry-contract.json` — the `keepalive.pings` description ("fired only when the keepalive loop observes idle connections") was misleading about when the counter increments. ✅ Reworded: increments on every ping attempt; per-cycle count tracks idle count capped at 16.
- **M2 — [Docs]** `docs/telemetry-contract.json` — the `keepalive.failures` description ("the failed close runs to completion") read like a close failure rather than a ping failure. ✅ Reworded to clarify it counts pings that raised, and that a non-zero rate is the expected self-heal canary.
- **M3/M4/M5 — [Observability]** `storage/backends/neo4j.py` — three keepalive log calls used f-string interpolation instead of loguru structured fields (the documented logging standard). ✅ Converted to structured `{error}` fields. (The pre-existing pool sampler uses the same f-string pattern in the equivalent spots; converting it is out of scope for this change and is a good repo-wide follow-up.)

### LOW / non-blocking notes (Devil's Advocate)
- **L1** — The `min(free, 16)` per-cycle fan-out cap lacked a rationale. ✅ Added an explanatory comment (warming any idle connection resets the network idle timer, so a small fixed cap suffices and bounds the concurrent-session burst).
- **Design rationale (no code change):**
  - *Simpler alternatives* — driver knobs (`max_connection_lifetime`, `liveness_check_timeout`, socket `keep_alive`) don't proactively warm idle connections (liveness only fires on acquire; socket keepalive interval is OS-controlled, often hours). A short application-level ping is the targeted mitigation; it's opt-in and default-off so the blast radius is limited.
  - *Sampler duplication* — the keepalive deliberately mirrors the existing pool-sampler task lifecycle. A shared `_PoolBackgroundTask` base is reasonable but premature at two instances; worth revisiting if a third pool task appears.
  - *Private driver internals* — reading `driver._pool` is getattr-guarded with warn-once degradation (returns 0 idle → fires nothing on a shape change; never crashes), matching the sampler's existing contract. A telemetry alert on a rising `keepalive`/sampler warn signal is a sensible future early-warning.
  - *Neo4j-only scope* — intentional; other backends carry no pool instrumentation. Not a gap for this change.

### Security Audit
**Verdict: PASS.** No findings across the 16 categories. The change adds no HTTP entry points, no user-controlled input, no new dependencies, no crypto, no deserialization, and no file handling. The keepalive query is a static `RETURN 1`; the new counters carry no `namespace_id` (cardinality-safe). One **INFO**: the gated integration test reads `KHORA_NEO4J_PASSWORD` with a local default of `"password"` — this is byte-identical to the existing Neo4j integration-test convention (`NEO4J_INTEGRATION_TEST`-gated), not a production secret.

### Highlights the reviewers confirmed
- **Metric isolation** is real and tested: pings run on a raw `driver.session(...)`, never the instrumented session, so `acquire_duration` / `session.duration` / the timeout counter stay clean (the unit test spies the hot-path instruments and asserts they're untouched).
- **No `asyncio.wait_for`** around the ping — a dead-connection teardown self-heals in the background instead of being stranded half-open.
- **Detached-task hygiene** — refs held in a set with a race-free done-callback discard; clean shutdown with in-flight pings; idempotent start; `weakref.finalize` GC guard.
- **Config parity** across all three construction paths plus both env-alias forms; opt-in default-off; interval validated/clamped `[50, 60000]`.
- **Test coverage** is thorough (41 unit incl. hygiene/shutdown-race, 5 env-alias, 4 gated integration) with meaningful (non-tautological) assertions.

probablyrobot (approved):

PR #969, opened 2026-06-01
Add case/separator-insensitive matching for source-timestamp metadata keys

filippetrovic:
## Summary

When a document is ingested without an explicit top-level `source_timestamp`, the event time is derived from `metadata` by matching a set of canonical timestamp fields. Previously this used **exact snake_case** key matching, so connectors emitting `occurredAt` (camelCase), `occurred-at` (kebab), `OCCURRED_AT` (SCREAMING_SNAKE), or `Occurred At` (Title Case) were silently not matched — every chunk fell back to ingest time, corrupting recency/decay scoring with no error. The advisory `validate_connector_metadata()` had the inverse blind spot (false "no timestamp" warnings on variant keys).

This change:

- **Tolerant key matching** — canonical timestamp fields are now matched case/separator-insensitively (a single `_normalize_ts_key` lowercases and strips `_ - ` and whitespace). An **exact** snake_case key still wins over a normalized variant when both are present; otherwise first-occurrence-in-metadata wins (deterministic).
- **`updated_at` now honored** by the extractor (appended at lowest priority for both event and non-event sources), bringing it in sync with the documented canonical field set.
- **New public `extract_source_timestamp(metadata)` helper**, re-exported from `khora.pipelines`, as the single shared implementation. The pipeline's internal extractor now delegates to it (no duplicated field knowledge for downstream consumers).
- **Validator mirrors the same normalization**, so its advisory no-timestamp warning no longer false-fires on variant keys, and it honors `source_type` supplied in the metadata mapping (still overridable by the explicit parameter). Empty-string / non-ISO / `source_system` / `tags` checks are unchanged.
- **No import cycle** — `coerce_source_timestamp` moves into the stdlib-only `connector_metadata` module (the dependency leaf); `ingest` imports and re-exports it, so existing `from ...flows.ingest import coerce_source_timestamp` / `_extract_source_timestamp` call sites keep working.
- **Tests** — parameterized matrices (every canonical field × every case/separator variant) added to the two existing test files; no parallel test files.

## Test plan

- [x] Unit tests pass (full `tests/unit` suite green when run sequentially; new parameterized matrices for the extractor and validator)
- [x] `ruff check`, `ruff format --check`, and type-check on changed files clean
- [ ] Integration tests + coverage gate (run in CI on the Docker-equipped runner)
- [x] Manual verification: `occurredAt` / `occurred-at` / `OCCURRED_AT` / `Occurred At` resolve to the canonical field; exact key wins over a variant; `updated_at` honored; validator no longer false-warns on variant-only metadata but still warns when genuinely absent

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,18 @@ All notable changes to Khora are documented here.
 
 Format: versions match git tags (`git tag vX.Y.Z`). Versions before 0.5.1 were internal (no git tags).
 
+## [0.18.2] - tolerant source-timestamp key matching
+
+Patch release. Fixes silent source-timestamp loss when connectors emit timestamp keys in non-snake_case styles, and adds a public extractor helper. No breaking changes.
+
+### Fixed
+
+- **Silent `source_timestamp` loss for non-snake_case connector keys**: the ingestion extractor only matched exact snake_case keys (`occurred_at`, `sent_at`, …), so connectors emitting camelCase / kebab-case / SCREAMING_CASE / TitleCase variants (`occurredAt`, `occurred-at`, `OCCURRED_AT`, `OccurredAt`) silently fell back to ingest time, breaking recency scoring. Key matching is now case/separator-insensitive; an exact key still wins over a normalized variant when both are present. `updated_at` is now honored by the extractor (appended last in priority for both event and non-event sources).
+
+### Added
+
+- **Public `khora.pipelines.extract_source_timestamp(metadata)` helper**: the timestamp-extraction logic shared by the ingestion pipeline is now an importable public function (`khora.pipelines.extract_source_timestamp`).
+
 ## [0.18.1] - silent-config and observability fixes
 
 Patch release. Fixes a batch of "accepted but silently ignored / silently dropped" config and observability gaps from Damir Krstanovic's reports. No breaking changes.
--- a/docs/extraction/ingestion-pipeline.md
+++ b/docs/extraction/ingestion-pipeline.md
@@ -567,10 +567,17 @@ This is a one-time migration for existing data.
 ## Canonical metadata fields per source
 
 Connector authors should populate `metadata.custom` with the canonical
-fields below. The ingestion pipeline reads them in
-`_extract_source_timestamp` to build `TemporalChunk.occurred_at`, which
-drives temporal recency scoring + per-source decay. The shape is exported
-as `khora.pipelines.ConnectorMetadata` (a `TypedDict`).
+fields below. The ingestion pipeline reads them via the public
+`khora.pipelines.extract_source_timestamp()` helper to build
+`TemporalChunk.occurred_at`, which drives temporal recency scoring +
+per-source decay. The shape is exported as
+`khora.pipelines.ConnectorMetadata` (a `TypedDict`).
+
+Key matching is **case/separator-insensitive** (since 0.18.2): a connector
+emitting `occurredAt`, `occurred-at`, `OCCURRED_AT`, or `Occurred At` all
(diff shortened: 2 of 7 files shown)


filippetrovic:

Reviewed by an agent team across security, performance, correctness, tests, docs, and a devil's-advocate pass. **2 of 6 reviewers (performance, docs) returned written findings; the remaining four did not respond in time, so correctness / tests / security / devil's-advocate below are the lead's direct verification + an inline security audit.** CI is green on all 9 required checks at time of writing.

**Overall assessment: APPROVE** — behavior is correct and fully covered by tests; the findings below are non-blocking polish (a perf micro-optimization and doc sync). They can land as a follow-up or a quick amend.

### CRITICAL
- None.

### HIGH
- None. (The performance reviewer labeled the regex-compilation item CRITICAL, but Python's `re` module internally caches compiled patterns, so the real-world impact is microseconds on an I/O+LLM-bound ingest path — reclassified LOW below.)

### MEDIUM
- **[Docs]** `docs/extraction/ingestion-pipeline.md` and `src/khora/pipelines/connector_metadata.py` comments still reference the internal `_extract_source_timestamp`; they should point at the new public `khora.pipelines.extract_source_timestamp`, and the doc should mention the new case/separator-insensitive matching (0.18.2). Doc sync only — no behavior impact.

### LOW
- **[Performance]** `connector_metadata.py` — pre-compile the normalizer regex at module level (`_NORMALIZE_RE = re.compile(r\"[\s_\-]+\")`) and normalize each field name once per loop iteration. Best-practice micro-opt (~1.4x on the normalizer itself; negligible absolute cost since `re` already caches).
- **[Docs]** Module docstring references a `§canonical-fields` anchor that doesn't exist in the target doc; add the anchor or drop the reference.

### Verified clean (no findings)
- **[Security / inline audit]** Normalizer regex `[\s_\-]+` is a single character class with a `+` quantifier — no nesting/alternation/backreference, so linear-time and **ReDoS-safe**. No `eval`/`exec`/`subprocess`/`pickle`/SQL-string-format; no secrets introduced; pure stdlib string + ISO-datetime parsing. ADR-084 N/A (no Pydantic/secret-named fields).
- **[Correctness]** Exact snake_case key wins over a normalized variant in both metadata orderings; first-occurrence-wins is deterministic; `updated_at` appended last in both priority lists (no change to existing outcomes); validator's `_EXTRACTOR_TIMESTAMP_FIELDS` + `source_type`-from-metadata fallback fire the no-timestamp warning iff nothing resolves; no import cycle (`connector_metadata` stays stdlib-only, `ingest → connector_metadata` one-way); back-compat re-exports keep `coerce_source_timestamp` / `_extract_source_timestamp` importable from `flows.ingest` (khora.py lazy imports unaffected). No placeholders.
- **[Tests]** Parameterized field×variant matrices cover every canonical field × every spelling variant in both existing test files (no parallel files); exact-datetime assertions; exact-wins / priority (event + non-event) / unknown-key / empty / unparseable / validator-absent / public-helper-import-smoke all covered. Full unit suite passes; CI `test` job (full suite + 77% coverage gate) green.
- **[Devil's advocate]** Approach is justified: silent `source_timestamp` loss on non-snake_case keys is a real, errorless data-quality bug, so tolerating variants in khora (rather than only fixing each connector) prevents recurring drift. Moving `coerce_source_timestamp` into the stdlib leaf is the minimal way to share one implementation without an import cycle. The per-call normalized-map build is negligible for realistic metadata sizes. `source_type`-from-metadata and the public `extract_source_timestamp` export are explicit ticket deliverables, not scope creep.

probablyrobot (approved):

PR #977, opened 2026-06-02
fix(integrations): cancel in-flight tasks on sync bridge teardown

filippetrovic:
## Problem

`khora.integrations._sync._shutdown_for_tests()` (a test-only helper) stopped the
process-wide bridge event loop **without cancelling the coroutines still running on it**.
If another test had a coroutine in flight on the bridge (submitted via
`asyncio.run_coroutine_threadsafe` inside `run_sync`), its future never resolved.
`run_sync`'s `future.result()` has no timeout and runs on a non-daemon
`ThreadPoolExecutor` thread, so the orphaned wait kept the pytest-xdist worker process
from ever exiting — the controller then waited forever and CI hung until GitHub's
6-hour cap. This showed up as an intermittent, multi-hour hang on the parallel
unit-test step.

## Fix

`_shutdown_for_tests()` now cancels every in-flight task on the loop and waits
(bounded, 2s) for the cancellation to settle **before** stopping it, so orphaned
`run_coroutine_threadsafe` futures resolve as `CancelledError` instead of wedging.
The wait is bounded and the loop is stopped unconditionally afterward, so a coroutine
that swallows `CancelledError` can't leak a zombie daemon thread. Production
`run_sync` / `_ensure_loop` paths are untouched — `_shutdown_for_tests` is test-only.

A new regression test (`test_shutdown_for_tests_cancels_in_flight_coroutine`) reproduces
the wedge: it submits a 30s coroutine to the bridge, calls `_shutdown_for_tests()`, and
asserts the future resolves as cancelled within 2s (a hang fails it).

## CI backstop

The suite had no per-test or per-job timeout, so any future wedge would again burn the
full 6 hours. This adds:
- `pytest-timeout` with `--timeout=120 --timeout-method=thread` on the unit and
  integration test steps — a hung test now fails fast and pytest-xdist names it
  (`crashed while running <nodeid>`).
- `timeout-minutes` on every CI job (test: 25, others: 15).

## Testing
- `tests/unit/integrations/test_sync_bridge.py` → 11 passed (incl. new regression test; fail-before / pass-after verified).
- `tests/unit/integrations/` → 402 passed, 31 skipped.
- `make format` clean; `ty` clean.


probablyrobot (approved):

filippetrovic:
## 🔍 Multi-perspective review — PR #977

**Overall assessment: ✅ APPROVE.** No CRITICAL or HIGH findings. Correctness, security, performance, and docs are clean; the regression test genuinely guards the fix; CI is green. The MEDIUM/LOW items below are deferred-scope or design-judgment calls, not blockers.

Reviewers: correctness, security, performance, test, docs, devil's advocate + a dedicated security audit.

### CRITICAL
- None.

### HIGH
- None.

### MEDIUM
- **M1 — [Devil's Advocate] Production `run_sync().result()` is still unbounded.** This PR fixes the *test-teardown* path (the actual cause of the intermittent multi-hour CI hangs), but a production coroutine that never resolves would still block a sync caller. **This is intentionally out of scope** — production hardening (bounding `run_sync`'s wait; joining the hermes executor with a timeout) is tracked as a separate follow-up. The teardown fix is the correct, zero-production-impact resolution for the CI hang.
- **M2 — [Test] The new regression test doesn't use the `shutdown_sync_bridge` fixture.** It calls `_shutdown_for_tests()` directly, leaving the bridge in a shutdown state. This is **consistent with the sibling `test_shutdown_for_tests_releases_loop`** (also no fixture; the next test rebuilds the bridge via `_ensure_loop()`), so it's a non-defect. Optional hardening for a follow-up: cover a task that swallows `CancelledError` and a multi-task case.

### LOW
- **L1 — [Devil's Advocate] `--timeout-method=thread` discards coverage for a timed-out worker.** True, but only when a test *actually hangs* — i.e. the failure case we want surfaced. `thread` was chosen deliberately: under pytest-xdist it reliably kills the wedged worker and xdist then **names the culprit** (`crashed while running <nodeid>`), which `signal` (SIGALRM, main-thread-only) does not do reliably for a thread blocked in a C-level `future.result()`. Verified empirically while reproducing the bug. Normal (non-hanging) runs are unaffected — CI just ran green.
- **L2 — [Devil's Advocate] Consider test-job `timeout-minutes` 25 → 30** for parity with the release workflow. The test job (unit + integration) is already at **25** (not 15 — the 15-min jobs are the fast ones, observed at 6s–3m28s this run), which is ~3× the normal ~8-min runtime. 25 is ample; the green run confirms it. Cosmetic.

### Security Audit
**PASS.** The diff has no security surface — no injection/exec/shell, no secrets/crypto/weak-random, no network/SSRF/file/deserialize/redirect, no auth or access-control paths. Supply chain: the only new dependency is `pytest-timeout` 2.4.0 (dev-only, MIT, maintained by pytest-dev), pinned in `uv.lock` with sha256 hashes, no install scripts, no known CVEs. CI workflow edits add only `timeout-minutes` and hard-coded pytest flags — no new secrets, permissions, or injection surface.

### What's strong here
- Correctness verified: loop ref snapshotted under `_loop_lock` before nulling (no race), `create_task` scheduled while the loop is still live, `if others:` guards `asyncio.wait([])`, bounded 2s wait + unconditional `loop.stop()` means a `CancelledError`-swallowing task can't leak a zombie thread.
- Production paths (`run_sync`, `_ensure_loop`) untouched — the fix is strictly test-teardown.
- Regression test is a real fail-before / pass-after guard.

Issue #985, opened 2026-06-04
VectorCypher: cross-encoder reranker lazy init has a check-then-set race that segfaults under concurrent queries on macOS arm64


sale87:
## Summary

`VectorCypherRetriever._apply_reranking` lazily constructs `self._reranker` with an unguarded check-then-set pattern. Under concurrent recall (the same `Khora` instance serving multiple async tasks), several tasks can simultaneously observe `self._reranker is None` and each race to construct a `CrossEncoderReranker`, which loads a `sentence_transformers.CrossEncoder` model. On macOS arm64 (Python 3.13, `torch==2.12.0`, `sentence-transformers==5.5.1`), this consistently segfaults the process the first time a recall path hits the reranker concurrently from cold.

The pattern in `khora/engines/vectorcypher/retriever.py:1966–1968`:

```python
if self._reranker is None:
    self._reranker = CrossEncoderReranker(model_name=self._config.reranking_model)
results = await self._reranker.rerank(...)
```

is not safe across `asyncio` tasks. The `CrossEncoder` constructed at `khora/query/reranking.py:200` is then driven via `asyncio.to_thread(model.predict, ...)` (`khora/query/reranking.py:242`); two concurrent loads + thread-offloaded `model.predict` on macOS arm64 manifest as `Segmentation fault: 11` accompanied by a leaked `loky` semaphore at shutdown (e.g. `/loky-21473-49yd3pye`).

## Environment

- **Khora**: `0.18.3` (built from `main` @ `4bceb6e71d5c906204b825cf22165b273180e9e3`)
- **Engine**: `vectorcypher`
- **Python**: 3.13.13 (cpython)
- **OS**: macOS arm64 (Apple Silicon)
- **Relevant deps**: `torch==2.12.0`, `sentence-transformers==5.5.1`, `transformers==5.9.0`, `tokenizers==0.22.2`, `scikit-learn==1.9.0` (brings `joblib`/`loky`)
- **Reranker model**: `cross-encoder/ms-marco-MiniLM-L-12-v2` (201 tensors, loaded lazily on first reranking call)

## Symptoms observed

Caught during a `--sample full` (2010 questions) benchmark run on top of khora `main`. Same code, same khora install, same Python: a prior `--sample medium` (559 questions) run on the same day completed cleanly. The full run crashed 75 min in, immediately as the retrieval phase began.

Tail of the crash log (timestamps in CEST), all from the same process:

```
[...]

PR #988, opened 2026-06-04
fix(vectorcypher): guard reranker lazy init with asyncio.Lock (#985)

sale87:
## Summary
Add an ``asyncio.Lock`` around the lazy init of both rerankers in ``VectorCypherRetriever``:

- ``khora/engines/vectorcypher/retriever.py:1966-1968`` — ``self._reranker`` (cross-encoder)
- ``khora/engines/vectorcypher/retriever.py:2162-2164`` — ``self._llm_reranker`` (temporal LLM reranker)

## Why
Both methods used an unguarded ``if self._reranker is None: self._reranker = …`` pattern. Today the constructors are synchronous and the event loop alone keeps the check-then-set safe, but only incidentally — any future change that adds an ``await`` between the ``is None`` check and the assignment would re-open a race where concurrent recall tasks could each kick off a parallel ``CrossEncoderReranker`` (or ``LLMReranker``) construction.

This was first surfaced by #985, where the downstream symptom (concurrent ``CrossEncoder`` loads driving 5 worker threads into torch-MPS from cold) segfaulted on macOS arm64. The actual segfault fix is to pin the cross-encoder device to CPU on darwin+arm64; that lands separately. This change makes the lazy-init invariant explicit and refactor-stable on every platform.

## Latency / metric impact
- **Linux + Intel Mac**: zero. The lock is uncontended after first call (a few µs); construction is synchronous so the event loop already serializes it.
- **macOS arm64**: zero (the constructor itself isn't the segfault — see the device-pin PR for that).
- **Quality metrics**: bit-identical. No change to scoring logic.

## Test plan
- [x] New ``TestRerankerLazyInitRace`` class in ``tests/unit/engines/test_vectorcypher_retriever.py`` covers:
  - retriever exposes ``asyncio.Lock`` on both rerankers after init.
  - 5 concurrent ``_apply_reranking`` calls construct exactly one ``CrossEncoderReranker``.
  - 5 concurrent ``_apply_llm_reranking`` calls construct exactly one ``LLMReranker``.
- [x] Full ``test_vectorcypher_retriever.py`` suite still green (47/47).

## Related
- Issue #985
- Companion PR: pin ``device='cpu'`` on darwin+arm64 (separate, fixes the actual segfault).

### DIFF
--- a/src/khora/engines/vectorcypher/retriever.py
+++ b/src/khora/engines/vectorcypher/retriever.py
@@ -521,11 +521,18 @@ class VectorCypherRetriever:
         # Lazy entity expansion cache: chunk_id -> expansion_score (0 = no match)
         self._expansion_cache: dict[UUID, float] = {}
 
-        # Cached cross-encoder reranker (lazy-init on first use, reused across queries)
+        # Cached cross-encoder reranker (lazy-init on first use, reused across queries).
+        # An ``asyncio.Lock`` makes the check-then-set explicit even though today's
+        # ``CrossEncoderReranker()`` constructor is synchronous: it survives a future
+        # refactor that adds an ``await`` between the check and the assignment, and
+        # documents at the call site that this state is shared across concurrent
+        # recall tasks (see issue #985).
         self._reranker: CrossEncoderReranker | None = None
+        self._reranker_lock = asyncio.Lock()
 
-        # Cached LLM reranker for temporal queries (lazy-init on first use)
+        # Cached LLM reranker for temporal queries (lazy-init on first use, same guard).
         self._llm_reranker: LLMReranker | None = None
+        self._llm_reranker_lock = asyncio.Lock()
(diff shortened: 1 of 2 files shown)


sale87:
Pushed a ``ruff format`` fix for the lint failure (the new test block had multi-line list comprehensions that ruff prefers to collapse).

The ``security`` job failure isn't from this PR — pip-audit is flagging two new ``aiohttp 3.13.5`` CVEs (CVE-2026-34993, CVE-2026-47265, fix in 3.14.0). Same failure shows on the companion PR #989 and presumably on ``main`` since aiohttp is a transitive dep. Happy to bump the pin or add to the ``--ignore-vuln`` allowlist in ``.github/workflows/`` if you want, but I'd rather leave that policy call to you.

probablyrobot (approved):

PR #989, opened 2026-06-04
fix(reranking): pin CrossEncoderReranker to CPU on macOS arm64 (#985)

sale87:
## Summary
On Apple Silicon, default ``CrossEncoderReranker(device=None)`` to ``device='cpu'`` instead of letting sentence-transformers auto-select MPS. On every other platform — Linux, Intel Mac, Windows — the path is dead and behavior is unchanged.

## Why
Sentence-transformers' ``CrossEncoder`` picks the MPS backend on macOS arm64 when ``device`` is not set. The khora-graphrag-benchmark harness drives 5 concurrent ``asyncio.to_thread(model.predict, …)`` calls from a shared retriever instance, and torch-MPS is not thread-safe under that pattern — concurrent ``.predict()`` on a shared model segfaulted the process on Apple Silicon. Full details + crash log in #985.

The companion PR #988 adds an ``asyncio.Lock`` around the lazy init of ``self._reranker`` / ``self._llm_reranker``. That makes the lazy-init invariant explicit, but the lock alone doesn't fix the segfault — the real hazard is concurrent ``.predict()`` on a shared MPS-backed model, not the construction. Pinning to CPU on Apple Silicon is the actual fix.

Callers who know their access pattern is serialized can opt back in by passing ``device='mps'`` explicitly.

## Latency / metric impact
- **macOS arm64**: cross-encoder rerank goes from MPS (~50 ms per batch) to CPU (~120–150 ms). On a full GraphRAG-Bench run that's ~+2.5 min wall-clock (~+0.8% of total). Quality metrics (accuracy, r_score, ar_metric, faithfulness, …) are unchanged within fp32-level numerical noise — ordering of the top reranked candidates is bit-identical in practice.
- **Linux + Intel Mac + Windows**: zero change. The new code path doesn't fire (``_is_darwin_arm64()`` returns False).
- **Memory / cost**: unchanged.

## Test plan
- [x] New ``TestCrossEncoderDeviceSelection`` class in ``tests/unit/test_query_reranking.py``, 5 cases:
  - Darwin + arm64 + ``device=None`` → ``device='cpu'``.
  - Linux x86_64 → unchanged (``None``).
  - Linux aarch64 (Graviton) → unchanged.
  - Darwin x86_64 (Intel Mac) → unchanged.
  - Darwin arm64 + explicit ``device='mps'`` → respected.
- [x] Full ``test_query_reranking.py`` suite still green (32/32).

## Related
- Issue #985
- Companion: PR #988 (asyncio.Lock around reranker lazy init).

### DIFF
--- a/src/khora/query/reranking.py
+++ b/src/khora/query/reranking.py
@@ -8,6 +8,7 @@ Provides neural re-ranking of search results using:
 from __future__ import annotations
 
 import asyncio
+import platform
 from abc import ABC, abstractmethod
 from collections.abc import Callable
 from dataclasses import dataclass, field
@@ -22,6 +23,16 @@ if TYPE_CHECKING:
     from khora.storage.coordinator import StorageCoordinator
 
 T = TypeVar("T")
+
+
+def _is_darwin_arm64() -> bool:
+    """True on Apple Silicon Macs.
+
+    Used to pin the cross-encoder reranker to CPU because torch-MPS is
+    not thread-safe under concurrent ``.predict()`` calls dispatched
+    through ``asyncio.to_thread`` (see issue #985).
+    """
+    return platform.system() == "Darwin" and platform.machine() == "arm64"
 
 
 @dataclass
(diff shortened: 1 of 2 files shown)


sale87:
The ``security`` job failure isn't from this PR — pip-audit is flagging two new ``aiohttp 3.13.5`` CVEs (CVE-2026-34993, CVE-2026-47265, fix in 3.14.0). Same failure shows on the companion PR #988 and presumably on ``main`` since aiohttp is a transitive dep. Happy to bump the pin or add to the ``--ignore-vuln`` allowlist in ``.github/workflows/`` if you want, but I'd rather leave that policy call to you.

probablyrobot (approved):

PR #1000, opened 2026-06-05
Add denormalized document-grained columns to khora_chunks

filippetrovic:
## Summary

Adds eight nullable, document-grained columns to the `khora_chunks` temporal-store table so recall filters can push down on a single chunk row without joining back to `documents`. This is a schema/catalog-only change — the write-path population and the post-backfill production indexes land in follow-up changes.

- **New catalog-only Alembic migration.** Adds `source_type`, `source_name`, `source_url`, `source_timestamp`, `external_id`, `content_type`, `source`, `title` — all **nullable with no `server_default`**, so it is an instant catalog update with no table rewrite on PostgreSQL 11+. The migration follows the established catalog-only precedent: Postgres dialect gate, `has_table` guard (the table is runtime-created, so fresh deploys/SQLite fixtures get the columns from the table definition instead), a bounded `lock_timeout`, and a structured applied/lock-timeout log. It creates **no indexes**, and skips cleanly on non-Postgres dialects. `downgrade()` symmetrically drops the eight columns.
- **Runtime table definition + indexes.** The runtime `khora_chunks` table definition gains the eight columns, and `connect()` creates `CREATE INDEX IF NOT EXISTS` for the filterable subset — `(namespace_id, source_type)`, `(namespace_id, source_name)`, a partial `(namespace_id, source_timestamp) WHERE source_timestamp IS NOT NULL`, `(namespace_id, external_id)`, `(namespace_id, content_type)`. `source_url`, `source`, and `title` are intentionally left unindexed.
- **Carrier model.** `TemporalChunk` carries the eight fields. `source_timestamp` (the producer's verbatim event-time) is now **distinct** from `occurred_at` (the chunk event-time) and is no longer aliased to it; it falls back to `occurred_at` only when unset, preserving today's temporal-decay behavior until the write-path populates the distinct value.

## Test plan
- [x] Unit tests for the `TemporalChunk` → `Chunk` carrier: distinct `source_timestamp` vs `occurred_at` survives the round-trip (date-collapse tripwire), fallback to `occurred_at` when unset, and both-unset → `None`.
- [x] Migration tests: upgrade adds the eight columns (nullable, no default), downgrade drops them, no-op when the table is absent, and the chain stays green on SQLite.
- [x] `ruff format` / `ruff check` clean; targeted suite green locally (Postgres-marked migration tests run in CI's integration lane).
- [ ] Full CI suite (unit + integration + e2e) green.

### DIFF
--- a/src/khora/db/migrations/versions/041_khora_chunks_denormalized_columns.py
+++ b/src/khora/db/migrations/versions/041_khora_chunks_denormalized_columns.py
@@ -1,0 +1,182 @@
+"""Add denormalized document-grained columns to khora_chunks.
+
+Revision ID: 041_khora_chunks_denormalized_columns
+Revises: 040_chunks_last_accessed_at
+Create Date: 2026-06-05
+
+Schema changes:
+
+* Eight nullable columns on the ``khora_chunks`` temporal-store table,
+  copied down from the parent document so recall filters can be applied
+  directly on the chunk row without a join:
+
+  - ``source_type`` ``VARCHAR`` — document source category (nullable; no
+    default — chunks predating the backfill stay NULL).
+  - ``source_name`` ``VARCHAR`` — human-readable source label.
+  - ``source_url`` ``TEXT`` — origin URL (may be long).
+  - ``source_timestamp`` ``TIMESTAMPTZ`` — the producer's verbatim event
+    time, distinct from ``occurred_at`` (the chunk event-time).
+  - ``external_id`` ``VARCHAR`` — upstream identifier.
+  - ``content_type`` ``VARCHAR`` — MIME / content category.
+  - ``source`` ``VARCHAR`` — raw source identifier.
+  - ``title`` ``TEXT`` — document title (may be long).
+
+  All eight are nullable with NO server_default. This is a pure catalog
+  update: a nullable ``ADD COLUMN`` with no default is an instant catalog
+  change on PG 11+ — no table rewrite — so we deliberately avoid the
+  NOT-NULL-DEFAULT fast-path (and its ``server_version_num`` assert) used
+  by migration 038.
+
+Index note: NO indexes are created here. The indexed subset
+(``source_type``, ``source_name``, ``source_timestamp`` (partial),
+``external_id``, ``content_type``) is created by the runtime
+``PgVectorTemporalStore.connect()`` via ``CREATE INDEX IF NOT EXISTS`` and
+will be added against the populated columns in a later post-backfill step.
+This migration only widens the catalog.
+
+Lock-timeout safety: a Postgres-only ``SET lock_timeout = '5s'`` is issued
+**before any DDL** so the ``ADD COLUMN`` AccessExclusiveLock acquisition is
+bounded. The setting is scoped to the migration's enclosing transaction
+(env.py wraps the whole upgrade in a single transaction; the SET expires at
+COMMIT). On lock-timeout the upgrade logs ``khora.migration.applied`` at
+ERROR with ``lock_timeout_tripped=True`` (detected by the Postgres SQLSTATE
+``55P03`` / ``lock_not_available`` on ``OperationalError.orig.pgcode`` — any
+other ``OperationalError`` class logs ``lock_timeout_tripped=False`` so
+dashboards don't conflate deadlocks / connection drops with the lock-timeout
+signal). Either path re-raises so Alembic rolls back the transaction.
+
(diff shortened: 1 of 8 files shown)


probablyrobot (approved):

filippetrovic:
## 🔍 Multi-Perspective Review — PR #1000

Reviewed by an 8-role review team (security, performance, correctness, tests, database, observability, docs, devil's advocate) plus a dedicated security audit. Scope: a **catalog-only** schema change adding eight nullable denormalized columns to `khora_chunks`, the runtime table/index definitions, and the `TemporalChunk` carrier — write-path population and the production (post-backfill) indexes are intentionally deferred to follow-up changes.

No CRITICAL or HIGH defects. Migration is correct, catalog-only (instant ADD COLUMN, no rewrite), well-tested, and faithful to the stated scope. The items below are advisory.

---

### CRITICAL
_None._

### HIGH
_None._ (The devil's advocate raised strong design challenges — denormalization-vs-join, the deferred write-path, and 5/8 index coverage. These are **settled design decisions / tracked follow-up work**, not defects in this change — see *Design discussion* below.)

### MEDIUM
- **M1 · [Tests]** `tests/integration/db/test_migration_041_chunks_denormalized_columns.py` — coverage gap: no test exercises the *fresh-deploy* path where the runtime table definition (`PgVectorTemporalStore.connect()`) creates the eight columns **without** running this migration. The upgrade/downgrade/no-op/SQLite paths are covered. Optional, non-blocking.
- **M2 · [Docs]** No `CHANGELOG.md` entry. This is consistent with the precedent for prior plain schema migrations (they roll into the next release PR rather than carrying a dedicated entry), so likely intentional — confirm against the release convention.

### LOW
- **L1 · [Tests]** `test_migration_041_…py:204-217` — column-metadata is read by positional tuple index (`cols[name][1]` / `[2]`). It is **correct** (`SELECT column_name, data_type, is_nullable, column_default`), but brittle to query/tuple reordering. Consider named extraction for clarity. (Note: this Postgres assertion runs in CI's integration lane, not locally without Docker.)
- **L2 · [Performance]** The five runtime `CREATE INDEX IF NOT EXISTS` in `connect()` are plain (not `CONCURRENTLY`). Safe today — the migration creates **no** indexes and the runtime path targets fresh/empty tables — but it would briefly block writers if ever run against a large pre-existing table. Documented; production index creation is the deferred post-backfill step.
- **L3 · [Docs]** A one-line comment in `pgvector.py` explaining *why* `source_url` / `source` / `title` are denormalized but left unindexed (TEXT/large or not equality-filtered) would preempt the "inconsistent coverage" question.

### INFO
- **[Security audit — PASS]** No injection (all SQL is hardcoded literals or SQLAlchemy `Column` objects; no interpolation of untrusted input), no secrets (the test fixture reads `KHORA_DATABASE_URL` from env with a localhost default), ADR-084-compliant (no secret-named model fields), no dependency/manifest changes.
- **[Security — pre-existing, out of scope]** The HNSW index DDL in `pgvector.py` (unchanged by this PR) builds its `WITH (m=…, ef_construction=…)` clause via f-string interpolation of config integers. Not introduced here and not user-controlled; worth a separate hardening follow-up.
- **[Observability — PASS]** The migration's `khora.migration.applied` structured log (INFO/ERROR + `lock_timeout_tripped`) follows the failure-observability convention and reuses an existing log key, so no telemetry-contract update is required (drift gate won't trip).
- **[Database / Correctness / Performance — PASS]** Nullable + no-default = instant catalog change; `has_table` guard + Postgres dialect gate; bounded `lock_timeout`; symmetric downgrade; single linear migration head (`040 → 041`); partial index on `source_timestamp` correct. The `source_timestamp`→`occurred_at` fallback is verified non-regressing (the column is unset on read until the write-path lands, so behavior is unchanged) and guarded by a distinct-timestamp tripwire test.

---

### Design discussion (non-blocking)
The devil's advocate fairly asked whether shipping nullable columns before the write-path/backfill creates a "half-built feature." In this change that's **deliberate sequencing**: the columns are inert (NULL) and no filter feature is exposed against them yet; the write-path population and the backfill-plus-indexes land as separate, already-scoped follow-up changes, and the distinct-`source_timestamp` carrier degrades safely to `occurred_at` in the interim. The denormalization-vs-join tradeoff and the indexed-subset choice are intentional design decisions carried by that broader plan. Recommend (optional) linking the follow-up work for reviewer visibility and documenting the chunk-staleness-on-document-update policy when the write-path lands.

PR #1001, opened 2026-06-05
Add RecallFilter model and two-regime validator

filippetrovic:
## Summary
- Add a public `RecallFilter` pydantic model — a Mongo-style, deterministic recall-filter contract intended for a future `recall(filter=…)` keyword. This change is the validation layer only: it validates and normalizes a filter document; it does not compile or execute queries.
- Typed system-key fields (`occurred_at`, `created_at`, `source_timestamp`, `source_type`, `source_name`, `source_url`, `external_id`, `content_type`, `source`, `title`) plus `StringOps`/`DateOps` operator submodels enforcing per-key operator subsets (`$exists` excluded on date keys; range operators excluded on string keys).
- Free-form `metadata` predicates via Mongo-style `metadata.<path>` dot-paths: a `model_validator(mode="before")` folds flat dot-path keys onto a recursion-safe per-instance carrier, and a `model_serializer` re-emits them flat so filters round-trip byte-stable (including under nested `$and`/`$or`). A bare `{"metadata": {…}}` is whole-blob equality.
- Two validation regimes, one error type: pydantic-structural (closed top-level key set via `extra="forbid"`, typed submodels, unset-vs-explicit-`null` via `model_fields_set`) and a recursive metadata-grammar walk (operator whitelist, `$in`/`$nin` must be lists, strict-bool `$exists`, operand opacity, and a loud operator-position-closure rule). Both surface as `RecallFilterValidationError` carrying structured `errors`.
- Bare-value sugar (scalar/list/dict ⇒ `$eq`; a bare list is exact-array equality, not membership), a `$date` typed literal for metadata timestamps, UTC normalization of datetimes, and logical `$and`/`$or`/`$nor`/`$not` (document and field position) with empty-array rejection.
- Export `RecallFilter`, `StringOps`, `DateOps`, `RecallFilterValidationError`, `RecallFilterUnsupportedError`, `Op`, `SYSTEM_KEYS` from both `khora` and `khora.filter`.

## Test plan
- [x] Validator unit tests (`tests/recall/test_filter_validator.py`, 112 tests): whitelist, per-key operator subsets, dot-path fold + byte-stable round-trip, bare-value sugar, operand opacity, `$date`, null semantics, logical recursion
- [x] Full unit suite green (`tests/unit` + `tests/recall`)
- [ ] Integration/e2e via CI

### DIFF
--- a/src/khora/__init__.py
+++ b/src/khora/__init__.py
@@ -51,6 +51,15 @@ from .dream import DreamConfig, DreamMode, DreamResult, DreamRunInfo, DreamScope
 from .engines import create_engine, list_engines, register_engine
 from .exceptions import EngineCapabilityError, KhoraError
 from .extraction.skills import EntityTypeConfig, ExpertiseConfig, RelationshipTypeConfig
+from .filter import (
+    SYSTEM_KEYS,
+    DateOps,
+    Op,
+    RecallFilter,
+    RecallFilterUnsupportedError,
+    RecallFilterValidationError,
+    StringOps,
+)
 from .hooks import SemanticFilter
 from .khora import (
     BatchHandle,
@@ -109,4 +118,12 @@ __all__ = [
     "DreamResult",
     "DreamRunInfo",
     "OpKind",
+    # Deterministic recall filter
+    "RecallFilter",
+    "StringOps",
+    "DateOps",
+    "RecallFilterValidationError",
+    "RecallFilterUnsupportedError",
+    "Op",
+    "SYSTEM_KEYS",
 ]
--- a/src/khora/filter/__init__.py
+++ b/src/khora/filter/__init__.py
@@ -1,0 +1,34 @@
+"""Public deterministic recall-filter surface.
+
+Exports the typed filter document (:class:`RecallFilter`), its operator
+submodels (:class:`StringOps`, :class:`DateOps`), the structured error types
+(:class:`RecallFilterValidationError`, :class:`RecallFilterUnsupportedError`),
+the operator vocabulary (:class:`Op`), and the system-key whitelist
+(:data:`SYSTEM_KEYS`).
+
+``FieldError`` is importable from here for callers inspecting a validation
+error's structured ``errors`` list, but is not part of the public ``__all__``.
+"""
+
+from __future__ import annotations
+
+from khora.filter.model import (
+    SYSTEM_KEYS,
+    DateOps,
+    Op,
+    RecallFilter,
+    RecallFilterUnsupportedError,
+    RecallFilterValidationError,
+    StringOps,
+)
+from khora.filter.model import FieldError as FieldError  # re-export (not in __all__)
+
+__all__ = [
+    "DateOps",
+    "Op",
+    "RecallFilter",
+    "RecallFilterUnsupportedError",
+    "RecallFilterValidationError",
+    "StringOps",
+    "SYSTEM_KEYS",
+]
--- a/src/khora/filter/model.py
+++ b/src/khora/filter/model.py
@@ -1,0 +1,555 @@
+"""Deterministic recall-filter model and validator.
+
+This module defines the public, MongoDB-flavored filter document accepted by
+recall operations. It is a *typed* pydantic model with named system-key fields
+(IDE autocomplete on the closed system-key whitelist, per-value-type operator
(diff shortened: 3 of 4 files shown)


filippetrovic:

Reviewed by an agent team across six dimensions (security, performance, correctness, tests, docs, devil's advocate) plus a dedicated security audit. The change is a self-contained, validation-only public model. Overall: **APPROVE** — the one higher-severity item (a docs/hygiene issue) has already been fixed on this branch; the remainder are non-blocking improvements with suggested follow-ups.

### HIGH
- **H1 [Docs] — RESOLVED on this branch.** Docstrings/comments carried internal cross-reference shorthands that don't belong in a public repository. They have been replaced with plain-English rule names (e.g. "operator-position closure", "operand opacity"); a grep confirms none remain in the committed source.

### MEDIUM
- **M1 [Security / Performance] Unbounded recursion in the metadata-grammar walk.** The recursive helpers (`_walk_predicate` / `_reject_nested_operator` / `_walk_operator` / `_walk_not_operand`) have no explicit depth bound, so an adversarially deep filter (hundreds of nesting levels) raises an uncaught `RecursionError` rather than a clean validation error. This is a *graceful* failure (catchable, no crash/segfault) and the public facade/gateway should bound request size regardless — so it's defense-in-depth, **not merge-blocking**. Suggested follow-up: add a depth cap that raises `RecallFilterValidationError` on excessive nesting.
- **M2 [Design] Two-representation fold + carrier-field approach.** Byte-stable round-trip is achieved via a `mode="before"` dot-path fold onto a reserved per-instance carrier + a `model_serializer` re-emit. It's recursion-safe and exhaustively tested, but adds a layer of model-lifecycle complexity. Defensible for v1; suggested follow-up: a short "internal, may be refactored post-v1" note near the carrier.

### LOW
- **L1 [Validation semantics] `$`-operators inside list / `$in` operands are silently accepted** (operand opacity — a list element is an opaque exact-array literal). This is the correct v1 behavior and consistent with bare-list / `$eq` operands; such constructs are rejected downstream at compile time. Worth a one-line doc note as a known user footgun.
- **L2 [API surface] `RecallFilterUnsupportedError` is exported before any code raises it.** Intentional forward-declaration so downstream callers can prepare handlers; acceptable, documented.
- **L3 [Performance] No fast-path for already-validated instances.** Each call re-validates; this is already noted as a future improvement. Non-blocking.
- **L4 [API surface] `FieldError` is importable but intentionally not in `__all__`** (a "gray export" for error introspection). Sufficient and clearly commented.

### Security audit — verdict: PASS
No injection, auth, crypto, network, file, redirect, or unsafe-deserialization surface (the module validates input; it does not build queries). `extra="forbid"` correctly prevents over-posting/mass-assignment. Error messages use static text + JSON-pointer paths — no raw user values or PII echoed. Only finding is M1 above (resource-exhaustion, defense-in-depth).

### What's clean
- **Correctness: CLEAN.** Per-key operator subsets, dot-path fold + byte-stable round-trip, operand opacity vs the operator-position-closure rule (incl. the list-vs-dict-nesting boundary), strict-bool `$exists`, `$date`/UTC normalization, unset-vs-explicit-null, empty-logical-array rejection, bare-value sugar — all verified against the contract.
- **Tests: CLEAN.** 112 cases, 100% acceptance-criteria coverage, byte-stable round-trip assertions, exhaustive operand-opacity boundary, consistent markers, no brittle patterns.
- **Scope discipline:** exactly the intended public surface (7 exports), validation-only — no query-compilation logic leaked in.

**Assessment: APPROVE.** No blocking findings. M1/M2 and the LOW items are recommended follow-ups, not merge gates.

filippetrovic:
Addressed review finding **M1** (unbounded recursion) in `a1f80a5`: added an iterative depth guard in the outermost validator that rejects filters nested beyond a fixed bound (100) with a clean `RecallFilterValidationError` (code `max_depth_exceeded`) — before pydantic's recursive model validation or the metadata-grammar walk recurse. An adversarially deep filter (metadata operand or nested `$and`/`$or`/`$not`) now fails as a structured validation error instead of an uncaught `RecursionError`. Covered by 3 new tests (deep metadata, deep logical, and a within-limit pass); suite is 115 green.

probablyrobot (approved):

PR #1002, opened 2026-06-05
Backfill khora_chunks denormalized columns and build selective filter indexes

filippetrovic:
## Summary
- Add a Postgres-only Alembic migration (`042`) that populates the eight nullable denormalized columns previously added to `khora_chunks` (`source_type`, `source_name`, `source_url`, `source_timestamp`, `external_id`, `content_type`, `source`, `title`) on existing rows from the parent `documents` row, so recall filters resolve on the chunk grain without a join.
- **Backfill** runs entirely in autocommit, **batched per namespace** (`SELECT DISTINCT namespace_id`, one `UPDATE` per namespace) with per-batch `COMMIT`s, so no single transaction spans the whole table — bounding MVCC bloat and advisory-lock hold time on large deployments. A `source_type IS NULL` sentinel makes the backfill idempotent and restartable.
- The `content_tsv` `BEFORE INSERT OR UPDATE` trigger is disabled around the backfill (re-enabled in a load-bearing `finally`) so it is not re-fired on rows whose content never changed. `source` / `external_id` are copied with `LEFT(col, 255)` to fit their narrower target columns, with a pre-update WARNING count so any truncation is auditable.
- Phase order: **backfill → `CREATE INDEX CONCURRENTLY` (five indexes) → `VACUUM (ANALYZE)`**. The five index names/definitions are byte-identical to the runtime store DDL so the two converge via `IF NOT EXISTS`; `source_url` / `source` / `title` are left unindexed. `downgrade()` drops only the five indexes (the data is removed when the prior migration drops the columns).
- Migration is guarded Postgres-only + `has_table("khora_chunks")`, mirroring the prior migration's lock-timeout and structured-logging conventions.

## Test plan
- [x] Unit tests pass — migration chain contiguity, single head, file-count/head bumps, SQLite no-op + clean downgrade (`70 passed` locally)
- [ ] Integration tests pass (Postgres, run in CI) — backfill happy path incl. `LEFT(,255)` truncation, multi-namespace batching, idempotent re-run, all five indexes present + valid (partial predicate on `source_timestamp`), TSV trigger re-enabled, `content_tsv` preserved, missing-table no-op
- [ ] Manual verification: `alembic upgrade head` then `downgrade` on a populated database

### DIFF
--- a/src/khora/db/migrations/versions/042_khora_chunks_backfill_denormalized.py
+++ b/src/khora/db/migrations/versions/042_khora_chunks_backfill_denormalized.py
@@ -1,0 +1,297 @@
+"""Backfill khora_chunks denormalized columns and build the filter indexes.
+
+Revision ID: 042_khora_chunks_backfill_denormalized
+Revises: 041_khora_chunks_denormalized_columns
+Create Date: 2026-06-05
+
+Migration 041 *added* eight nullable denormalized document-grained columns
+to the runtime-managed ``khora_chunks`` temporal store. This migration
+populates them on existing rows and creates the filterable-subset indexes
+so recall filters resolve without a join. Three phases, in order:
+
+* **PHASE 1 — backfill (autocommit, per-namespace, per-batch COMMIT).**
+  Copies ``source_type``, ``source_name``, ``source_url``,
+  ``source_timestamp``, ``external_id``, ``content_type``, ``source`` and
+  ``title`` down from the parent ``documents`` row (joined on
+  ``khora_chunks.document_id = documents.id``). The update is **batched
+  per namespace** — ``SELECT DISTINCT namespace_id`` then one ``UPDATE``
+  per namespace — so no single transaction spans the whole ~1M-row table:
+  each per-ns UPDATE auto-commits inside the autocommit block, bounding
+  open snapshots / MVCC bloat / advisory-lock contention. The
+  ``AND kc.source_type IS NULL`` sentinel makes the backfill **idempotent
+  and restartable** — re-running (or resuming after a crash mid-loop)
+  touches zero already-populated rows, because ``documents.source_type``
+  is ``NOT NULL`` so a backfilled chunk is always non-NULL there.
+
+  Two target columns are narrower than their ``documents`` source:
+  ``source`` (1024→255) and ``external_id`` (512→255). Both are copied
+  with ``LEFT(col, 255)`` rather than a bare cast — a value over 255 chars
+  would raise SQLSTATE 22001 and abort the batch; ``LEFT`` is
+  deterministic and never errors. The denormalized chunk copy is a
+  best-effort filter hint; the authoritative ``documents`` values (wider,
+  uniquely indexed) are untouched, so callers needing exact external_id
+  correctness join to ``documents``. A guarded ``count(*)`` before the
+  loop counts rows that will be truncated and logs them at WARNING
+  (``truncated_external_id`` / ``truncated_source``) so the truncation is
+  auditable.
+
+  The runtime ``BEFORE INSERT OR UPDATE`` trigger
+  ``khora_chunks_content_tsv_update`` recomputes ``content_tsv`` on every
+  UPDATE, including this backfill (which never touches ``content``). It is
+  disabled once before the loop and re-enabled in a ``finally`` on every
+  exit path. **The finally is load-bearing, not defensive:** the backfill
+  runs in autocommit, so the DISABLE commits immediately and there is no
+  outer transaction to roll it back. ``SET lock_timeout = '5s'`` bounds the
+  ``AccessExclusiveLock`` the DISABLE takes. The DISABLE/ENABLE pair is
+  guarded on ``information_schema.triggers`` and skipped if the trigger is
+  absent (a 041 sidecar apply before any runtime boot has no trigger yet —
+  and then no trigger fires on the UPDATE anyway).
+
+  **Crash caveat:** a hard kill (SIGKILL / pod OOM) *between* the committed
+  DISABLE and the ENABLE leaves the TSV trigger disabled until the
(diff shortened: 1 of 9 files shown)


filippetrovic:
## Multi-perspective review — consolidated (final)

Seven reviewers (security, performance, correctness, test, database, docs, devil's-advocate) reviewed the diff. CI is green (9/9).

### Overall assessment: ✅ APPROVE — no merge blockers

All acceptance criteria are met and CI passes. The findings below are **non-blocking follow-ups**; two were raised as "blockers" by the devil's advocate and are adjudicated below.

### Adjudicated "blockers" (downgraded)

- **Advisory lock held across the multi-minute backfill → other services fail to boot (DA, HIGH).** *Downgraded — premise is incorrect.* `run_migrations()` uses `pg_advisory_xact_lock`, which is **transaction-scoped**. The backfill runs inside `op.get_context().autocommit_block()`, which commits the wrapping transaction — **releasing the advisory lock** before the long backfill, not holding it. The Database Reviewer independently confirmed this. So deploys are *not* blocked by a held lock; the residual (a concurrent runner acquiring the lock during the unlocked backfill) is rendered safe by the idempotent `source_type IS NULL` sentinel + `CREATE INDEX … IF NOT EXISTS` and Alembic's revision linearization. This is the established pattern of precedents `029`/`039`.
- **No sub-batching within a 359K-row namespace / no progress logging (DA, HIGH).** *Downgraded to follow-up.* Per-namespace batching is exactly the ticket's acceptance criterion ("ns-batched UPDATE"); finer sub-batching was not required, and an id-range escape hatch is documented in the migration docstring. Per-batch progress logging is a reasonable **operability follow-up** (see L-list) but not a correctness defect.

### Findings

**LOW / recommended follow-ups (non-blocking)**
- **[Security] L1** — `042:219,226`: `DISABLE/ENABLE TRIGGER` uses an f-string with a module-constant trigger name. No injection is possible (the name is a hardcoded constant), but quoting the identifier — `... DISABLE TRIGGER "{_TSV_TRIGGER}"` — is a cheap defense-in-depth hardening for future edits.
- **[Test] L2** — orphaned chunk (a `khora_chunks` row whose `document_id` has no `documents` match) is correctly left NULL by the `kc.document_id = d.id` join — this is **intended** behavior (confirmed by the correctness reviewer) but is **untested**. A small `test_orphaned_chunks_remain_null` case would lock in the contract.
- **[Database] L3** — `042:133-142`: the pre-flight truncation `count(*)` joins `documents` without a `namespace_id` filter, so its WARNING reflects table-wide truncation rather than per-batch. Functionally correct (`documents.id` is globally unique); the log is advisory-only.
- **[Performance] L4** — `CREATE INDEX CONCURRENTLY` runs on the not-yet-vacuumed heap (VACUUM is last, per the required phase order); modest cost given the 5 low-cardinality `(namespace_id, X)` btrees. No action — intentional and AC-mandated.
- **[Docs] L5** — consider a short `CLAUDE.md` "Migrations & Schema" gotchas entry documenting this migration's non-obvious patterns (autocommit per-ns backfill, load-bearing trigger `finally`, hard-kill caveat), mirroring the `029` precedent.

**Verified — no defects**
- **Correctness**: trigger ENABLE reached on every exit path; autocommit per-ns batching commits each batch; `source_type IS NULL` sentinel sound (`documents.source_type` is NOT NULL); downgrade drops only the 5 indexes; exception path logs + re-raises. *(Zero findings.)*
- **Database**: 5 index names/defs byte-identical to the runtime DDL (converge via `IF NOT EXISTS`); partial `WHERE source_timestamp IS NOT NULL` present; `has_table` + Postgres-only guards; fresh-deploy/SQLite no-op.
- **Security**: all data-bound SQL is parameterized (`:ns`, `:trigger`); no secret exposure; `LEFT(,255)` truncates only the non-authoritative denormalized copy (the wider uniquely-indexed `documents` value is untouched).
- **Test**: 8 cases with the same CI-gated `_pg_reachable()` skip guard as the sibling migration test (Postgres cases execute in CI — confirmed green); strong assertions (LEFT truncation, partial index, `indisvalid`, `tgenabled='O'`, content_tsv preservation).
- **Docs**: module docstring matches the implementation across all three phases; CHANGELOG deferral to release-tag time is correct per repo policy.

### Recommended follow-ups (optional, non-blocking)
L1 (identifier quoting), L2 (orphaned-chunk test), L5 (CLAUDE.md note), and the DA's per-batch progress-logging suggestion are all sensible incremental improvements that can land in this PR or a follow-up. None gate merge.

probablyrobot (approved):

filippetrovic:
## Reconciled to the widened columns + re-chained

Following the decision to **widen** `khora_chunks.source` → `TEXT` and `external_id` → `VARCHAR(512)` (rather than truncate), this PR has been rebased and reconciled:

- **Re-chained** the migration onto the widened chain: renamed to `044_khora_chunks_backfill_denormalized`, `down_revision` now `043_khora_chunks_metadata_backfill` (it is the single chain head).
- **Full copy instead of truncation:** `external_id` and `source` are now copied verbatim (`= d.external_id` / `= d.source`) — the widened columns hold the full producer value, so every value stays filterable. Dropped the pre-flight truncation count and its WARNING-log branch.
- **Tests:** the migration test (renamed to `044`) seeds the chunks table with the widened column types, steps back to `043`, and asserts the long `source` / `external_id` are stored in full. Chain-head / migration-count expectations bumped to the new head across the bundling, sqlite, dream-runs, `041`, and `042`-widen migration tests.

Unchanged: the per-namespace batched `UPDATE`, the TSV-trigger disable/enable, the five `CREATE INDEX CONCURRENTLY` statements, the `VACUUM (ANALYZE)`, the `source_type IS NULL` sentinel, and the backfill → index → vacuum ordering.

CI is green (9/9).

PR #1003, opened 2026-06-05
Populate denormalized chunk columns from document attrs at ingest

filippetrovic:
## Summary
- Copy the eight document-grained provenance fields (`source_type`, `source_name`, `source_url`, `source_timestamp`, `external_id`, `content_type`, `source`, `title`) from the typed `Document` onto every persisted chunk at ingest, via a shared `document_denorm_fields()` helper used by the VectorCypher and Skeleton write paths and the ingest pipeline.
- Persist and read these columns on the `pgvector` and `sqlite_lance` temporal backends (the schema columns already landed previously).
- Add `occurred_at` to the public `Chunk` DTO and stop coalescing it with `source_timestamp` in the temporal-chunk adapter, so the producer's verbatim time (`source_timestamp`) and the chunk event-time (`occurred_at`) are carried distinctly end-to-end.
- Build `RecallChunk.occurred_at` from the chunk's own `occurred_at`, falling back to `source_timestamp` only when no chunk-level event-time exists.
- Out of scope (deferred to follow-ups): Neo4j chunk-node properties and the SurrealDB / Weaviate temporal backends; their absence is intentional and the `None`-fallback keeps them functionally safe.

## Test plan
- [x] `ruff check` + `ruff format --check` + `ty check src/` clean
- [x] New contract test asserts `source_timestamp` and `occurred_at` survive ingest→recall as **distinct** values (parametrized over VectorCypher + Skeleton, vector + keyword modes); it fails on the pre-change behavior
- [x] Per-backend write/read-back coverage for the eight columns (pgvector + sqlite_lance)
- [x] Targeted + engine unit tests pass locally (adapter, recall projection, pgvector row mapping)
- [ ] Full PostgreSQL / Neo4j integration matrix — exercised in CI

### DIFF
--- a/src/khora/core/models/document.py
+++ b/src/khora/core/models/document.py
@@ -172,6 +172,9 @@ class Chunk:
     # Timestamps
     created_at: datetime = field(default_factory=lambda: datetime.now(UTC))
     source_timestamp: datetime | None = None
+    # Chunk event-time, distinct from source_timestamp (the producer's
+    # verbatim time). NULL when the chunk carries no event-time of its own.
+    occurred_at: datetime | None = None
     # Reinforcement-on-recall (#855). NULL until the chunk is first
     # returned by a recall path that has reinforcement enabled.
     last_accessed_at: datetime | None = None
--- a/src/khora/engines/chronicle/engine.py
+++ b/src/khora/engines/chronicle/engine.py
@@ -1896,7 +1896,7 @@ class ChronicleEngine:
                 content=chunk.content,
                 score=score,
                 created_at=chunk.created_at,
-                occurred_at=chunk.source_timestamp,
+                occurred_at=(chunk.occurred_at if chunk.occurred_at is not None else chunk.source_timestamp),
                 chunker_info=chunk.chunker_info or {},
             )
             for (chunk, _), score in zip(chunks_with_scores, normalized_chunk_scores, strict=False)
--- a/src/khora/engines/skeleton/backends/__init__.py
+++ b/src/khora/engines/skeleton/backends/__init__.py
@@ -17,6 +17,7 @@ from khora.core.models.document import Chunk
 
 if TYPE_CHECKING:
     from khora.config import KhoraConfig
+    from khora.core.models.document import Document
 
 
 @dataclass
@@ -53,6 +54,26 @@ class TemporalChunk:
     content_type: str | None = None
     source: str | None = None
     title: str | None = None
+
+
+def document_denorm_fields(document: Document) -> dict[str, Any]:
+    """Return the denormalized document-grained chunk fields from a Document.
+
+    Copies the eight provenance fields off the typed Document so chunk
+    builders can stamp them onto a :class:`TemporalChunk` without repeating
+    the mapping at every write site. ``source_timestamp`` is the producer's
+    verbatim time, kept distinct from the chunk event-time ``occurred_at``.
+    """
+    return {
+        "source_type": document.source_type,
+        "source_name": document.source_name,
+        "source_url": document.source_url,
+        "source_timestamp": document.source_timestamp,
+        "external_id": document.external_id,
+        "content_type": document.content_type,
+        "source": document.source,
+        "title": document.title,
+    }
 
 
 @dataclass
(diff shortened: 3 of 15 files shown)


filippetrovic:
# Multi-Perspective Review — Consolidated

Eight independent reviewers (security, performance, correctness, tests, database, observability, docs, devil's advocate) plus an OWASP-focused security audit pass.

**Overall assessment: APPROVE.** No blocking code defects. The write-path denormalization and the distinct `occurred_at` / `source_timestamp` carry are correct; INSERT/column arity, runtime-vs-migration schema consistency, and read-back paths were verified across both backends; the distinct-values tripwire genuinely fails on the pre-change behavior. Findings below are hardening, test-coverage, docs, and deferred-scope notes — none block merge.

## HIGH
- None confirmed. (Two findings were initially raised at HIGH and downgraded on adjudication — see *Notes* below.)

## MEDIUM
- **M1 [Security]** `core/models/document.py` — Six provenance string fields (`source_type`, `source_name`, `source_url`, `source`, `author`, `source_system`) are unbounded; a pathological document could bloat chunk rows. These were already unbounded on the document model, so this is not a regression, but consider max-length validators mirroring the existing `external_id` pattern. SQL is parameterized, so there is no injection risk.
- **M2 [Tests]** `engines/vectorcypher/retriever.py` (recency channel) and `pipelines/flows/ingest.py` (batch ingest) — the `source_timestamp` fallback on the recency channel and the batch-ingest write path are correct but lack direct test coverage. Recommend targeted cases.
- **M3 [Docs]** `docs/architecture/storage-backends.md` — the new `Chunk.occurred_at` field and the event-time-then-producer-time contract (`occurred_at` falls back to `source_timestamp`) are not documented in the public architecture guide. Add a short subsection.
- **M4 [Perf/DB]** `engines/skeleton/backends/sqlite_lance.py` — denormalized columns are written to the SQLite relational side only, not the LanceDB vector table (which holds only vectors). Not a defect today; if future filter-pushdown needs them on the vector store, a backfill would be required. Worth a one-line comment.

## LOW
- **L1 [Correctness]** `engines/vectorcypher/retriever.py` (SurrealDB entity-fetch fallback) — on this path `occurred_at` and `source_timestamp` collapse (both mirror the chunk's `source_timestamp`). This is pre-existing behavior on the **deferred** SurrealDB path (that backend does not yet persist a distinct chunk event-time); reading the chunk's `occurred_at` there would yield `None` and regress. Correctly handled by the deferred backends slice, not this change.
- **L2 [Perf]** Per-insert / read-back overhead of the 8 fields is negligible; the `document_denorm_fields()` helper is good DRY (single source for the 8-field map across the write sites).

## Deferred scope (intentional — not defects)
- Neo4j chunk-node 8-key properties and the SurrealDB / Weaviate temporal backends are out of scope by design; reads fall back safely. Recommend tracking the Neo4j chunk-property population as a follow-up so it isn't lost.

## Security Audit (OWASP-focused)
- **Injection:** PASS — all writes parameterized (SQLite `?` placeholders; pgvector via SQLAlchemy bound params).
- **Secrets / typed-secret policy:** PASS — no secret-named fields introduced; the 8 fields are non-sensitive provenance.
- **Sensitive-data exposure:** PASS — fields already existed on the document model; denormalization adds no new exposure surface.
- **Dependencies / cryptography / auth:** N/A — no dependency, crypto, or auth/authz changes in the diff.
- **Resource exhaustion:** see M1 (unbounded provenance strings).

## Notes on downgraded findings
- The correctness reviewer initially raised **L1** at HIGH; on reading the code it is a pre-existing limitation of the deferred SurrealDB path (the suggested fix would yield `None` and regress), so it is reclassified LOW.
- A test reviewer raised the recency-channel gap (**M2**) at HIGH; the code path is correct (verified), so it is a test-coverage gap, not a defect — reclassified MEDIUM.
- An earlier temporal-decay regression hypothesis was empirically refuted: the decay contract test passes against the change, and the decay path reads chunks from the main chunks table (whose `source_timestamp` is unchanged by this PR), not the temporal-chunk adapter.

probablyrobot (approved):

filippetrovic:
Rebased onto the latest `main` (clean, no conflicts). This picked up the companion change that widens the `khora_chunks` `source` → `TEXT` and `external_id` → `VARCHAR(512)` columns to match the `documents` widths — the prerequisite for this write-path populating those columns without `StringDataRightTruncation`. That also addresses the review's MEDIUM note about unbounded provenance strings for `source` / `external_id` / `source_url` (now `TEXT`/wide). No code changes were needed on this branch — the two changes are complementary (widths vs. population). Affected unit/contract tests re-run green locally after the rebase (72 passed).

probablyrobot (approved):

PR #1004, opened 2026-06-05
Carry document metadata onto Skeleton chunks and backfill existing chunk metadata

filippetrovic:
## Summary
- Skeleton ingest now spreads the parent document's free-form `metadata` onto each chunk's `metadata` dict, with chunk bookkeeping keys (`chunk_index`/`start_char`/`end_char`) still taking precedence on collision. This mirrors the VectorCypher engine and lets `metadata.<path>` recall filters match on the Skeleton-pgvector backend (previously they matched nothing because Skeleton chunks only carried bookkeeping keys).
- Both ingest paths are patched: single-document (`_process_document`) and batch (`remember_batch`).
- New Postgres-only Alembic migration backfills existing `khora_chunks` rows by merging `documents.metadata` into each chunk's `metadata` (chunk keys win, matching the ingest order). The merge is idempotent via a JSONB containment guard (`NOT (c.metadata @> d.metadata)`), namespace-batched to bound per-statement cost, and disables the `content_tsv` trigger for the duration (the update never touches `content`, so re-deriving the tsvector would be wasted work). Lock-timeout-bounded with structured progress logging, dialect-gated (no-op on SQLite), and a documented irreversible no-op downgrade.
- The migration scopes strictly to the `metadata` JSONB column — no schema/column/index changes.

## Test plan
- [x] Unit tests pass (new `TestChunkMetadataMerge` covers both ingest paths and the bookkeeping-key collision case; migration-chain head pins updated; single Alembic head verified)
- [x] Lint + format clean (`ruff check`, `ruff format --check`)
- [ ] Full integration/e2e suite + Postgres migration run — gated by CI (requires Docker + optional extras not available in the authoring sandbox)

### DIFF
--- a/src/khora/db/migrations/versions/043_khora_chunks_metadata_backfill.py
+++ b/src/khora/db/migrations/versions/043_khora_chunks_metadata_backfill.py
@@ -1,0 +1,213 @@
+"""Backfill khora_chunks.metadata by merging in the parent document's metadata.
+
+Revision ID: 043_khora_chunks_metadata_backfill
+Revises: 042_widen_khora_chunks_source_external_id
+Create Date: 2026-06-05
+
+Data change only — NO schema change. Every existing ``khora_chunks`` row
+gets the parent ``documents.metadata`` JSONB merged into its own
+``metadata``, with the chunk's own keys taking precedence on conflict
+(``d.metadata || c.metadata`` — the right-hand operand wins in Postgres
+JSONB concatenation). New chunks written after this migration already
+carry the merged metadata from the ingest path; this migration only
+catches chunks that predate that behavior.
+
+Merge predicate / idempotency: the UPDATE is guarded by
+``NOT (c.metadata @> d.metadata)`` so a chunk whose metadata already
+contains every parent key/value is skipped. Re-running the migration (or
+re-running within the same transaction) therefore updates zero rows once
+the merge has landed — the operation converges. The ``@>`` containment
+check is conservative: a chunk that overrides a parent key with a
+different value does NOT contain the parent pair, so it is re-visited, but
+the merge is stable (chunk key wins again, producing the identical row),
+so this is correct, just a touch less selective.
+
+Parent filter: only documents with a non-empty ``metadata`` participate
+(``d.metadata IS NOT NULL AND d.metadata <> '{}'::jsonb``). A parent with
+no metadata contributes nothing, so those joins are pruned up front.
+
+Scale (~1M chunk rows): the backfill runs namespace-batched — one UPDATE
+per distinct ``khora_chunks.namespace_id`` — rather than as a single
+table-wide statement. Per-namespace statements take an index scan on the
+``namespace_id`` index instead of a full-table hash join against
+``documents``, keep each statement's working set bounded, and let the
+structured log report progress per namespace. Note the whole Alembic
+chain runs inside ONE transaction (``env.py:do_run_migrations`` wraps
+``context.run_migrations()`` in a single ``context.begin_transaction()``),
+so the row locks taken by each batch are held until the chain commits —
+batching bounds per-statement cost and planning, not lock-hold duration.
+
+Trigger suppression: ``khora_chunks`` carries a ``BEFORE INSERT OR UPDATE``
+trigger (``khora_chunks_content_tsv_update``) that recomputes
+``to_tsvector('english', content)`` on every updated row. This backfill
+touches only ``metadata`` and never ``content``, so re-deriving the
+tsvector for every row would be wasted CPU on a ~1M-row table. The trigger
+is disabled for the duration of the backfill and re-enabled in a
+``finally`` so it is restored on every exit path (success, lock-timeout,
+or any error). ``DISABLE TRIGGER`` / ``ENABLE TRIGGER`` is itself
+transactional, so a rolled-back migration leaves the trigger enabled.
+
+Lock-timeout safety: a Postgres-only ``SET lock_timeout = '5s'`` is issued
+before the trigger toggle so the brief ``ACCESS EXCLUSIVE`` lock that
+``ALTER TABLE ... DISABLE TRIGGER`` acquires, and every row lock the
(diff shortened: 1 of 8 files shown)


filippetrovic:
## 🧭 Multi-perspective review — Overall: ✅ APPROVE

Seven independent reviewers (security, performance, correctness, tests, database/migration, docs, devil's advocate) plus a focused security audit. No CRITICAL or HIGH findings. The change is correct, production-safe, and well-tested. The remaining items are advisory tradeoffs and follow-ups — the core behavior (carry document metadata onto chunks for `metadata.<path>` filtering, mirroring the VectorCypher engine; idempotent namespace-batched backfill) is sound.

### MEDIUM
- **M1 — [Performance / Devil's Advocate]** `engines/skeleton/engine.py` (both ingest sites) + migration — denormalizing the full `documents.metadata` onto every chunk's JSONB makes `khora_chunks.metadata` storage scale with chunk count, not document count. This is the **intended design** (parity with VectorCypher, and required so `metadata.<path>` predicates can read the value off the chunk row), but the storage tradeoff is real. *Recommend:* document the tradeoff and watch `khora_chunks` table-size growth post-deploy.
- **M2 — [Database / Performance / Devil's Advocate]** `042_*.py` — the Alembic chain runs in a single transaction, so row locks taken by the backfill are held until the whole chain commits; the `lock_timeout=5s` bounds lock *acquisition*, not *hold* duration. This is honestly called out in the docstring and is acceptable for a one-time data backfill, but for very large tables a future out-of-band backfill job would avoid holding locks across the chain.
- **M3 — [Docs]** `CLAUDE.md` gotchas don't yet mention that Skeleton chunks now denormalize document metadata or that this migration backfills existing rows. Worth a one-line note to prevent schema-assumption drift.
- **M4 — [Tests]** No direct test of the backfill migration's merge/idempotency at the data level (the `@>` guard, re-run convergence). The application-level merge is well covered by the new `TestChunkMetadataMerge`; the migration is exercised only via the chain. *Recommend* a follow-up integration test that seeds rows, runs the migration twice, and asserts convergence.

### LOW
- **L1 — [Correctness / Devil's Advocate]** The JSONB `||` (and the Python `{**doc_metadata, ...}` spread) is a shallow merge — nested dict/list values are not deep-merged. This is **by design and consistent** between ingest and backfill (a deep-merge would make backfilled rows diverge from freshly-ingested rows); flagged as informational only.
- **L2 — [Performance]** `SELECT DISTINCT namespace_id FROM khora_chunks` is a full scan on a large table; one-time and lock-timeout-bounded, but worth noting it assumes a usable `namespace_id` index.
- **L3 — [Devil's Advocate]** The migration docstring is thorough but could state *why* the full-metadata spread was chosen over a key whitelist (answer: arbitrary `metadata.<path>` filters need all keys, and whitelisting would break parity).

### Verified correct (no action)
- Merge direction `d.metadata || c.metadata` (chunk keys win) matches the ingest spread order; bookkeeping keys (`chunk_index`/`start_char`/`end_char`) always win — covered by a mutation-sensitive collision test on **both** ingest paths.
- Idempotency guard `NOT (c.metadata @> d.metadata)`, dialect gating (Postgres-only, SQLite no-op), trigger DISABLE/ENABLE in `try/finally`, `lock_timeout` + `55P03` detection, `has_table` guards on both tables, single Alembic head (`down_revision` correct) — all verified by the correctness and database reviewers.
- **Scope** is correct: this PR carries only the free-form `metadata` JSONB; the typed denormalized columns were a separate prior change.

## 🔒 Security Audit — Verdict: PASS
- **Injection:** the merge UPDATE binds `namespace_id` as a parameter; the `SELECT DISTINCT` is static SQL; the `ALTER TABLE … DISABLE/ENABLE TRIGGER` f-string interpolates a **hardcoded constant** (DDL identifiers can't be parameterized — correct pattern). No user-controlled SQL.
- **Data exposure:** copying document metadata onto chunks is intentional and tested; same trust boundary as the existing `documents.metadata`.
- **Supply chain:** no dependency/manifest changes. **Secrets/crypto:** none. No CRITICAL/HIGH/MEDIUM security findings.

_Generated by an automated multi-agent review. MEDIUM/LOW items are advisory — no blocking issues._

probablyrobot (approved):

filippetrovic:
🔄 **Rebased onto latest `main`** (picks up #1001 RecallFilter, #1003 ingest denormalization, #1005 column widening).

Adaptations made during the rebase:
- **Migration renumbered to avoid a collision.** `main` added `042_widen_khora_chunks_source_external_id` (off `041`); this PR's backfill migration was also `042` off `041`, which would have created two Alembic heads. It is now **`043_khora_chunks_metadata_backfill`**, re-parented `down_revision = 042_widen_khora_chunks_source_external_id`. Verified a single head (`alembic heads` → `043`) and a linear chain `041 → 042_widen → 043`.
- **Ingest co-exists with #1003.** Both `TemporalChunk` sites now keep the `{**doc_metadata, "chunk_index": …}` metadata spread (this PR) *and* the `**document_denorm_fields(document)` kwarg (#1003) — the typed denormalized columns and the free-form `metadata` JSONB are populated independently.
- **Migration-chain test fixtures updated:** head-pin assertions → `043_khora_chunks_metadata_backfill`; bundled-migration count `43 → 44`.

Locally green: ruff clean; SQLite migration-chain tests (head = `043`), the file-count test, and `TestChunkMetadataMerge` (both ingest paths) all pass. Full suite gated by CI.

filippetrovic:
Follow-up to the rebase: also updated `tests/integration/db/test_migration_042_widen_chunks_source_external_id.py` (added by #1005) — its `_HEAD` constant and an `upgrade("head")` chain-reaches-head assertion pointed at `042_widen_…`, which my new `043` head supersedes (→ `043_khora_chunks_metadata_backfill`). Also added the always-present `metadata JSONB` column to that file's synthetic legacy `khora_chunks` DDL so the `043` backfill runs against it on Postgres. (Two earlier `test`-job reds were unrelated flaky xdist worker crashes in recall tests; this was the one real adaptation gap.)

PR #1005, opened 2026-06-05
Widen khora_chunks source and external_id columns to documents widths

filippetrovic:
## Summary
- The denormalized `source` and `external_id` columns on `khora_chunks` were introduced one `varchar(255)` narrower than their parent `documents` columns (`documents.source` is `varchar(1024)`, `documents.external_id` is `varchar(512)`). An upcoming write-path change and one-shot backfill copy these fields down from `documents`, which would raise `StringDataRightTruncation` on long values.
- Add a Postgres-only catalog migration that widens `khora_chunks.source` → `TEXT` and `khora_chunks.external_id` → `VARCHAR(512)`. Both columns are empty today, so this is a no-rewrite catalog change (instant on any table size). It mirrors the preceding migration's guards: dialect-gated to Postgres, `has_table` guard, `lock_timeout` set before DDL, and structured apply-logging.
- `downgrade` is NULL-safe and guards against narrowing-truncation: it probes for any value longer than 255 chars and skips the narrowing (logging a warning) rather than destroying data.
- Update the runtime `khora_chunks_table` definition so fresh deploys and the embedded SQLite fixture get the correct widths directly.
- Bump the affected migration head-pin and versions-directory file-count test assertions to account for the new revision.

## Test plan
- [x] Unit tests pass (migration bundling/file-count, SQLite migration head-pin, migration drift/contiguity)
- [x] Integration tests pass: new migration test covers upgrade widths (via `information_schema`), a clean downgrade, a guarded downgrade that inserts an over-length value and asserts the narrowing is skipped, the table-absent no-op, and the SQLite early-return
- [x] Single Alembic head; chain remains linear

### DIFF
--- a/src/khora/db/migrations/versions/042_widen_khora_chunks_source_external_id.py
+++ b/src/khora/db/migrations/versions/042_widen_khora_chunks_source_external_id.py
@@ -1,0 +1,204 @@
+"""Widen two denormalized ``khora_chunks`` columns to match their source widths.
+
+Revision ID: 042_widen_khora_chunks_source_external_id
+Revises: 041_khora_chunks_denormalized_columns
+Create Date: 2026-06-05
+
+Schema changes:
+
+* ``khora_chunks.source`` ``VARCHAR(255)`` -> ``TEXT``
+* ``khora_chunks.external_id`` ``VARCHAR(255)`` -> ``VARCHAR(512)``
+
+Both columns were added (empty, no backfill) by the immediately preceding
+migration 041. An upcoming write-path change plus a one-shot backfill will
+copy these two fields down from the parent ``documents`` table, whose source
+columns are wider than the 255-char placeholder 041 used:
+
+  - ``documents.source`` is ``VARCHAR(1024)`` -> chunk side widened to ``TEXT``
+    (``TEXT`` >= 1024, so no value can truncate).
+  - ``documents.external_id`` is ``VARCHAR(512)`` -> chunk side widened to the
+    matching ``VARCHAR(512)``.
+
+Without this widening the backfill ``UPDATE khora_chunks ... FROM documents``
+would raise ``StringDataRightTruncation`` on long values. The columns are
+EMPTY today, so widening a ``varchar`` (and ``varchar`` -> ``text``) is a pure
+catalog change in Postgres -- no table rewrite, instant on any table size,
+zero risk.
+
+Lock-timeout safety: a Postgres-only ``SET lock_timeout = '5s'`` is issued
+**before any DDL** so the ``ALTER COLUMN`` AccessExclusiveLock acquisition is
+bounded, mirroring migration 041. The setting is scoped to the migration's
+enclosing transaction (env.py wraps the whole upgrade in a single
+transaction; the SET expires at COMMIT). On lock-timeout the upgrade logs
+``khora.migration.applied`` at ERROR with ``lock_timeout_tripped=True``
+(detected via the Postgres SQLSTATE ``55P03`` / ``lock_not_available`` on
+``OperationalError.orig.pgcode``); any other ``OperationalError`` logs
+``lock_timeout_tripped=False`` so dashboards don't conflate deadlocks /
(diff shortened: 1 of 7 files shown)


filippetrovic:

Reviewed across 7 perspectives (database, correctness, tests, security, performance, docs, devil's advocate) plus a security audit pass.

**Overall assessment: APPROVE — no blockers.** The change is correct, surgical, and well-tested. All findings below are advisory (design notes / optional polish); none block merge. The migration meets its stated acceptance criteria, the runtime table definition matches, the Alembic chain stays linear with a single head, and the change's own test suite passes (6 new migration tests + the updated denormalized-columns tests).

### CRITICAL / HIGH
None.

### MEDIUM (advisory)
- **M1 — Downgrade narrows both-or-nothing (correctness).** `downgrade()` skips narrowing *both* `source` and `external_id` when *either* holds an over-length value, rather than per-column. This is **safe and intentional**: it never truncates, the columns are empty today, downgrades are rare/manual, and "skip if data would truncate" is exactly the specified behavior. A per-column split is a possible refinement but adds complexity beyond what's needed. Left as-is by design.
- **M2 — No `external_id`-specific over-length downgrade test (correctness/tests).** The guarded-downgrade test inserts a 300-char `source`; the `external_id` arm shares the same `OR`-probe branch, so it's logically exercised, but an `external_id`-specific case would add direct coverage. Optional.
- **M3 — `source` → `TEXT` vs `VARCHAR(1024)` (devil's advocate).** The chunk column is now wider (unbounded) than its `documents.source` parent (`VARCHAR(1024)`). This is harmless (the value is gated upstream) and `TEXT` was an explicitly sanctioned choice; `TEXT ≥ 1024` satisfies the width criterion. Could note the future-proofing intent in the docstring. Optional.
- **M4 — `documents.source` width divergence note (devil's advocate).** The denormalized copy diverging in type from its source-of-truth column is design debt worth a one-line comment, not a defect.

### LOW (advisory)
- **L1 — Concurrent check-and-alter window on downgrade (devil's advocate).** The over-length probe is point-in-time, but `env.py`'s `pg_advisory_xact_lock` serializes migration transactions, so the window is closed in practice. No change needed.
- **L2 — `lock_timeout_tripped` logging nuance (devil's advocate).** Non-`55P03` failures in the protected window log `lock_timeout_tripped=False` (by design — only the specific SQLSTATE flags a true lock timeout). Telemetry-only.
- **L3 — SurrealDB equivalence (devil's advocate).** Verified **N/A**: `khora_chunks` exists only in the Postgres/SQLite temporal stores; the SurrealDB backend has no `khora_chunks` table (its `source`/`external_id` live on the `document` table as unbounded `option<string>`). SQLite ignores `VARCHAR` lengths (type affinity), so the Postgres-only scope is correct and complete.

### Test-fixture note (reviewed, no change)
The devil's advocate flagged the `test_migration_041` fixture reset changing from `WHERE version_num = '041…'` to `WHERE version_num != '040…'`. This broadening is **required** (the chain head moved to 042, so an equality-on-041 reset would no longer fire) and is **more future-proof** than the suggested explicit `IN ('041…','042…')` allowlist, which would need editing on every future migration. Kept as-is.

### Security audit
Clean. Static, parameterless SQL (`SET lock_timeout`, the length probe) — no injection surface; no user input; no secrets in the `khora.migration.applied` log fields or the downgrade warning; no Pydantic models (ADR-style secret-typing N/A); no dependency, crypto, auth, or network-surface changes.

### Verified correct
- Widths: `source` → `TEXT` (≥ parent 1024), `external_id` → `VARCHAR(512)` (= parent 512).
- Catalog-only / no-rewrite ALTERs; Postgres-only + `has_table` guards; `lock_timeout` set before DDL; `55P03` detection + structured logging mirroring the preceding migration.
- NULL-safe downgrade truncation guard (`length(NULL)` is NULL).
- Single linear Alembic head; runtime `khora_chunks_table` def synchronized; other six denormalized columns untouched.
- OSS hygiene: no internal identifiers in any changed file.

probablyrobot (approved):

PR #1007, opened 2026-06-05
Add internal filter AST and compiler-dispatch infrastructure

filippetrovic:
## Summary

Adds the internal compilation layers that backend query compilers will plug into. Nothing here is part of the public package surface — it all lives under `khora.filter` and is intentionally excluded from the top-level `__all__`.

- **`khora/filter/ast.py`** — `parse_to_ast` lowers an already-validated `RecallFilter` into a canonical, backend-agnostic `FilterNode` AST:
  - dot-path metadata keys resolve to path-segment tuples (never dot-strings);
  - all sugar is desugared (bare value → `$eq`, bare list → exact-array `$eq`, `$nor` → `$not($or)`, implicit sibling-key AND made an explicit `$and`);
  - logical nesting is normalized (same-operator flatten + single-child collapse) so two ways of writing the same filter produce one tree;
  - comparison operands stay opaque (a dict operand is data, never re-parsed as a clause).
  - `canonical_hash` derives an order-stable cache key — commutative `$and`/`$or` are order-insensitive, while list / exact-array / `$not` operands preserve order.
- **`khora/filter/context.py`** — `CompileContext` (engine-supplied per-compile-pass context: backend target, table alias, bind-param namespace, optional system-key→column mapping, backend capabilities, unsupported-node policy), `SchemaCapabilities` flags, and a `CompileError` hierarchy.
- **`khora/filter/registry.py`** — `CompilerRegistry`, a thread-safe, process-wide compiler dispatch keyed by `(engine_id, storage_target)` and empty at import, plus the `CompiledFilter` / `CompilerFn` protocol types.

No backend compilers, no engine wiring, and no schema changes are included — those are separate, follow-on changes. The validation model is untouched.

## Test plan
- [x] Unit tests pass — `tests/recall/test_filter_ast.py`, `test_compiler_registry.py`, `test_compile_context.py` (lowering invariants, hash stability, context defaults, registry round-trip + thread-safety, internal-only contract)
- [x] `make format` clean; `make lint` (ruff + ty) clean
- [x] Full `make test` green, total coverage 82.23%
- [ ] CI green

### DIFF
--- a/src/khora/filter/__init__.py
+++ b/src/khora/filter/__init__.py
@@ -12,6 +12,37 @@ error's structured ``errors`` list, but is not part of the public ``__all__``.
 
 from __future__ import annotations
 
+# Internal compilation seam (the AST + compiler layers). These are reachable as
+# ``khora.filter.X`` for khora's own engines/compilers, but are deliberately NOT
+# in ``__all__`` and NOT re-exported from ``khora.__init__`` — promoting them to
+# a public plugin seam is deferred to a future improvement.
+from khora.filter.ast import (  # noqa: F401
+    DateLiteral as DateLiteral,
+)
+from khora.filter.ast import (
+    FilterClause as FilterClause,
+)
+from khora.filter.ast import (
+    FilterNode as FilterNode,
+)
+from khora.filter.ast import (
+    FilterOp as FilterOp,
+)
+from khora.filter.ast import (
+    canonical_hash as canonical_hash,
+)
+from khora.filter.ast import (
+    parse_to_ast as parse_to_ast,
+)
+from khora.filter.context import (  # noqa: F401
+    CompileContext as CompileContext,
+)
+from khora.filter.context import (
+    CompileError as CompileError,
+)
+from khora.filter.context import (
+    SchemaCapabilities as SchemaCapabilities,
+)
 from khora.filter.model import (
     SYSTEM_KEYS,
     DateOps,
@@ -22,6 +53,15 @@ from khora.filter.model import (
     StringOps,
 )
 from khora.filter.model import FieldError as FieldError  # re-export (not in __all__)
+from khora.filter.registry import (  # noqa: F401
+    CompiledFilter as CompiledFilter,
+)
+from khora.filter.registry import (
+    CompilerFn as CompilerFn,
+)
+from khora.filter.registry import (
+    CompilerRegistry as CompilerRegistry,
+)
 
 __all__ = [
     "DateOps",
--- a/src/khora/filter/ast.py
+++ b/src/khora/filter/ast.py
@@ -1,0 +1,630 @@
+"""Canonical filter AST (Layer 3) — ``@internal``.
+
+The AST is the desugared, normalized intermediate form that backend compilers
+(Layer 4) consume. It sits between the public :class:`~khora.filter.RecallFilter`
+pydantic model (Layer 2, the validated wire contract) and the per-backend
+compilers (Layer 4). :func:`parse_to_ast` lowers a *validated* ``RecallFilter``
+into this form; :func:`canonical_hash` derives a stable cache key from it.
+
+``@internal``. Nothing here is exported from :mod:`khora.__init__`. The AST is
+``__all__``'d under :mod:`khora.filter` only and is free to evolve in minor
+releases (a new node kind for ``$regex`` later does not break the public
+surface). Backend compilers reach it through the (internal) ``CompileContext``;
+callers never see it.
+
+Two lowering invariants the rest of the subsystem depends on — **compilers
+never see dot-strings or sugar**:
+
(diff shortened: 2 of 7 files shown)


filippetrovic:

Six independent reviewers (Security, Performance, Correctness, Test, Docs, Devil's Advocate) plus a dedicated security audit. **Overall assessment: ✅ APPROVE** — no CRITICAL or HIGH blocking findings. Full CI is green (9/9 checks). The items below are documentation clarifications, a deferrable micro-optimization, and one intentional-by-scope design note.

### CRITICAL
_None._

### HIGH
_None._

### MEDIUM

- **M1 · [Correctness/Docs]** `src/khora/filter/ast.py` (`_normalize` docstring vs impl ~L347) — A single-child **`OR`** whose child is a leaf clause also collapses to `AND([leaf])` (the op is normalized to AND). The behavior is correct; the docstring only mentions the AND case. Clarify that both single-leaf AND **and** OR collapse to `AND([leaf])`.
- **M2 · [Correctness/Docs]** `src/khora/filter/ast.py` (`canonical_hash` docstring ~L567–583) — The docstring is ambiguous about where key-sorting happens. Children are sorted by their canonical serialization; operand-internal dict keys are sorted separately in `_canonicalize_operand`. Behavior is correct; tighten the wording.
- **M3 · [Performance]** `src/khora/filter/ast.py` (~L605) — The commutative-children sort calls `json.dumps` once per child as the sort key, then serializes again for the final hash (k+1 serializations). Minor allocation churn on the recall path; typical sibling counts are 2–4. Optional: memoize the per-child serialization. Non-blocking.
- **M4 · [Test]** `tests/recall/test_compiler_registry.py` (~L200–225) — The concurrent conflicting-register test asserts a single winner exists, but could be stricter (assert every losing thread saw a clean conflict error and the registry is left in a well-defined state). Current coverage is adequate; this is a hardening opportunity.
- **M5 · [Design / Devil's Advocate]** The AST invariants (uniform `FilterNode` root, `AND([leaf])` wrapping), `SchemaCapabilities` flag set, `canonical_hash`, and the process-wide `CompilerRegistry` ship **ahead of any backend compiler that consumes them** in this PR. This is **intentional**: this slice is the foundational seam that unblocks the per-backend compilers to be built in parallel. Flagged so the first compiler PR validates these against a real consumer (capability dimensions may expand; confirm the hash is actually used as a cache key; confirm the root invariant helps rather than adds wrapper overhead). The empty-at-import registry + manual `register()` is a deliberate seam, but note the dependency-injection/test-isolation trade-off (mitigated today by the `_clear()` test hook).

### LOW

- **L1 · [Performance]** `ast.py` (~L605) — Add a brief comment noting the sort key is a deterministic JSON serialization (stability guarantee).
- **L2 · [Correctness]** `src/khora/filter/registry.py` (`_clear`, ~L160) — Add an explicit "test-only; drops all registrations" docstring so the escape hatch isn't mistaken for production API.
- **L3 · [Correctness]** `src/khora/filter/context.py` (~L127) — Briefly document the `SchemaCapabilities.DEFAULTS` frozen class-var default pattern.
- **L4 · [Test]** `canonical_hash` accepts a bare `FilterClause` root; since the compiler input contract is `FilterNode`, ensure the engine/compiler boundary enforces that (parse_to_ast already always returns a `FilterNode`). No negative `DateLiteral` test (malformed `$date` is rejected upstream by the validator) — acceptable.

### Security Audit

**Verdict: PASS.** Scanned the diff across all 16 categories (injection, authn/authz, crypto/secrets, access control, SSRF, deserialization, file/path, ReDoS, supply chain, boundary validation, …). No findings.
- No untrusted-input trust boundary in the diff — `parse_to_ast` operates only on an already-validated filter; recursion is bounded upstream (`_MAX_FILTER_DEPTH=100`).
- No `eval`/`exec`/`subprocess`, no network/SSRF sinks, no deserialization, no file/path handling, no regex construction.
- `hashlib.sha256` is used solely for **cache-key derivation** (not an integrity/auth control) — appropriate, no security overreach.
- No new dependencies, manifest, or CI/workflow changes.
- ADR-084 secret-typing: no secret-named fields introduced.

### Notes

- **Docs:** high-quality docstrings; `@internal` markers present on all Layer 3/4 symbols; internal-only contract verified (absent from the top-level package `__all__`); no internal identifiers in committed code (public repo).
- **Tests:** all four acceptance criteria covered with assertion-rich, shape-asserting tests (lowering invariants, hash stability both directions, context defaults, registry round-trip + thread-safety, internal-only contract); registry global state reset between tests.

**Recommendation:** Merge. The MEDIUM/LOW items are non-blocking — the doc-clarity nits (M1, M2, L2, L3) are worth a quick pass here or in the first compiler PR, and M5's invariants should be validated by that first consumer.

probablyrobot (approved):

PR #1008, opened 2026-06-05
Add Postgres recall-filter compiler for the pgvector backend

filippetrovic:
## Summary

Adds the PostgreSQL backend compiler that lowers the canonical recall-filter
AST into a single-table `khora_chunks` `WHERE` predicate (no `documents` join —
the filterable document columns are denormalized onto the chunk row). The
compiler implements a deterministic field-match contract so the same filter
returns the same row set regardless of stored value shape:

- **Never abort.** Metadata scalar casts are type-gated through `jsonb_typeof`,
  so an array or non-numeric string value can never abort a numeric/date
  comparison on the rest of the scan.
- **Negation polarity.** `$ne`/`$nin` include missing/NULL rows — system columns
  via `col IS NULL OR …`, metadata via a coalesced-totality wrapper — matching
  MongoDB semantics (a negation includes documents lacking the field).
- **Impossible operand/column pairs short-circuit.** An exact-array operand
  against a scalar system column compiles to a constant `false` predicate
  instead of an invalid `varchar = text[]` comparison.
- **Presence/null use JSONB primitives.** `$exists` and null-or-missing use
  `?` / `#> = 'null'::jsonb`, never value extraction.

Metadata equality/set operators use array-aware JSONB containment (`@>` / `?|`);
range operators use the type-gated scalar cast. The `$date` typed literal
parses-or-excludes through a new `khora_try_timestamptz()` SQL helper
(`IMMUTABLE PARALLEL SAFE`, installed by a Postgres-only migration) behind a
cheap regex pre-gate, under a `TimeZone='UTC'` connection.

The compiler is wired into the pgvector backend through an additive `filter_ast`
seam and registered in the internal compiler registry; the previous lexicographic
metadata range comparison (which mis-ordered numeric strings like `"10" < "2"`)
is replaced by the type-gated path. The legacy temporal-filter path is preserved
during its deprecation window.

## Test plan

- [x] Unit tests pass — SQL-emission assertions for every operator and contract rule
- [x] `make lint` / `make format` clean
- [ ] Postgres-backed row-set tests over adversarial rows (array, non-numeric
      string, numeric-string ordering, missing key, NULL column, malformed date)
      — run against a live PostgreSQL

### DIFF
--- a/src/khora/db/migrations/versions/045_khora_try_timestamptz.py
+++ b/src/khora/db/migrations/versions/045_khora_try_timestamptz.py
@@ -1,0 +1,66 @@
+"""Add khora_try_timestamptz() — a safe text→timestamptz cast for $date filters.
+
+Revision ID: 045_khora_try_timestamptz
+Revises: 044_khora_chunks_backfill_denormalized
+Create Date: 2026-06-05
+
+Part of the deterministic recall-filter contract (§4/§7). The Postgres recall-filter compiler needs to
+compare metadata strings against a timestamp for the ``$date`` operator. A bare
+``txt::timestamptz`` cast aborts the whole statement the moment it meets a
+non-parseable string, which is fatal when a JSONB metadata value is malformed or
+of an unexpected shape. This function wraps the cast in an exception handler so a
+bad value yields ``NULL`` (and therefore a non-matching row) instead of erroring
+out the query.
+
+It is ``IMMUTABLE`` (a pure cast of its input — the same text always maps to the
+same timestamptz, given a fixed session ``TimeZone``, which the connection pins
+to UTC) and ``PARALLEL SAFE`` (no side effects, no shared state). ``IMMUTABLE``
+also leaves the door open to backing a functional index on the cast later.
+
+Postgres-only: the function is plpgsql. The sqlite_lance test fixtures run the
+full Alembic chain against SQLite, so we skip silently on non-Postgres dialects
+rather than emit SQL SQLite cannot parse — following migrations 029 and 044.
+"""
+
+from collections.abc import Sequence
+
+from alembic import op
+
+revision: str = "045_khora_try_timestamptz"
+down_revision: str | Sequence[str] | None = "044_khora_chunks_backfill_denormalized"
+branch_labels: str | Sequence[str] | None = None
+depends_on: str | Sequence[str] | None = None
+
+
+_CREATE_FUNCTION_SQL = """
+CREATE OR REPLACE FUNCTION khora_try_timestamptz(txt text)
+RETURNS timestamptz
+LANGUAGE plpgsql
+IMMUTABLE
+PARALLEL SAFE
+AS $$
+BEGIN
+    RETURN txt::timestamptz;
+EXCEPTION WHEN others THEN
+    RETURN NULL;
+END;
+$$;
+"""
+
+_DROP_FUNCTION_SQL = "DROP FUNCTION IF EXISTS khora_try_timestamptz(text);"
+
+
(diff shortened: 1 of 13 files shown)


filippetrovic:

Independent review by an 8-role panel (security, performance, correctness, tests, database, observability, docs, devil's advocate), plus a dedicated security audit. Each reviewer examined the diff in its own domain.

**Overall assessment: ✅ APPROVE** — no CRITICAL, HIGH, or MEDIUM findings.

### Per-domain results

| Domain | Verdict | Notes |
|---|---|---|
| **Correctness** | ✅ clean | All four field-match rules verified line-by-line: type-gated casts (never abort), `$ne`/`$nin` NULL/absent inclusion, narrow exact-array-vs-scalar → constant `false`, presence/null via `?`/`#>` (not value extraction). Empty AST → `true`; metadata NULL-coalesced to `'{}'` for total presence ops. |
| **Security** | ✅ clean | No injection surface: column tokens derive from a closed system-key whitelist + engine context (never user text); the JSONB path builder escapes `\` and `"` correctly; operands are bound; the AST is pre-validated. No secrets, no unsafe logging, no DOS vector. |
| **Database** | ✅ clean | Migration is `IMMUTABLE PARALLEL SAFE`, dialect-gated (Postgres-only; SQLite no-op), correct `down_revision` + `downgrade`. JSONB operators well-formed; `TimeZone='UTC'` is a correct per-connection setting with a documented shared-engine caveat. Migration head-shift consistently updated across all dependent tests. |
| **Performance** | ✅ clean | System-key equality/`IN` stay sargable against the selective indexes; metadata `@>`/`?` are GIN-friendly; the date regex pre-gate keeps the exception path off the common case; compilation is O(\|AST\|) with no quadratic emission. |
| **Tests** | ✅ clean | Unit tests pin emitted SQL for every operator/rule; integration tests assert exact row-sets over adversarial rows (array, non-numeric string, numeric-string ordering, missing key, NULL column, malformed date). Hard imports (no silent skips); session-local temp-table isolation. |
| **Observability** | ✅ clean | Appropriately introduces no new spans/metrics (none needed for a query-plan compiler); does not trip the telemetry-contract drift gate. |
| **Docs** | ✅ clean | Docstrings accurately describe the emission rules; the compiler is correctly kept internal (not re-exported on the top-level public surface). |
| **Devil's Advocate** | ✅ sound | No over-engineering or scope creep; the context-derived column approach and the four-rule decomposition are justified; constant-`false` (vs raising) for the impossible pair is the correct, determinism-preserving choice. |

### Security audit
No dependency-manifest changes (no new dependency-vulnerability surface); no cryptographic, authentication, network, `eval`/`exec`, or subprocess code introduced. The only security-relevant surface is SQL/JSONB injection, which is fully defended (see Security row). **No findings.**

### Informational (LOW / non-blocking)
- **L1 [Tests/CI]** The Postgres-backed row-set tests use the repo-standard `skipif(pg-unreachable)` guard and therefore **skip in CI**, since no CI job currently provisions Postgres for the integration step. This is a **repo-wide** condition (every real-PG integration test behaves this way), not specific to this change — the tests gate locally (`make dev`) and in soak. A dedicated `postgres-integration` CI job (mirroring the existing service-backed integration job) would let them gate in CI; tracked as a follow-up.
- **L2 [Design]** The additive `filter_ast` compilation seam is exercised only by this backend until the facade/cross-engine wiring lands in a later slice — intentional and behind a default-`None` parameter, so it is inert until then.

**Recommendation:** merge. CI is green (all checks passing) and the implementation faithfully meets the deterministic field-match contract.

probablyrobot (approved):

PR #1009, opened 2026-06-06
Add filter= kwarg to recall() with canonical filter AST engine carrier

filippetrovic:
## Summary
- Add a public `filter=` keyword to `recall()` that validates **once** at the facade and lowers a `RecallFilter` (or `dict`) to the canonical filter AST (`parse_to_ast`). Validation failures propagate as the existing structured filter-validation error.
- Carry the canonical `FilterNode` across the engine boundary as an **additive** `filter_ast` parameter on `MemoryEngineProtocol.recall()` and all engine implementations (vectorcypher, chronicle, skeleton). Engines accept-and-ignore it for now — no SQL pushdown is wired in this change (same pattern as the existing `temporal_filter` carrier).
- Surface a per-engine pushdown summary in `RecallResult.engine_info["filter"]` (`{engine, supported, pushed_down}`) on **every** call.
- Replace the legacy temporal block with a **deprecation shim** for `start_time`/`end_time`:
  - Emits a `DeprecationWarning` (once) and folds the bounds into an `occurred_at` filter through the same single validation path.
  - `start_time` → `$gte` (inclusive), `end_time` → `$lt` (exclusive) to match the existing temporal-window semantics (`occurred_at >= start AND occurred_at < end`).
  - Only non-`None` bounds are emitted (no `null` operand stubs).
  - Inverted ranges (`start > end`) fold to an **empty** filter (no error); naive/aware datetimes normalize to UTC.
  - Passing both `filter=` and `start_time`/`end_time` raises `ValueError`.

## Test plan
- [x] Unit tests pass — `tests/unit/test_khora.py` + `tests/recall`: **424 passed**. Rewrote the temporal-bounds tests and added filter-kwarg coverage (validate-once, `FilterNode` reaches the engine, invalid filter raises, `engine_info["filter"]` populated with/without a filter, deprecation/inverted-range/mixed-tz behavior, equal-bounds edge case).
- [x] `ruff format` + `ruff check` clean.
- [ ] Full integration/e2e suite + coverage gate (CI).
- [ ] Manual verification: N/A — library change, covered by unit tests.

### DIFF
--- a/src/khora/engines/chronicle/engine.py
+++ b/src/khora/engines/chronicle/engine.py
@@ -66,6 +66,7 @@ from khora.telemetry.metrics import metric_counter, metric_histogram
 if TYPE_CHECKING:
     from khora.extraction.chunkers import ChunkStrategy
     from khora.extraction.skills import ExpertiseConfig
+    from khora.filter import FilterNode
 
 ChronicleStorageBackend = Literal["pgvector", "lancedb"]
 
@@ -1462,6 +1463,7 @@ class ChronicleEngine:
         min_similarity: float = 0.0,
         temporal_filter: Any | None = None,
         recency_bias: float | None = None,
+        filter_ast: FilterNode | None = None,
     ) -> RecallResult:
         """Recall memories using 4-channel parallel retrieval with RRF fusion.
 
@@ -1482,6 +1484,8 @@ class ChronicleEngine:
             min_similarity: Minimum similarity threshold
             temporal_filter: Reserved for Phase 2 temporal filtering
             recency_bias: Override temporal decay weight (0.0-1.0)
+            filter_ast: Canonical recall-filter AST. Accepted for protocol
+                parity; Chronicle does not push filters down yet (ignored).
 
         Returns:
             RecallResult with fused and decay-scored chunks
--- a/src/khora/engines/protocol.py
+++ b/src/khora/engines/protocol.py
@@ -12,6 +12,7 @@ if TYPE_CHECKING:
     from khora.core.models import Document, Entity, MemoryNamespace
     from khora.extraction.chunkers import ChunkStrategy
     from khora.extraction.skills import ExpertiseConfig
+    from khora.filter import FilterNode
     from khora.khora import BatchResult, RecallResult, RememberResult, Stats
     from khora.query import SearchMode
 
@@ -114,6 +115,7 @@ class MemoryEngineProtocol(Protocol):
         # Temporal parameters (optional — engines may ignore these)
         temporal_filter: Any | None = None,
         recency_bias: float | None = None,
+        filter_ast: FilterNode | None = None,
     ) -> RecallResult:
         """Recall memories relevant to a query.
 
@@ -128,6 +130,8 @@ class MemoryEngineProtocol(Protocol):
                 Engines that do not support temporal filtering may ignore this.
             recency_bias: Optional recency bias weight (0.0–1.0).
                 Engines that do not support recency biasing may ignore this.
+            filter_ast: Optional canonical recall-filter AST. Engines that do
+                not filter may ignore it.
 
         Returns:
             RecallResult with matched memories
--- a/src/khora/engines/skeleton/engine.py
+++ b/src/khora/engines/skeleton/engine.py
@@ -48,6 +48,7 @@ from .backends import (
 if TYPE_CHECKING:
     from khora.extraction.chunkers import ChunkStrategy
     from khora.extraction.skills import ExpertiseConfig
+    from khora.filter import FilterNode
 
 
 class SkeletonConstructionEngine:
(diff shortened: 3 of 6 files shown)


filippetrovic:
## 🤖 Multi-perspective PR review

Reviewed by an agent team across six domains (security, performance, correctness, tests, docs, devil's advocate) plus a dedicated security audit.

**Overall assessment: APPROVE — no merge-blocking findings.** The change is additive (new `filter=` kwarg + additive `filter_ast` engine-protocol param) with a behavior-preserving deprecation shim. Targeted unit suite green (424 passed); `ruff format`/`check` clean.

### Security audit — PASS
- No injection, secret-handling, deserialization, file/path, SSRF, open-redirect, ReDoS, or mass-assignment patterns in the diff. No dependency/manifest changes.
- The only untrusted input (the `filter` dict) is validated by `RecallFilter.model_validate` (`extra="forbid"`, iterative depth bound `_MAX_FILTER_DEPTH = 100` → no stack exhaustion) **before** the AST is built, and the AST is **not** compiled to any backend query on the recall path (engines accept-and-ignore `filter_ast`). No source→sink path.

### MEDIUM
- **M1 · Docs** — `recall()` docstring `Returns:` doesn't mention the newly-populated `engine_info["filter"]` (`{engine, supported, pushed_down}`). Add a one-line note. *(Non-blocking.)*

### LOW
- **L1 · Correctness/Docs** — `engine_info["filter"]` reports `supported`/`pushed_down` as a per-engine capability matrix independent of whether a filter was supplied (correct by design). An optional `applied: bool` field would disambiguate "engine can push down" from "pushed down on this call" for downstream consumers.
- **L2 · Maintainability** — `supported = (engine_name == "skeleton")` is hardcoded (intentional this revision; skeleton is the planned pushdown target). Consider promoting to a backend-capability query once a second engine gains pushdown, so the flag can't silently regress.
- **L3 · Performance** — `engine_info` is rebuilt via `{**existing, "filter": …}` on every call (O(keys), negligible at typical sizes). Fine as-is.

### Considered & adjudicated — intentional, no change needed
The devil's advocate challenged four points; each is **intentional per this change's design contract**:
- **Dual `temporal_filter` + `filter_ast` from deprecated bounds** — required. `filter_ast` is inert on the recall path today (no engine threads it to its backend yet), so the legacy `temporal_filter` must be retained or temporal filtering would silently stop. Documented in a code comment.
- **Inverted range (`start > end`) → empty filter, no error** — deliberate behavior for the deprecation path (supersedes the old `ValueError`).
- **Mixed naive/aware datetimes → UTC-normalize, no error** — deliberate (supersedes the old `ValueError`); inverted-range comparison runs *after* normalization, so no naive-vs-aware crash.
- **`end_time` → `$lt` (exclusive)** — matches the legacy temporal window (`occurred_at < occurred_before`); a boundary-inclusivity slip here was caught and fixed during implementation.

### Notes
- **CHANGELOG** — additive + a deprecation (not a breaking change); per the repo's release-time CHANGELOG convention, fold a note into the next release entry (Added: `filter=`; Deprecated: `start_time`/`end_time` with migration path; Added: `engine_info["filter"]`).
- `engine_info["filter"]` is a free-form `dict` key (not a telemetry span/metric) → no telemetry-contract entry required. The public filter model is already exported.

### Tests — strong
16 tests (8 deprecated-bounds + 8 filter-kwarg). Assertions verify a real `FilterNode` AST reaches the engine (not raw dicts); every acceptance criterion + edge cases (equal/inverted bounds, mixed-tz, warn-once, invalid→validation-error) covered; offline and isolated.

probablyrobot (approved):

filippetrovic:
### Follow-up: inverted time range now folds to an empty-matching filter

Fixed the inverted-range (`start_time > end_time`) behavior in the deprecated-bounds shim.

**Before:** an inverted range was special-cased to *no filter at all* — `temporal_filter=None` and `filter_ast=None` — so the recall ran **unfiltered and returned every row** in the namespace. A footgun.

**After:** the special case is removed. An inverted pair flows through the same fold as any other bounds, yielding `{occurred_at: {"$gte": t1, "$lt": t2}}`. With `t1 > t2` nothing can satisfy `>= t1 AND < t2`, so the predicate is **empty-matching** and the recall returns **zero rows** — parity with the equivalent `filter=` predicate (Mongo semantics). The legacy `temporal_filter` window is built from the same inverted bounds and is likewise empty-matching, so the two stay consistent. One fold path now handles every bound combination; the no-null-stub guard, UTC normalization, mutual-exclusion `ValueError`, and the `DeprecationWarning` are unchanged, and the upper-bound operator stays `$lt`.

**Test:** `test_inverted_range_folds_to_empty` → `test_inverted_range_folds_to_empty_matching_filter`; it now asserts a real `filter_ast` (both `$gte` and `$lt` `occurred_at` clauses with the given `t1`/`t2`), not `None`. This unit assertion is on the synthesized predicate shape — the test class uses a mocked engine, so the **zero-row recall behavior follows from the empty-matching predicate** rather than a DB row-count assertion.

PR #1011, opened 2026-06-07
Add recall-filter telemetry counters and canonical-hash span attribute

filippetrovic:
## Summary
- Add three OpenTelemetry counters for the deterministic recall filter, behind a lazy-init + double-checked-lock helper in a new `filter/telemetry.py`:
  - `khora.recall.filter.unindexed_metadata` — incremented once per metadata leaf that compiles to an unindexed JSONB column access (full-scan risk), split only by the bounded operator enum.
  - `khora.recall.filter.under_filled` and `khora.recall.filter.graph_channel_empty` — defined now, left quiet until a later graph-engine change wires them.
- Emit `unindexed_metadata` from the Postgres filter compiler on the metadata-leaf branch only (system-key predicates do not emit). Compilation runs once per recall, so the hybrid vector + keyword path does not double-count.
- Tag the `khora.recall` span with `filter.canonical_hash` (only when a filter is supplied) so traces correlate with the engine's filter cache key.
- Register all three counters in the telemetry contract (drift gate).
- Cardinality discipline: no `namespace_id` and no metadata key-path on any metric label — both stay span attributes only.

## Test plan
- [x] New unit tests pass (`tests/unit/filter/test_filter_telemetry.py`, 11 tests): counter construction, per-metadata-leaf firing, system-key silence, quiet V2 counters, and the span attribute present/absent
- [x] Telemetry contract drift gate green (`tests/unit/telemetry/test_contract.py`, both forward and reverse checks)
- [ ] CI: full `make test` suite green

### DIFF
--- a/docs/observability.md
+++ b/docs/observability.md
@@ -176,6 +176,7 @@ Highlights:
 - **Public metrics**: `khora.memory.{recall,ingest}.duration`,
   `khora.llm.tokens`, `khora.llm.cost_usd`,
   `khora.neo4j.pool.{acquire_duration,timeout,connections.*,utilization}`,
+  `khora.recall.filter.{unindexed_metadata,under_filled,graph_channel_empty}`,
   `khora.chronicle.abstention_signal`, `khora.log.queue.depth`.
 - **Khora-contributed resource attribute**:
   `khora.telemetry.contract.version` - bumped alongside contract changes
--- a/docs/telemetry-contract.json
+++ b/docs/telemetry-contract.json
@@ -1455,6 +1455,33 @@
       ],
       "description": "Issue #932. A queued document identity could not be re-loaded at dequeue time because the row was deleted/forgotten between enqueue and dequeue. The pending processor carries lean items (doc_id + namespace_id) and re-loads the full Document in the worker; when the loader returns None the item is skipped per ADR-001 (logged INFO, batch registration decremented so BatchHandle.wait can't hang). NO namespace_id label - cardinality rule.",
       "_comment": "Issue #932. NO namespace_id label - cardinality rule."
+    },
+    {
+      "name": "khora.recall.filter.unindexed_metadata",
+      "type": "counter",
+      "stability": "public",
+      "unit": "1",
+      "owner": "filter",
+      "labels": [
+        {"name": "op", "values": ["$eq", "$ne", "$gt", "$gte", "$lt", "$lte", "$in", "$nin", "$exists"], "_comment": "The metadata leaf's comparison operator (Op enum wire literal). Logical operators ($and/$or/$not) are nodes, not leaves, so they never appear here. Bounded by the closed Op enum."}
+      ],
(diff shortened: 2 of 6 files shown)


filippetrovic:

8 reviewers (security, performance, correctness, test, database, observability, docs, devil's advocate) + a dedicated security audit. **Overall: ✅ APPROVE.**

### CRITICAL / HIGH
None.

### MEDIUM
- **[Docs] `docs/observability.md`** — the "Public metrics" highlights list omitted the three new public counters. **✅ Fixed in this PR** (added `khora.recall.filter.{unindexed_metadata,under_filled,graph_channel_empty}`).
- **[Test] `tests/unit/filter/test_filter_telemetry.py`** — operator coverage exercises `$eq`/`$gt`/`$in` of the 9 enum values. *Follow-up (optional):* parametrize across all operators. Low risk — the counter is emitted at the dispatch level (`compile_clause`) before operator-specific handling, and the `op` attribute is a pass-through of `clause.op.value`.
- **[Test]** No explicit test for nested logical operators (`$and`/`$or`/`$not`) over multiple metadata leaves, nor a bare-blob `metadata` `$eq`. *Follow-up (optional):* both route through `compile_clause` per leaf, so behavior is covered transitively; an explicit test would harden against future compiler refactors.

### LOW
- **[Docs]** `src/khora/filter/telemetry.py` lazy-getter inline descriptions are slightly more condensed than the contract entries (module + helper docstrings are thorough). Optional polish.
- **[Test]** `span_exporter` fixture could assert no pre-existing spans at entry; `_mock_config` could guard against future config drift. Optional.

### Devil's Advocate — "no merge" claim refuted
The devil's advocate recommended blocking, arguing the contract reverse-gate (`test_public_metrics_are_emitted`) requires an **emit** (`.add()`) site for the two quiet public counters. This is incorrect: the reverse-check requires a literal `metric_counter("name")` **declaration** site, which all three counters have in their lazy getters — `.add()` is not required. Verified empirically: `test_public_metrics_are_emitted` and `test_no_undeclared_metrics` both **pass** (full contract suite: 14 passed). The DA's substantive points all concur the implementation is sound (per-leaf granularity correct, module isolation reasonable, `filter.canonical_hash` genuinely useful for cache-hit correlation).

### Security Audit — ✅ PASS
No findings across the 16 categories. No untrusted-input→sink paths, no secrets, no new dependencies; `canonical_hash` is SHA-256 used for filter identity/cache-key correlation (not a security primitive). Cardinality discipline upheld: only the bounded `op` enum on the counter, `namespace_id` and metadata key-paths kept off all metric labels.

### Verification
- `tests/unit/filter` (62) + `tests/unit/telemetry/test_contract.py` (14) — green (`--no-cov`).
- `make format` / `make lint` — clean.
- The single MEDIUM caused by this change (doc drift) is fixed; remaining items are optional test-hardening follow-ups.

probablyrobot (approved):

PR #1012, opened 2026-06-07
Thread recall filter_ast from skeleton engine to backend search()

filippetrovic:
## Summary
- The skeleton engine's `recall()` accepted a canonical recall-filter AST (`filter_ast`) but dropped it before calling the temporal store, so a live `recall(filter=...)` validated and built the AST yet returned **unfiltered** results — the filter evaporated at the engine boundary.
- Forward `filter_ast` from `recall()` into `temporal_store.search()`, and declare the kwarg on the `TemporalVectorStore` protocol plus all five backends so the polymorphic call stays type-safe and no backend breaks.
- The pgvector backend already compiles the AST into a SQL `WHERE` predicate — it is now actually reached on the live recall path. The other four backends (`sqlite_lance`, `surrealdb`, `weaviate`, `turbopuffer`) accept the kwarg for protocol parity and ignore it for now (per-backend pushdown is a later slice).
- Correct a stale facade comment that claimed the skeleton engine does not thread the filter through.
- Add a fast, no-Postgres regression test that fails if the engine drops the kwarg or any backend signature loses it.

## Test plan
- [x] `make format` clean; `make lint` (ruff + ty) exits 0
- [x] New unit test `tests/unit/engines/test_skeleton_filter_ast.py` passes — engine-forwarding identity check + all-five-backend signature guard
- [x] `tests/unit/engines/` suite green locally (1240 passed, 6 skipped)
- [ ] Full Postgres integration suite (`make test-integration`) — **gated by CI** (requires Docker; not runnable in the local sandbox). Pre-existing unit failures in the local sandbox are limited to suites needing uninstalled optional extras (`langgraph`, `lancedb`) and are identical on the base branch.
- [ ] Manual verification: a live `recall(filter=...)` on the skeleton+pgvector path returns only matching chunks (the full end-to-end assertion lives in the integration-test slice)

## Follow-up (out of scope here)
- The recall facade still reports `engine_info["filter"]["pushed_down"] = False` unconditionally. Now that the skeleton+pgvector path pushes the filter into SQL, that summary is inaccurate for that path. Making the per-engine pushdown matrix honest (and asserting it) is owned by the integration-test slice and intentionally left untouched here to keep this change surgical.

### DIFF
--- a/src/khora/engines/skeleton/backends/__init__.py
+++ b/src/khora/engines/skeleton/backends/__init__.py
@@ -18,6 +18,7 @@ from khora.core.models.document import Chunk
 if TYPE_CHECKING:
     from khora.config import KhoraConfig
     from khora.core.models.document import Document
+    from khora.filter.ast import FilterNode
 
 
 @dataclass
@@ -162,6 +163,7 @@ class TemporalVectorStore(Protocol):
         temporal_filter: TemporalFilter | None = None,
         hybrid_alpha: float | None = None,  # None = vector only, 0-1 = blend
         query_text: str | None = None,  # Required for hybrid search
+        filter_ast: FilterNode | None = None,
     ) -> list[TemporalSearchResult]:
         """Search for similar chunks with temporal filtering.
 
@@ -173,6 +175,9 @@ class TemporalVectorStore(Protocol):
             temporal_filter: Structured temporal and metadata filters
             hybrid_alpha: Blend factor for hybrid search (0=BM25 only, 1=vector only)
             query_text: Original query text for BM25 (required if hybrid_alpha is set)
+            filter_ast: Canonical recall-filter AST. The pgvector backend
+                compiles it to a WHERE predicate; the other backends accept
+                it for protocol parity and ignore it (no compilation yet).
 
         Returns:
             List of matching chunks with similarity scores
--- a/src/khora/engines/skeleton/backends/sqlite_lance.py
+++ b/src/khora/engines/skeleton/backends/sqlite_lance.py
@@ -49,6 +49,7 @@ from khora.storage.backends.sqlite_lance._helpers import (
 from khora.telemetry import trace_span
 
 if TYPE_CHECKING:
+    from khora.filter.ast import FilterNode
     from khora.storage.backends.sqlite_lance.connection import EmbeddedStorageHandle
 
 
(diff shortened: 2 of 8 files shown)


filippetrovic:
## 🤖 Multi-Perspective Review

Reviewed by a 7-role agent team (security, performance, correctness, tests, docs, database, devil's advocate) plus a dedicated security audit.

**Overall assessment: ✅ APPROVE** — no CRITICAL / HIGH / MEDIUM / LOW findings.

### Scope verified
- `recall()` now forwards `filter_ast` **unconditionally** into `temporal_store.search()` (`engine.py:587`), including the `None` case.
- The `TemporalVectorStore` protocol + all five backends declare `filter_ast: FilterNode | None = None`; pgvector compiles it (unchanged here, now actually reached), the other four accept-and-ignore by design.
- Stale facade comment corrected; docstrings accurate.
- New no-Postgres regression test guards both the engine-drop and backend-signature-drift regressions.

### Findings by severity
**None.** (CRITICAL: 0 · HIGH: 0 · MEDIUM: 0 · LOW: 0)

### Highlights per lens
- **Security** — `filter_ast` stays in the validated-AST domain until the SQL layer; pgvector binds operands via parameterized queries, escapes JSON path segments, and whitelists system keys. No new injection/exposure surface.
- **Performance** — zero overhead on the recall hot path; the compiler imports are lazy, inside the `filter_ast is not None` guard; accept-and-ignore backends run no body for the kwarg.
- **Correctness** — unconditional forwarding; all 5 signatures structurally match the protocol (keyword-passable); no placeholders; `None`-vs-set handled.
- **Database** — `filter_ast` AND-s correctly alongside the legacy `temporal_filter`; boundary semantics (`>=` lower / `<` upper) stay consistent across the deprecated-bounds and new-AST paths, so no boundary rows are dropped; single-table denormalized compile (no joins).
- **Tests** — 8/8 pass; identity-based forwarding assertion + parametrized signature-drift guard across protocol + 5 backends; correctly defers the Postgres end-to-end assertion to a later slice.
- **Docs** — all comments/docstrings accurate and consistent; no remaining stale "filter dropped at engine" claims.
- **Devil's Advocate** — approach is right (a polymorphic kwarg requires a protocol-wide param); the turbopuffer backend is genuinely reachable, so adding its param is required, not scope creep; accept-and-ignore is documented, not silent.

### Security Audit
**PASS** — the diff adds only type-signature params, docstrings/comments, and a mock-based unit test. No new trust boundary, dangerous sink, secret handling, or input parsing.

### Noted (intentional, out of scope)
- The recall facade still reports `engine_info["filter"]["pushed_down"] = False` unconditionally — now inaccurate for the skeleton+pgvector path. Deliberately deferred (changing it would alter returned recall semantics); flagged for the follow-up integration slice that owns the pushdown matrix.

---
*Consolidated from an automated multi-agent review; advisory.*

probablyrobot (approved):

PR #1015, opened 2026-06-08
Add Neo4j Cypher recall-filter compiler with system-key pushdown

filippetrovic:
## Summary
- Add `compile_cypher(ast, ctx)` — the Layer-4 backend compiler that lowers a canonical recall-filter AST to a Cypher `WHERE` fragment against a recall `Chunk` node, mirroring the existing Postgres compiler in structure and contract.
- **System-key pushdown:** the date keys and the eight projected document keys compile to first-class node-property predicates. `datetime` operands bind as ISO-8601 strings to match the chunk node's stored representation, so comparisons are lexicographic (ISO-8601 string order == chronological order for the UTC-normalized values the validator produces).
- **Mongo-faithful negation guard:** `$ne` / `$nin` emit `(prop IS NULL OR prop <> $v)` / `(prop IS NULL OR NOT prop IN $vs)` so a node missing the property is **included** (a naive `<>` evaluates to null and silently excludes it). `$eq` / range / `$in` are built total via `coalesce(..., false)` so a wrapping `NOT` flips absent rows correctly. `$exists` and null matches use `IS [NOT] NULL` (an absent node property is null).
- **Metadata routing:** metadata predicates are not pushed down (the graph backend stores metadata as a serialized JSON string). They honor the `on_unsupported` policy — `raise`, or `split` which emits a non-constraining placeholder and omits the leaf from `consumed_keys` so the engine can over-fetch and post-filter. `consumed_keys` reports exactly the pushed-down system keys.
- Add a focused unit-test module covering every operator, null-guard polarity, totality under negation, ISO-string date binding, exact-array vs membership handling, empty-list guards, `consumed_keys` reporting, context customization (node variable / field mapping / param namespace), and the raise/split unsupported paths.
- Scope is surgical: one new compiler module + its tests. No engine, registry, or protocol wiring (that is handled separately).

## Test plan
- [x] Unit tests pass (`uv run pytest tests/unit/filter/ tests/recall/ -q` → all green; 39 new tests for the compiler)
- [x] Lint/format/typecheck clean (`ruff check`, `ruff format --check`, `ty check`)
- [ ] Integration tests pass (CI, with the full backend environment)
- [x] Manual verification: empirically compiled representative ASTs (every operator, AND/OR/NOT composition, metadata raise/split) and confirmed the emitted Cypher + bind params

### DIFF
--- a/src/khora/filter/compilers/cypher.py
+++ b/src/khora/filter/compilers/cypher.py
@@ -1,0 +1,281 @@
+"""Neo4j Cypher recall-filter compiler — ``@internal``.
+
+Lowers a canonical :class:`~khora.filter.ast.FilterNode` to a single Cypher
+``WHERE`` fragment against a recall ``Chunk`` node (no relationship traversal —
+the filterable document keys are projected onto the chunk node). The output is a
+:class:`~khora.filter.registry.CompiledFilter` whose ``predicate`` is a Cypher
+boolean *string* the engine splices into its own ``WHERE``, paired with a
+``params`` dict of ``$name`` binds the engine passes through to the driver.
+
+The compiler is the Layer-4 half of the §4/§7 filter contract. It speaks
+the four emission rules:
+
+1. **Never abort.** Cypher comparisons against a property of the wrong type or
+   an absent property evaluate to ``null`` (not an error), so a numeric compare
+   on a string-valued property never blows up — it yields ``null`` (then
+   ``false`` via ``coalesce``) instead of erroring.
+2. **Polarity.** Negations include absent/null rows: a system ``$ne`` is
+   ``(var.key IS NULL OR var.key <> $v)``; a ``$nin`` is ``(var.key IS NULL OR
+   NOT var.key IN $vs)``. Never drop a NULL row on a negation.
+3. **Impossible pairs (narrow).** ONLY an ``$eq`` exact-array (tuple) operand
+   against a scalar node property is compiled as a normal ``=`` against a list
+   bind — a scalar property never equals a list, so Cypher yields ``false``; no
+   special-cased constant is emitted. ``$in`` / ``$nin`` are normal membership.
+4. **Presence/null.** ``$exists`` resolves to ``var.key IS NOT NULL`` /
+   ``var.key IS NULL`` and a ``{k: null}`` match resolves to ``var.key IS NULL``
+   (Neo4j has no stored-null property — an absent property *is* null).
+
+**Totality (the rule that makes negation uniform).** ``NOT`` is compiled as
+``(NOT (<child>))``; Cypher ``NOT null`` is ``null`` (drops the row), which would
+violate Rule 2. So every leaf that *can* produce ``null`` (an absent property in
+an ``$eq`` / range / ``$in`` compare) is wrapped in ``coalesce(<expr>, false)``
+— a total boolean. The positive use is unchanged (``false`` excludes exactly as
+``null`` would), and a wrapping ``NOT`` then flips absent rows to ``true``
+correctly. ``IS NULL`` / ``IS NOT NULL`` are already total.
+
+**Dates compare as ISO strings.** A chunk stores every datetime property as an
+ISO-8601 string (``.isoformat()``), so a :class:`~datetime.datetime` /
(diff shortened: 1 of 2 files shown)


filippetrovic:

Seven independent reviewers (correctness, security, database/Cypher, performance, test-quality, docs, devil's advocate) plus a dedicated security audit reviewed this PR. Reviewers fetched the diff directly; the security audit scanned only added/modified lines.

**Overall assessment: APPROVE.** No CRITICAL/HIGH/MEDIUM defects. One devil's-advocate finding was evaluated in depth and determined to be the intended layered design (not a defect); it carries one informational note for the downstream engine-integration work.

### Findings by severity

_No CRITICAL, HIGH, or MEDIUM findings._

#### LOW
- **L1 — [Devil's Advocate / Correctness] split-mode placeholder in an `$or` — evaluated, by-design (not a defect).** In `on_unsupported="split"`, an unconsumed metadata leaf inside an `$or` makes the graph predicate `(coalesce(c.<key>=…, false) OR true)` ≡ `true`, so the graph channel returns a **superset** and `consumed_keys` omits the metadata leaf. This is the contract: the engine engages over-fetch precisely because `consumed_keys < all leaves`, then post-filters every candidate through `compile_python` on the **full** AST — which re-applies the metadata constraint, so the final row-set is correct. Superset-then-post-filter is the safe direction; emitting `false` instead would turn `(A OR false)` into `A` and **drop** rows matching only the metadata arm (a real subset/data-loss bug). This is byte-for-byte the same pattern as the already-merged, tested `compile_postgres._unsupported` (returns `sa.true()` in split mode), and split mode is not wired into any engine in this PR (diff is the two new files only) — so there is zero production risk here. **No code change.**

#### INFO
- **I1 — [Devil's Advocate]** Carry-forward for the engine-integration work (separate change): the engine's post-filter must run `compile_python` on the **full** filter AST (not only the unconsumed leaves), and over-fetch must engage whenever `consumed_keys` is a strict subset of the filter's leaves. End-to-end split-mode correctness should be covered there.
- **I2 — [Security audit / Security review]** No injection surface. The compiler emits a Cypher string but never executes it; all operand values are bound out-of-band as `$name` parameters, and only controlled tokens (the node variable from `ctx.table_alias`, property names from the closed `SYSTEM_KEYS` whitelist / `field_mapping`, fixed operators) are interpolated. No secrets, crypto, deserialization, SSRF, file/network/redirect sinks, or new dependencies. Security audit verdict: **PASS**.

### Per-domain verdicts
- **Correctness — CLEAN.** All four emission rules verified; `$ne`/`$nin` IS-NULL guard includes absent rows; totality via `coalesce(…, false)` makes `NOT` flip absent rows correctly; ISO-8601 lexicographic date ordering holds for UTC-normalized operands; exact-array, empty-`$in`/`$nin`, and composition all match `compile_postgres` row-set semantics.
- **Security — CLEAN** (+ audit **PASS**). Parameterized binding throughout; no injection.
- **Database / Cypher — CLEAN.** Emitted Cypher is valid and stable on the targeted Neo4j; `IN []`, `coalesce`, `IS [NOT] NULL`, `<>`-on-absent, and `$`-param/list binds all correct; ISO-string storage assumption cross-checked against the chunk-projection code.
- **Performance — CLEAN.** Single-pass walk, `.join`-based composition (no quadratic concat), linear bind growth; predicate forms are index-friendly.
- **Test quality — CLEAN.** 39 tests covering every operator, null-guard polarity, totality under negation, ISO-string date binding, exact-array vs membership, empty-list guards, `consumed_keys` reporting, raise/split unsupported paths, and context customization; assertions match the actual emitted strings (not literal `false`/`true` for exact-array). Hard import, `pytest.mark.unit`. Full `tests/unit/filter/` + `tests/recall/` suite green.
- **Docs — CLEAN.** Module/function docstrings accurately document the four rules, totality, ISO-string dates, and the metadata-not-pushed-down rationale, consistent with the Postgres compiler; **no internal identifiers** (ticket IDs, ADR numbers, private URLs/names) anywhere in the diff.

### Scope
Surgical — only `src/khora/filter/compilers/cypher.py` and `tests/unit/filter/test_compile_cypher.py`. No engine, registry, or protocol edits.

CI: all required checks green.

probablyrobot (approved):

PR #1016, opened 2026-06-08
Split CI test job into unit and integration jobs with Postgres provisioning

filippetrovic:
## Summary
- Split the single `test` CI job into two parallel jobs:
  - **`test-unit`** — unit + recall suites, no service containers, fast PR feedback. It owns the per-path coverage-floor check, since those paths (the embedded SQLite/LanceDB stack, chronicle engine, acceleration module) are exercised by the unit/embedded suites rather than the database-backed tests.
  - **`test-integration`** — provisions a `pgvector/pgvector:pg17` Postgres service side-car so the Postgres-backed integration tests **actually execute in CI** instead of self-skipping. It waits for the database to be ready, applies migrations (`alembic upgrade head`), then runs the integration suite **serially** (several tests reset the shared schema, so they cannot run in parallel against one database).
- Add a session-level guard (`tests/integration/conftest.py`) that **fails the run loudly** when the integration job expects a database but it is unreachable — a missing database can no longer pass green by silently skipping its tests. The guard is a no-op for local runs that do not opt in, so `make dev`-less local workflows are unaffected.
- Per-path coverage floors were re-measured against the unit-only run and all hold with headroom, so they move to the unit job unchanged. The aggregate coverage target stays enforced via the combined coverage status across both jobs' flags.
- All other CI jobs (lint, security, install, adapter unit tests, examples smoke, and the existing vector-database integration job) are unchanged.

Note: graph-database-backed integration tests remain gated and are deferred to a follow-up that builds on this new integration job.

## Test plan
- [x] `test-unit` suite passes locally (7071 passed)
- [x] Integration suite passes locally for every test that can run without a database (all database-gated tests skip cleanly); zero failures
- [x] Lint, format, and type checks pass (`make lint`, `make format`)
- [x] Workflow YAML validates; the loud-fail guard aborts the run when a database is required but unreachable, and is a no-op otherwise
- [ ] Database-backed integration tests run green in CI (first real execution of these tests in CI — any failures triaged as a follow-up)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -232,9 +232,15 @@ jobs:
         run: uv run python tools/check_optional_imports.py
 
   # ============================================================================
-  # Test — pytest with coverage (parallel with lint and security)
+  # Test (unit) — pytest unit + recall suites with coverage, no services.
   # ============================================================================
-  test:
+  # Hermetic: no database side-car. The recall-filter Postgres smoke test in
+  # tests/recall/ self-skips when Postgres is unreachable, so it is collected
+  # here without provisioning a server. This job also owns the per-path
+  # coverage floor check: all floor paths (sqlite_lance/*, chronicle/engine.py,
+  # _fts5.py, _accel.py) are exercised by the unit + embedded suites, not by
+  # the Postgres integration tests, so the floors must be measured here.
+  test-unit:
     needs: install
     runs-on: ubuntu-latest
     timeout-minutes: 25
@@ -272,6 +278,12 @@ jobs:
         run: |
           uv run pytest tests/unit/ tests/recall/ --cov=src/khora --cov-branch --cov-report=xml:coverage-unit.xml --cov-fail-under=0 -n auto --timeout=120 --timeout-method=thread
 
+      - name: Generate JSON coverage for floor check
+        run: uv run coverage json -o coverage.json
+
+      - name: Check per-path coverage floors
+        run: uv run python scripts/check_coverage_floors.py
+
       - name: Upload unit coverage to Codecov
         uses: codecov/codecov-action@v5
         if: always()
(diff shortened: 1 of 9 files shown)


filippetrovic:
## 🔎 Multi-perspective review — PR #1016

7 independent reviewers (security, performance, correctness, test, database, docs, devil's advocate) + a dedicated security audit. **Overall: APPROVE (design).** The CI split, Postgres provisioning, loud-fail guard, and coverage handling are sound. No CRITICAL/HIGH findings.

### CRITICAL / HIGH
_None._

### MEDIUM
- **[Coverage — Devil's Advocate]** `.github/workflows/ci.yml` — `--cov-fail-under=0` on `test-integration` + Codecov-combined is a *potential* integration-coverage-regression loophole: a PR could drop Postgres-path coverage while the combined number stays within Codecov's threshold. **Mitigated** by per-flag Codecov reporting (the `integration` flag is reported separately) and the rationale is documented inline. Not a blocker — recommend a follow-up to either set a modest integration floor or document the "integration coverage is reviewer-checked, not hard-gated" policy. (The aggregate 77% floor + the per-path floors in `test-unit` remain real gates.)

### LOW
- **[Clarity — Devil's Advocate]** `tests/integration/conftest.py` — add a one-line comment explaining the `.replace("+asyncpg", "")` strip (it normalises the DSN scheme to extract host:port for the socket probe).
- **[Structure — Devil's Advocate]** Collecting `tests/unit/dream/test_locks.py` from the integration job is awkward cross-tree coupling; a future refactor could move the PG-lock tests under `tests/integration/`. Well-documented inline; not a defect.
- **[Docs]** Optional follow-up: add a CLAUDE.md note that CI now runs `test-unit` (hermetic) + `test-integration` (Postgres side-car) in parallel.

### Security Audit — **PASS**
- No injection / SSRF / deserialization surface. The conftest socket-probe parses a job-controlled `KHORA_DATABASE_URL` only.
- `POSTGRES_PASSWORD: khora` is a throwaway, ephemeral CI-service credential (not a real secret).
- Actions (`checkout@v6`, `setup-uv@v7`, `cache/restore@v4`) reuse the repo's existing `@vN` pin convention; the pgvector image is tag-pinned like `compose.yaml`. No dependency/lockfile changes.

### Per-domain verdicts
| Domain | Verdict |
|---|---|
| Security | ✅ clean |
| Performance | ✅ clean (split keeps unit feedback fast; `-n 0` serial runs concurrently with unit) |
| Correctness | ✅ clean (DSN/port aligned, marker filter correct, `--cov-fail-under=0` overrides addopts) |
| Test | ✅ approved (3-layer anti-silent-skip; test_locks correctly collected) |
| Database | ✅ approved (pgvector image, advisory-locked migration step, serial justified by DROP SCHEMA) |
| Docs | ✅ clean (comments accurate; no internal IDs leaked) |
| Devil's Advocate | ✅ approve with the coverage caveat above |

> **Note:** the `test-integration` CI job is now executing the ~120 Postgres-gated tests for the first time. Initial run surfaced real test-level failures (e.g. an asyncpg `TIMESTAMPTZ`-binding strictness issue + migration-test schema-state interactions) — these are being triaged to green (fix-the-fixable / tracked-xfail) as part of this PR, per the ticket's triage scope.

probablyrobot (approved):

PR #1022, opened 2026-06-08
Report honest filter pushdown status in recall engine_info

filippetrovic:
## Summary
- `RecallResult.engine_info["filter"]["pushed_down"]` was hardcoded `False` on the live recall path, so a caller reading that contract field saw `False` even when the skeleton-pgvector backend had compiled the recall filter into a SQL `WHERE` predicate and fully pushed it down.
- The skeleton engine now reports `pushed_down` honestly: a constraint-bearing filter on the pgvector backend reports `True`; a no-filter recall, a constraint-free (empty) filter, and the accept-and-ignore non-pgvector backends report `False`. The value is derived from the pgvector compiler's all-or-nothing (`on_unsupported="raise"`) contract — a recall that returns with a compiled filter consumed every leaf — so no recompile happens on the hot path.
- The facade now surfaces the engine-reported value instead of hardcoding `False`, defaulting to `False` when an engine reports nothing, and the stale "no engine pushes filters down in this revision" comment is removed.

## Test plan
- [x] Unit tests pass (`make test-unit`): facade contract tests cover the `True` case, the engine-silent / non-pgvector default-`False` cases, and the bare empty-filter `False` case
- [ ] Integration tests pass: live skeleton-pgvector `True`/`False` pushdown assertions run in CI's Postgres matrix (self-skip locally without Postgres)
- [x] Manual verification: `engine_info["filter"]["pushed_down"]` is `True` for a constraint-bearing skeleton-pgvector filter and `False` for no-filter / empty-filter / non-pgvector backends

### DIFF
--- a/src/khora/engines/skeleton/engine.py
+++ b/src/khora/engines/skeleton/engine.py
@@ -642,6 +642,34 @@ class SkeletonConstructionEngine:
                 )
             )
 
+        engine_info: dict[str, Any] = {
+            "engine": "skeleton",
+            "backend": self._backend_type,
+            "hybrid_alpha": hybrid_alpha,
+            "temporal_filter": str(temporal_filter) if temporal_filter else None,
+        }
+        # Report honest filter pushdown. The pgvector compiler runs in
+        # on_unsupported="raise" mode, which is all-or-nothing: an unsupported
+        # leaf raises RecallFilterUnsupportedError before recall returns, so a
+        # recall that returns WITH a filter_ast on pgvector means every leaf was
+        # consumed (consumed_keys == all leaves) = fully pushed down. We derive
+        # the flag from that contract without recompiling. When split-mode
+        # (partial pushdown) lands, read CompiledFilter.consumed_keys and
+        # compare against the AST leaves instead.
+        #
+        # A constraint-free filter (the empty-AND root that ``filter={}`` /
+        # ``RecallFilter()`` normalizes to) carries zero leaves, so "all leaves
+        # consumed" is vacuous - nothing was pushed down. We report False for it,
+        # matching the no-filter case rather than claiming a pushdown that did
+        # not narrow anything. ``filter_ast.children`` is the "has constraints"
+        # check (the root is always an AND ``FilterNode``).
+        #
+        # The carrier is written on every recall (the no-filter case lands here
+        # too, with ``filter_ast is None`` -> False); only the boolean varies.
+        engine_info["filter"] = {
+            "pushed_down": filter_ast is not None and bool(filter_ast.children) and self._backend_type == "pgvector"
+        }
+
         return RecallResult(
             query=query,
             namespace_id=namespace_id,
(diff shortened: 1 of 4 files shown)


probablyrobot (approved):

filippetrovic:
## Multi-perspective review

**Overall assessment: APPROVE** — no CRITICAL/HIGH/MEDIUM findings across six review dimensions plus a dedicated security audit. All required CI checks are green.

**Scope reviewed:** honest `engine_info["filter"]["pushed_down"]` derivation for the skeleton-pgvector recall path (engine + facade) and the accompanying unit/integration tests.

### Findings by severity

**CRITICAL / HIGH / MEDIUM:** none.

**LOW**
- **L1 — [Devil's Advocate / API contract]** The facade defaults `pushed_down` to `False` when an engine omits the key. This is fail-safe today (a downstream consumer never misreads a non-pushdown as a pushdown), but a *future* engine that does push down yet forgets to report would silently surface `False`. Recommendation: leave as-is for this change; revisit when a second pushdown-capable engine lands — an integration test would catch the regression. No action required now.

**INFO**
- **Performance:** the derivation is O(1) (reference check + tuple-length check + string compare); no filter recompile, no DB query, no allocation on the recall hot path. The facade read is two nested `dict.get()` calls.
- **Correctness:** all four cases verified — constraint-bearing filter on pgvector → `True`; no-filter → `False`; constraint-free `{}` (empty-AND, zero children) → `False`; non-pgvector skeleton backend → `False`. The `on_unsupported="raise"` all-or-nothing contract is sound (an unsupported leaf raises before the recall returns, so a successful return implies every leaf was consumed), and the root AST is always `AND`-wrapped, so `bool(filter_ast.children)` is the exact "has constraints" check.
- **Tests:** strong `is True` / `is False` assertions throughout (not mere key-presence); the `True` case runs on live pgvector and the `False` cases (no-filter / empty / non-pgvector / engine-silent) are covered at both unit and integration levels; the obsolete "deliberately does NOT pin" interim note was removed.
- **Docs:** the engine and facade comments accurately describe the raise-mode/empty-AND derivation; the stale "no engine pushes filters down in this revision" comment was removed; the diff contains no internal identifiers; `docs/telemetry-contract.json` is unaffected (`engine_info` is a result field, not a span/metric).

### Security Audit
**Verdict: PASS.** Scanned the 227 added lines across all 16 categories (injection, authn/authz, crypto & secrets, access control, SSRF, deserialization, file/path handling, open redirects, mass assignment, ReDoS, supply chain, boundary validation, etc.). The change is a boolean derivation plus a defensive dict read — no untrusted-input-to-sink paths, no new endpoints, no secret/crypto handling, and no dependency or manifest changes. No findings.

### CI
All 9 required checks green: `test`, `security`, `lint`, `weaviate-integration`, `examples-smoke`, `test-google-adk`, `install`, `secret-typing`, `no-agents-md`. The live skeleton-pgvector pushdown assertions pass on the Postgres matrix. (An earlier run hit a flaky xdist worker crash in an unrelated sync-bridge test; it passed on re-run.)

**Recommendation: APPROVE — ready to merge.**

PR #1024, opened 2026-06-08
Project denormalized document keys onto Neo4j chunk nodes

filippetrovic:
## Summary
- Project the eight denormalized document keys (`source_type`, `source_name`, `source_url`, `source_timestamp`, `external_id`, `content_type`, `source`, `title`) onto the Neo4j `Chunk` node at ingest — on both the single and batch create paths — so the Cypher recall-filter compiler can push them down as first-class node properties. `source_timestamp` is stored as an ISO-8601 string to match the compiler's lexicographic date comparison.
- Add five Neo4j filter indexes on the `Chunk` node covering `source_type`, `source_name`, `source_timestamp`, `external_id`, and `content_type`.
- Make the non-atomic create-path dual-write observable: when the Neo4j `Chunk`-node mirror write fails *after* the chunks are already committed to pgvector, record a failure-observability degradation, increment a new `khora.vectorcypher.chunk_mirror.degraded_total{channel,reason}` counter, and continue instead of aborting ingest. The chunk stays queryable on the pgvector/BM25 channel and the graph-side divergence is surfaced on `RememberResult.metadata['degradations']`.
- Update the telemetry contract with the new public counter.

## Test plan
- [x] Unit tests pass (`make test-unit`): denorm-key projection on both create paths (incl. `source_timestamp` ISO serialization), the five new indexes, the create-path degradation path (Neo4j write failure does not abort ingest; degradation recorded + counter emitted), Cypher pushdown of all eight keys with no `UnsupportedFilterError`, and telemetry-contract drift.
- [ ] Integration tests pass (CI — require Postgres/Neo4j via Docker)
- [x] Lint/format/typecheck clean (`make lint`, ruff-format, ty); pre-commit hooks green

### DIFF
--- a/docs/telemetry-contract.json
+++ b/docs/telemetry-contract.json
@@ -1046,6 +1046,18 @@
       "description": "Issue #904 (ADR-001). VectorCypher relationship-fetch silent fallback - incremented when the parallel rels_task raises and the retrieve path continues without relationships. The same event is also appended to RecallResult.engine_info['degradations']. Asymmetric prior art: _cypher_expand failure already sets engine_info['graph_fallback'] / 'graph_error'; this counter closes the rel-fetch gap. NO namespace_id label - cardinality rule. See docs/architecture/failure-observability-contract.md."
     },
     {
+      "name": "khora.vectorcypher.chunk_mirror.degraded_total",
+      "type": "counter",
+      "stability": "public",
+      "unit": "1",
+      "owner": "vectorcypher",
+      "labels": [
+        {"name": "channel", "values": ["graph"], "_comment": "Which mirror channel degraded; graph = Neo4j Chunk-node mirror."},
+        {"name": "reason", "values": ["neo4j_write_failed"], "_comment": "Low-cardinality enum; extend when adding new chunk-mirror fallback paths."}
+      ],
+      "description": "ADR-001. VectorCypher Neo4j chunk-mirror silent fallback - incremented when the create-path Chunk-node write to Neo4j raises and ingest continues with chunks persisted only in pgvector. The same event is also appended to RememberResult.metadata['degradations']. NO namespace_id label - cardinality rule. See docs/architecture/failure-observability-contract.md."
+    },
+    {
       "name": "khora.forget.cascade.degraded_total",
       "type": "counter",
       "stability": "public",
--- a/docs/telemetry-contract.json
+++ b/docs/telemetry-contract.json
@@ -1055,7 +1055,7 @@
         {"name": "channel", "values": ["graph"], "_comment": "Which mirror channel degraded; graph = Neo4j Chunk-node mirror."},
         {"name": "reason", "values": ["neo4j_write_failed"], "_comment": "Low-cardinality enum; extend when adding new chunk-mirror fallback paths."}
       ],
-      "description": "ADR-001. VectorCypher Neo4j chunk-mirror silent fallback - incremented when the create-path Chunk-node write to Neo4j raises and ingest continues with chunks persisted only in pgvector. The same event is also appended to RememberResult.metadata['degradations']. NO namespace_id label - cardinality rule. See docs/architecture/failure-observability-contract.md."
+      "description": "ADR-001. VectorCypher Neo4j chunk-mirror silent fallback - incremented when a Chunk-node mirror write to Neo4j raises (create, replace, or batch ingest path) and ingest continues with chunks persisted only in pgvector. The same event is also appended to RememberResult.metadata['degradations'] where a diagnostics sink exists. NO namespace_id label - cardinality rule. See docs/architecture/failure-observability-contract.md."
     },
     {
       "name": "khora.forget.cascade.degraded_total",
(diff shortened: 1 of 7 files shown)


filippetrovic:

Reviewed by an 8-reviewer panel — **security, performance, correctness, tests, database, observability, docs, devil's advocate** — plus a dedicated security audit. **Overall assessment: APPROVE.** CI is green; the two actionable findings were fixed in the latest push; the remainder are by-design trade-offs or out-of-scope follow-ups.

### ✅ Clean dimensions
- **Correctness** — all 8 denormalized keys land on both create paths; `source_timestamp` is ISO-serialized and None-guarded; the create-path `try/except` wraps only the Neo4j write, records a `Degradation`, emits the counter, and continues without re-raising; no orphaned-edge bug (entity→chunk `MERGE` no-ops when the chunk node is absent — no dangling edge).
- **Database (Neo4j schema)** — the 5 `CREATE INDEX … IF NOT EXISTS` are idempotent and consistently named; node property names match the system keys exactly (identity field-mapping); the ISO-string `source_timestamp` matches the compiler's lexicographic date binding; no migration needed (declarative `ensure_indexes`).
- **Observability** — the counter `khora.vectorcypher.chunk_mirror.degraded_total` carries bounded `channel`/`reason` labels and **no** `namespace_id` (cardinality rule); WARNING + `exc_info=True`; the `Degradation` surfaces on `RememberResult.metadata`; telemetry-contract drift gate passes.
- **Security** — Cypher values are fully parameterized (`$param` for the single path, `UNWIND` row dicts for the batch) → no injection surface; no secrets/crypto; no new dependencies, endpoints, or trust boundaries. Security-audit verdict: **PASS**.

### Fixed in the latest push
- **[HIGH · Tests]** The degradation test asserted the `Degradation` record but not the counter increment. Added an assertion that the counter is incremented once with the bounded `{channel, reason}` labels — closing the observable-contract loop.
- **[Hygiene]** Scrubbed stray internal ticket references from 5 test comments/docstrings (public-repo rule).

### MEDIUM (non-blocking)
- **[Perf]** Only 5 of the 8 projected keys are indexed (`source`, `source_url`, `title` are intentionally not). This mirrors the selective PG-side indexes; filters on the 3 unindexed keys fall back to a scan rather than failing. Worth a follow-up if those become hot filter keys.
- **[Architecture]** The sibling entity→chunk dual-write (`link_entities_to_chunks_batch`, pre-existing and **not** part of this diff) has no equivalent degradation handling. This change deliberately scopes observability to the chunk mirror; entity-link symmetry is a reasonable follow-up.
- **[Perf]** `source_timestamp` is a higher-cardinality datetime index, and the batch `UNWIND` payload grows by 8 props/chunk — both negligible at expected batch sizes; monitor the new counter and index size under load.

### LOW
- **[Docs]** Consider adding a reference-implementation subsection for this mirror-and-fallback pattern to the failure-observability contract doc (the telemetry entry already cross-references it).
- **[Style]** A one-line comment at the `source_timestamp` write site noting the ISO-string serialization rationale (lexicographic Cypher comparison) would aid future readers.
- **[Tests]** The capturing-driver test helper is elegant but coupled to `execute_write`'s closure signature; a short docstring would help maintenance.

**No CRITICAL/HIGH blockers remain after the fixes in the latest push. CI is green across all required checks.**

<sub>Generated by an automated multi-perspective review team + security audit.</sub>

filippetrovic:
### Degrade-and-continue covers only one of three `create_chunk_nodes_batch` call sites

This PR makes the Neo4j chunk-mirror write non-fatal in `_process_document`: on failure it logs a warning, records a degradation, increments `khora.vectorcypher.chunk_mirror.degraded_total{channel,reason}`, and continues with the chunks already durable in pgvector (the guarded call is `engine.py:1122`).

But `create_chunk_nodes_batch` has two other call sites with the **same** "pgvector first, then mirror to Neo4j" shape that don't get the same handling:

- **`engine.py:1773`** (document replace) — `await temporal_store.create_chunks_batch(...)` runs first, then `await dual_nodes.create_chunk_nodes_batch(...)`. A Neo4j failure here is caught by the surrounding self-heal handler, which marks the document FAILED and re-raises, even though the new chunks are already persisted in pgvector.
- **`engine.py:2835`** (windowed batch ingest) — `await temporal_store.create_chunks_batch(...)`, then `await dual_nodes.create_chunk_nodes_batch(...)`. A Neo4j failure aborts the whole batch after pgvector has committed.

So the same transient Neo4j failure produces three different behaviors depending on the ingest entrypoint, and the new `chunk_mirror.degraded_total` counter fires from only one of them — mirror failures on the replace and batch paths stay invisible to the metric, which undercuts its purpose.

**Suggestion:** pull the try/except + counter + degradation block into a small helper (e.g. `_mirror_chunks_or_degrade(dual_nodes, chunks, namespace_id, out_diagnostics)`) and call it at all three sites, so the fallback behavior and the observability signal are uniform. The replace path needs one extra consideration: a mirror-only failure probably shouldn't mark the document FAILED, since pgvector is consistent.

filippetrovic:
Implemented in `d6204ec`. Extracted the fallback into a shared `_mirror_chunks_or_degrade(dual_nodes, chunks, namespace_id, out_diagnostics=None)` helper and routed all three `create_chunk_nodes_batch` call sites through it (create / replace / batch), so the fallback behavior and the metric are uniform:

- **Counter fires from every path** on a mirror failure — closes the gap where the replace and batch paths were invisible to `chunk_mirror.degraded_total`.
- **Replace path** no longer marks the document FAILED on a mirror-only failure (pgvector is already consistent). It degrades-and-continues and surfaces the `Degradation` on `RememberResult.metadata`, consistent with the existing post-PG-commit graph-mirror handling downstream.
- **Batch path** degrades-and-continues instead of aborting the whole batch after the pgvector commit (the counter fires; `BatchResult` has no per-document metadata sink, so no `Degradation` record there — counter only).

Added unit tests for the helper (no-op when there's no graph backend, success records nothing, failure increments the counter + records a `Degradation` without raising, and failure-without-a-diagnostics-sink still increments the counter). Telemetry-contract description updated to cover all three paths.

probablyrobot (approved):

Issue #1020, opened 2026-06-08
Triage Postgres-integration test failures surfaced by the new test-integration CI job


filippetrovic:
The new `test-integration` CI job (which provisions Postgres and runs the previously-self-skipping PG-backed tests for the first time) surfaced two clusters of pre-existing, never-run-in-CI test failures. They are quarantined as tracked `xfail(strict=False)` so the suite is green while they are fixed — they are **not** silently skipped.

### Cluster 1 — migration tests assume a non-shared DB
`tests/integration/db/test_migration_032_dream_runs.py`, `test_migration_034_chronicle_events_bitemporal.py`, `test_migration_041_chunks_denormalized_columns.py`, `test_migration_042_widen_chunks_source_external_id.py`, `test_migration_044_chunks_backfill.py`

These tests reset the alembic version pointer (e.g. to `031`) and drop a single table, but the CI job runs `alembic upgrade head` once up front against the shared service database. The `_reset()` fixtures don't drop the sibling tables a migration creates (e.g. `khora_dream_communities`), so re-applying the migration raises `DuplicateTableError: relation "…" already exists`. Fix: make each migration test's setup CI-shared-DB-safe (e.g. `DROP SCHEMA public CASCADE` + re-migrate to the pre-revision, mirroring the matrix `*_pg` tests) so it is self-contained regardless of the up-front migration.

### Cluster 2 — stale backend API in a storage isolation test
`tests/integration/storage/test_chunk_namespace_isolation_pg.py`

The test calls `PgVectorBackend.create_namespace(...)`, but `PgVectorBackend` (the vector backend) has no such method — namespace creation lives on the relational backend. The test was written against an API that changed and never ran in CI to catch it. Fix: route namespace creation through the correct backend (or the coordinator) and re-enable.

### Acceptance
- [ ] Remove the `xfail` markers as each cluster is fixed and passes against the CI Postgres service.
- [ ] All previously-quarantined tests run green in `test-integration`.

filippetrovic:
### Cluster 3 — Postgres filter compiler `#>` / asyncpg incompatibility (product bug)
`tests/integration/test_compile_postgres_rowset.py` (9 cases)

[...]

PR #1026, opened 2026-06-08
Fix Postgres recall-filter compiler to bind JSONB path operand as text[]

filippetrovic:
## Summary
- The PostgreSQL recall-filter compiler rendered the `#>` / `#>>` JSONB path operand as a bare string literal, which asyncpg binds as `varchar`. Real Postgres then rejected `jsonb #> character varying` at execute time.
- This broke every metadata filter that compiles to the path-operator form: nested (`metadata.a.b`) paths, nested-path `$exists`, whole-subdocument equality, `$date` gates, and nested range comparisons. The single-level `@>` containment and `?` key-presence forms were unaffected (they never route through the path operand).
- Fix: bind the path segments as a `text[]` array via `ARRAY(Text)` so asyncpg transmits a native `text[]` over the protocol. This is the operand type `#>` / `#>>` require, and it also removes the manual array-literal escaping (the protocol encoder now handles framing, so segments containing `,` `{` `}` `"` `\` or whitespace round-trip verbatim).
- The unit tests assert the rendered SQL still emits the `#>` / `#>>` operator tokens; the fix only changes the operand binding, so that contract is preserved.
- Removed the integration xfail markers that previously quarantined these cases now that they execute correctly against real Postgres.

## Test plan
- [x] Unit tests pass — recall-filter compiler suite (`tests/unit/filter/`) green; rendered-SQL shape/contract assertions unchanged
- [x] Lint + format clean (`ruff`, `ruff format --check`, `ty`)
- [ ] Integration tests pass on real Postgres — the `test-integration` CI job (provisioned Postgres + asyncpg) runs the previously-xfailed nested-path / `$exists` / `$date` / range cases; these cannot run in a database-less local environment

### DIFF
--- a/src/khora/filter/compilers/postgres.py
+++ b/src/khora/filter/compilers/postgres.py
@@ -44,7 +44,7 @@ from typing import Any
 
 import sqlalchemy as sa
 from sqlalchemy import ColumnElement
-from sqlalchemy.dialects.postgresql import JSONB
+from sqlalchemy.dialects.postgresql import ARRAY, JSONB
 
 from khora.filter import (
     CompileContext,
@@ -428,14 +428,21 @@ class _Builder:
 
 
 def _jpath(segments: tuple[str, ...]) -> ColumnElement[Any]:
-    """Build the Postgres ``text[]`` path literal ``'{a,b}'`` for ``#>`` / ``#>>``.
+    """Build the Postgres ``text[]`` path operand for ``#>`` / ``#>>``.
 
-    Each segment is double-quoted (and its backslashes / quotes escaped) so a
-    segment containing ``,`` ``{`` ``}`` or whitespace survives the array-literal
-    parse.
+    The ``#>`` / ``#>>`` operators require a ``text[]`` right operand. Binding the
+    segments as a Python list typed ``ARRAY(Text())`` makes asyncpg transmit a
+    native ``text[]`` array over the protocol — no manual array-literal string,
+    so a segment containing ``,`` ``{`` ``}`` ``"`` ``\\`` or whitespace round-trips
+    verbatim (the protocol-level encoder handles framing, not us).
+
+    Prod impact of getting this wrong: only the ``#>`` / ``#>>`` path forms were
+    affected — nested ``metadata.a.b`` paths, nested-path ``$exists``, ``$date``
+    gates, exact-array ``$eq``, and nested range ops. The single-level ``@>``
+    containment and ``?`` key-presence forms never route through ``_jpath`` and
+    were unaffected.
     """
-    inner = ",".join('"' + s.replace("\\", "\\\\").replace('"', '\\"') + '"' for s in segments)
-    return sa.literal("{" + inner + "}")
+    return sa.literal(list(segments), type_=ARRAY(sa.Text()))
 
 
 def _jsonb_literal(value: Any) -> ColumnElement[Any]:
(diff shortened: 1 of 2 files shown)


filippetrovic:
## Consolidated Review — PR #1026

**Overall assessment: APPROVE.** CI is fully green on the latest commit (all 10 checks SUCCESS, including `test-integration` against real Postgres). The `#>`/`#>>` path-operand fix is verified end-to-end on real PG; the change is surgical (2 files), the SQL contract is preserved, and the security surface is reduced.

### CRITICAL
- _None._

### HIGH
- **[Correctness — discovered during review]** `tests/integration/test_compile_postgres_rowset.py` — Un-quarantining the row-set proof surfaced a **second, distinct compiler defect** (separate from this PR's `#> varchar` root cause): `_md_containment`'s `metadata @> '{k: v}'` form does **not** match a row whose field `k` is an *array* containing `v` on real Postgres (the JSONB array-contains-primitive exception doesn't apply at the nested-object key level). This affects the documented array-aware semantics of scalar `$eq` / `$in` / `$ne`. **Resolution in this PR:** the 4 affected `@>` cases are re-quarantined under a dedicated, accurately-labeled `_xfail_array_containment` marker (`strict=False`) tracked in **#1027** — *not* conflated with the `#>` bug or the prior `#1020` catch-all. The 5 genuine `#>`/`#>>` cases stay enabled and pass on real PG. Fixing the `@>` defect is out of scope here (it changes `_md_containment` semantics + the `@>` unit-SQL contract) and is tracked separately in #1027.

### MEDIUM
- **[Test hygiene]** `tests/integration/test_compile_postgres_rowset.py` — The original single `xfail` marker masked **two** different bugs (`#> varchar` + `@>` array containment). This PR correctly splits them: the path-operand cases are un-marked (fixed), and the containment cases get their own honestly-labeled marker → #1027. This resolves the devil's advocate's concern about removing markers without a clear signal: the remaining xfails are now precise and individually tracked.

### LOW
- **[Docs — devil's advocate]** `src/khora/filter/compilers/postgres.py:430` — The `_jpath` fix relies on asyncpg's native list→`text[]` protocol encoding (stable in `asyncpg>=0.31`, the repo's pinned floor; proven by the green `test-integration` run). Optionally note the version assumption near the binding. Non-blocking.
- **[Correctness — devil's advocate]** `_jpath(())` with an empty segments tuple would bind `[]`. Unreachable in practice — the bare-blob `len(path)==1` case is handled before path extraction, so every `#>`/`#>>` call site passes ≥1 segment. Informational only.

### INFO
- **[Security audit — PASS]** 32 added lines; no injection / secret / crypto / SSRF / deserialization / path-handling sinks. The one SQL-relevant line binds path segments as a **parameterized** native `text[]` array (never string-interpolated), which *reduces* the injection surface versus the prior hand-escaped string literal. The column token (`_col`) is unchanged and whitelisted. **Verdict: PASS.**
- **[Performance]** No index/plan regression: `@>`/`?` GIN-friendly forms are untouched; only the `#>`/`#>>` path-extraction operand binding changed (parameter type, not operator).

### Reviewer coverage note (transparency)
This review team dispatched 7 reviewers. The **devil's advocate** returned full findings (incorporated above) and the **security audit** (lead-run, Step 6) is complete and clean. The other 6 reviewers (security, performance, correctness, test, database, docs) did not return findings within the window — a transient disk-exhaustion event on the runner during their spawn left them unresponsive. Their domains are substantially covered here by: the dedicated security audit (security), the green `test-integration`/`test-unit`/`lint` CI (correctness, tests, performance, database), the upstream feature-team QA pass (51 + 103 unit tests, 92% touched-module coverage) and its approving code review, and the surgical 2-file diff (docs). The HIGH finding above (the `@>` defect, #1027) was caught via the real-PG CI signal during this phase, which is the exact gap the reviewers would have targeted.

sale87 (approved):

PR #1029, opened 2026-06-08
Add Chronicle recall-filter compiler with date-bound pushdown and Python post-filter

filippetrovic:
## Summary
- Add deterministic recall-filter support to the Chronicle engine: system date-key bounds push down to the engine's recency window, and all other predicates are enforced by an in-memory post-filter (default post-filter; no schema migration).
- New internal `compile_python` compiler — a pure `record -> bool` predicate that evaluates the canonical filter AST in-process, mirroring the Postgres compiler's field-match semantics: runtime type-gating, operator polarity with Mongo-style missing-key handling (`$ne`/`$nin` include absent/wrong-type), array-aware metadata containment, exact sub-document equality, typed-date parse-or-exclude, and presence/null distinctions. It is the in-process reference oracle for the filter compilers.
- New internal `compile_chronicle` compiler — lowers conjunctive `source_timestamp` range/`$eq` clauses into a narrowing `ChronicleDateBound`. `source_timestamp` is the only key whose bound is safe to push against the engine's `COALESCE(source_timestamp, created_at)` recency window; `occurred_at`/`created_at`, the denormalized document keys, and all metadata are left unconsumed and post-filtered (pushing a cross-dimension bound would silently drop matching rows).
- Wire both into `ChronicleEngine.recall()`: intersect the compiled date-bound with the recency window (narrow-only, never widening) before retrieval, then apply the `compile_python` post-filter to candidates after cross-session expansion and before top-k. A record adapter maps each chunk to the predicate's record shape, anchoring `occurred_at` to the chunk's event time (falling back to `source_timestamp` where the chunk row leaves it unset).
- Report an honest per-engine pushdown matrix on `RecallResult.engine_info["filter"]` (`pushed_down` / `post_filtered`), and document the per-key support matrix (pushdown / post-filter) in the engine + compiler docstrings.
- The unindexed-metadata recall counter increments on the post-filter path (once per metadata leaf, at compile time).
- Engine behavior is unchanged when no filter is supplied. The new compilers are internal (not exported from the top-level package surface).

## Test plan
- [x] Unit tests pass locally: `compile_python` field-match semantics, the Chronicle compiler pushdown split, and the engine composition (narrow-only window intersection, metadata post-filter, telemetry) — 577 passed / 8 skipped.
- [x] Lint + format clean (`make format`, `make lint`).
- [ ] Unit + integration jobs pass in CI (CI-provisioned Postgres).
- [ ] Manual verification: `recall(filter={"source_timestamp": {"$gte": ...}, "metadata.tag": {"$in": [...]}})` narrows correctly on a Chronicle lake.

### DIFF
--- a/src/khora/engines/chronicle/engine.py
+++ b/src/khora/engines/chronicle/engine.py
@@ -349,6 +349,104 @@ def _extract_temporal_bounds(temporal_filter: Any | None) -> tuple[datetime | No
     return _to_utc(start), _to_utc(end)
 
 
+def _intersect_lower(window: datetime | None, filter_bound: datetime | None) -> datetime | None:
+    """Intersect a lower bound with a filter's lower bound — narrow only.
+
+    Returns the LATER of the two (the tighter lower bound); ``None`` on either
+    side means unbounded below, so the other wins. The result is always >= the
+    incoming ``window`` lower bound — the filter can shrink the recency window
+    but never widen it. Both operands are coerced to tz-aware UTC so a zoneless
+    bound never raises against an aware one.
+    """
+    fb = _to_utc(filter_bound)
+    if fb is None:
+        return window
+    if window is None:
+        return fb
+    return max(window, fb)
+
+
+def _intersect_upper(window: datetime | None, filter_bound: datetime | None) -> datetime | None:
+    """Intersect an upper bound with a filter's upper bound — narrow only.
+
+    Returns the EARLIER of the two (the tighter upper bound); ``None`` on either
+    side means unbounded above, so the other wins. The result is always <= the
+    incoming ``window`` upper bound — narrow only, never widening.
+    """
+    fb = _to_utc(filter_bound)
+    if fb is None:
+        return window
+    if window is None:
+        return fb
+    return min(window, fb)
+
+
+# System keys the recall-filter post-filter reads off a chunk record. The eight
+# denormalized document keys live on the chunk's ``source_document`` projection
+# when present; ``DocumentSource`` only carries a subset (title / source /
+# source_type), so the rest resolve to absent on the legacy ``chunks`` path —
+# a documented limitation (positive predicates on them return empty there).
+_DOC_PROJECTION_KEYS: tuple[str, ...] = (
+    "source_type",
+    "source_name",
+    "source_url",
+    "external_id",
+    "content_type",
+    "source",
+    "title",
+)
+
+
+def _chunk_to_record(chunk: Chunk) -> dict[str, Any]:
+    """Map a :class:`Chunk` to the record dict the recall-filter post-filter reads.
+
+    The post-filter (``compile_python``) evaluates the full filter against this
+    record; ``compile_chronicle`` pushes the ``source_timestamp`` bound into the
+    recency window (its primary axis):
+
+    * ``occurred_at`` is the effective EVENT time ``COALESCE(occurred_at,
+      source_timestamp)`` — the chunk's real ``occurred_at`` column when carried
+      (sqlite_lance), recovered from ``source_timestamp`` when the literal field is
+      ``None`` (the legacy pgvector ``chunks`` DTO has no ``occurred_at`` column, so
+      the faithful event time there IS ``source_timestamp``). This recovery is why
+      an ``occurred_at`` filter does NOT false-empty on PG. Mirrors the engine's
+      ``RecallChunk.occurred_at`` surface derivation. ``occurred_at`` is enforced by
+      this post-filter, not pushed (it is the event-time axis, not the window axis).
+    * ``created_at`` and ``source_timestamp`` are the LITERAL column values (a
+      filter on those names is post-filtered against the real column).
+
+    The eight denormalized document keys are read off ``chunk.source_document``
+    when present (``DocumentSource`` carries only a subset; the rest stay absent,
+    so a positive predicate on them returns empty on the legacy path — documented
+    limitation, no faithful fallback exists). ``metadata`` is the chunk's own dict.
+    """
+    record: dict[str, Any] = {
+        # Effective event time = COALESCE(occurred_at, source_timestamp). occurred_at
+        # is conceptually PRESENT (it's the event time, == source_timestamp on the
+        # legacy path where the DTO leaves the literal field None); the adapter
+        # recovers it so an occurred_at filter does not false-empty. No created_at
+        # fallback — ingest time is not event time.
+        "occurred_at": chunk.occurred_at if chunk.occurred_at is not None else chunk.source_timestamp,
+        "created_at": chunk.created_at,
+        "source_timestamp": chunk.source_timestamp,
+        "metadata": chunk.metadata or {},
+    }
+    # Resolve each denorm doc key defensively: chunk attr → chunk.source_document
+    # attr → absent. The Chunk dataclass does not carry these today (they live on
+    # the source_document projection), but trying the chunk first is robust if a
+    # future DTO denormalizes them onto the chunk. A key found nowhere stays absent
+    # → compile_python's §4 missing-semantics → positive predicate empty
+    # (documented (v) limitation; no faithful fallback exists for these).
+    source_doc = chunk.source_document
+    for key in _DOC_PROJECTION_KEYS:
+        value = getattr(chunk, key, None)
+        if value is None and source_doc is not None:
+            value = getattr(source_doc, key, None)
+        if value is not None:
+            record[key] = value
+    return record
+
+
 def _temporal_proximity(
     referenced_date: datetime | None,
     start: datetime | None,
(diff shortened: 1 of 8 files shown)


filippetrovic:

Eight review perspectives (security, performance, correctness, database, observability, docs, test, devil's-advocate) plus a dedicated security audit. **Overall: APPROVE** — no blocking defects; CI is green (10/10). One MEDIUM is worth a 2-line pre-merge fix; the rest are documentation/defensive/design notes.

### CRITICAL
- None.

### HIGH
- None confirmed as live defects. (A reviewer initially flagged the date-bound intersection as HIGH; analysis below downgrades it to MEDIUM — it is not reachable as a live crash because the sole call site pre-normalizes the window.)

### MEDIUM
- **M1 [correctness] `engines/chronicle/engine.py:352-381`** — `_intersect_lower`/`_intersect_upper` coerce only `filter_bound` to UTC, but the docstring promises *both* operands are coerced. Not a live bug: the only call site (`:1729-1730`) receives a window already UTC-normalized by `_extract_temporal_bounds` (`:349`), so a naive/aware mismatch can't occur on the engine path. Still a docstring/code mismatch + latent fragility if reused. **Recommend** coercing `window` too (e.g. `return max(_to_utc(window), fb)` / `return _to_utc(window)`), making the code match its stated contract.
- **M2 [architecture/devil's-advocate] `filter/compilers/python.py`** — the Python compiler is a complete field-match oracle; its oracle role (validating the SQL/Cypher compilers) has no in-repo consumer yet, while this engine only uses it for post-filtering. The completeness is intentional (single source of truth for filter semantics), but consider a short note pointing at the intended future consumer so it doesn't read as speculative surface.
- **M3 [devil's-advocate] `engines/chronicle/engine.py:429`** — `occurred_at = occurred_at or source_timestamp` is a silent event-time fallback (correct, and documented in the docstring), but a reader of the call site gets no signal. **Recommend** a one-line inline comment; optionally surface the fallback in `engine_info["filter"]`.
- **M4 [observability/performance] `filter/compilers/python.py:192`** — `recall.filter.unindexed_metadata` fires once per metadata leaf (per `FilterClause`) keyed by `op`, at compile time. Confirm per-clause (vs per-unique-path) is the intended cardinality contract; it matches the documented "once per metadata leaf" and the `op`-only label is cardinality-safe.
- **M5 [devil's-advocate] `filter/compilers/chronicle.py:20-39`** — the "only `source_timestamp` is same-axis-safe" rule is sound but couples the compiler to this engine's `COALESCE(source_timestamp, created_at)` retrieval window. **Recommend** a docstring note that the consumed-key set must be revisited if the window axis ever changes.

### LOW
- **L1 [security] `engine.py:361-366`** — defensive double-normalize reads slightly awkwardly; maintainability note only (no vulnerability).
- **L2 [security] `python.py`** — metadata vs system-key resolution use separate paths; no key-collision/isolation breach (intended design).
- **L3 [devil's-advocate] `engine.py:351-381`** — consider a property-based test asserting the narrow-only invariant (result is always within the incoming window) to guard against an accidental min/max swap.
- **L4 [devil's-advocate] `engine.py` doc-projection keys** — the denormalized doc-key tuple is hand-maintained; add a comment flagging it as a manual sync point if the document projection grows.
- **L5 [devil's-advocate] `engine.py:2000-2012`** — the post-filter runs after cross-session expansion (correct, ensures exactness), but a highly selective filter can over-fetch expansion candidates then discard them; acceptable and already commented.

### Security Audit
Clean. No dependency manifests changed; the new modules import only the standard library + internal modules; no `eval`/`exec`/`subprocess`/`pickle`/`os.system`; no OWASP Top-10 surface (no auth/access-control/injection/deserialization/SSRF changes) and no cryptographic code. Filter operands are validated at the model boundary; the date-bound compiler emits `datetime` bounds (not query strings) and the post-filter is a pure in-memory predicate — no injection vector. Namespace isolation is preserved: the filter only narrows (AND-composed), enforced mathematically by `max`/`min` window intersection.

### Per-dimension outcomes
- **Security:** approve (3 LOW observations). **Performance:** approve — predicate compiled once per recall; per-record adapter O(1); zero overhead on the no-filter path. **Correctness:** sound; one MEDIUM (M1). **Database:** approve — no coordinator signature change, no migration, source-timestamp pushdown is superset-safe against the window. **Observability:** approve — honest `engine_info["filter"]` matrix; cardinality-safe counter; telemetry contract unchanged (no new metric definition). **Test:** thorough — field-match semantics, pushdown split, and engine composition all covered, hermetic fixtures, regressions for key-resolution + unanchored-exclusion + the metadata counter. **Docs:** support-matrix docstring present and accurate; no internal-identifier leakage. **Devil's-advocate:** architecturally sound; scope/transparency notes captured as M2/M3/M5.

**Recommendation:** APPROVE for merge. Suggested pre-merge: M1 (2-line window coercion). M2–M5 + LOWs are non-blocking follow-ups.

filippetrovic:
Thanks for this — the `compile_python` post-filter is thorough and the date-axis (`COALESCE(occurred_at, source_timestamp)`) reasoning is carefully worked out. A few things worth resolving before merge:

**1. `_consumable_date_clause` docstring contradicts the code.**
`chronicle.py` defines `_PUSHDOWN_DATE_KEYS = frozenset({"source_timestamp"})`, and `_consumable_date_clause` rejects any leaf whose key isn't in that set — so the only date key actually pushed down is `source_timestamp`, with `occurred_at` / `created_at` left to the Python post-filter. But that function's docstring says *"A pushdownable clause is a leaf on `occurred_at` / `created_at`…"*, and the module-level pushdown summary reads the same way. A reader can't tell whether `source_timestamp`-only pushdown is intended or a bug. If it is intended, please update both docstrings to match the code — right now they point the opposite way.

**2. The engine-composition tests mock the storage layer.**
`test_chronicle_filter_composition.py` drives `ChronicleEngine` with a `MagicMock()` storage whose `search_similar_chunks` / `search_fulltext_chunks` return hand-built lists, so the filter is exercised against fabricated candidate sets rather than real retrieval over the embedded SQLite + LanceDB store. That skips exactly the parts most likely to break: metadata serialization round-trips and the `COALESCE(occurred_at, source_timestamp)` recovery against real columns. The repo already has an embedded-store fixture used elsewhere — an end-to-end `remember()` → `recall(filter=…)` over a real embedded Chronicle would be a much stronger guard than the mocked candidate lists.

**3. `importorskip` on first-party modules that ship in this PR.**
The new test modules use `pytest.importorskip(...)` for the `chronicle` / `python` compilers that land in this same PR. The sibling `tests/integration/test_compile_postgres_rowset.py` documents the opposite convention:

> Hard import (NOT importorskip): the compiler is on the branch, so an import failure must be a LOUD test error — never a silent module skip.

These should be hard imports for the same reason — otherwise a broken import silently skips the entire module instead of failing.

**4. (Nice-to-have) cross-compiler parity.**
There's a hand-authored oracle for `compile_python`, but nothing asserts the Chronicle path returns the same surviving row IDs as the other per-backend compilers for the same filter. Given several compilers have to agree, a same-filter → same-IDs parity check would be a high-value addition here.

Happy to take another look once these are addressed.

filippetrovic:
Thanks — all four addressed (pushed in `59d1a66`, CI green 10/10 incl. the integration job):

**1. Docstring vs code (`source_timestamp`-only pushdown)** — resolved. `_consumable_date_clause` and the module-level summary now state plainly that **only `source_timestamp`** (the window's primary axis) is pushed down; `occurred_at` / `created_at` are cross-dimensional and post-filtered. The implementation was already correct; the docs now match.

**2. Real embedded e2e (not mocked storage)** — added `tests/integration/test_chronicle_filter_embedded.py`: a real SQLite + LanceDB coordinator (the existing `_sqlite_lance_fixtures` helper), seeded via the coordinator's own `create_chunks_batch`, driven through `ChronicleEngine.recall(filter=…)`. Covers metadata serialization round-trip, `source_timestamp` pushdown against the real `COALESCE(source_timestamp, created_at)` window, and the `occurred_at` recovery against real columns. Runs in the `test-integration` CI job (green).

   This immediately earned its keep: it surfaced that the **embedded sqlite_lance write path does not persist a distinct `occurred_at`** (the `create_chunks_batch` INSERT in `sqlite_lance/vector.py` writes `source_timestamp` but omits `occurred_at`, so it reads back `NULL`). The filter contract is still correct there — `occurred_at` post-filters against `COALESCE(occurred_at, source_timestamp)`, so it faithfully recovers via `source_timestamp` and never false-empties — but a *distinct* `occurred_at` value isn't honored on that backend. That's a pre-existing storage-layer gap, independent of this filter PR; I've documented it in the test and will file it separately so it doesn't get lost.

**3. `importorskip` → hard imports** — done. All three `tests/recall/` modules now hard-import the `chronicle` / `python` compilers (and dropped the now-stale interim-skip docstrings), matching the `test_compile_postgres_rowset.py` convention — a broken import is a loud error, not a silent module skip.

**4. Cross-compiler parity** — added `test_engine_recall_matches_compile_python_oracle`: it asserts the Chronicle engine's real-storage recall returns the **same surviving chunk IDs** as the in-process `compile_python` oracle for the same composed filter (`metadata.tier` + `source_timestamp`).

msrdic (approved):

PR #1031, opened 2026-06-08
Provision Neo4j in the integration CI job and enable the Neo4j-backed tests

filippetrovic:
## Summary
- Add a `neo4j:2025.12.1` service side-car to the `test-integration` CI job (alongside the existing Postgres service), with matching `KHORA_NEO4J_URL` / `KHORA_NEO4J_USERNAME` / `KHORA_NEO4J_PASSWORD` env and `NEO4J_INTEGRATION_TEST=1`, plus a bolt-port readiness wait. This flips the per-module skip gate so the Neo4j-backed integration tests (~55 functions across 13 files) run in CI instead of self-skipping.
- Add a loud-abort guard in `tests/integration/conftest.py` mirroring the existing Postgres `KHORA_PG_REQUIRED` tripwire: when `NEO4J_INTEGRATION_TEST=1` but Neo4j is unreachable, the pytest session fails loudly (`pytest.exit(returncode=1)`) instead of silently skipping. The guard is a no-op when the flag is unset, so the unit job and local dev without `make dev` keep skipping cleanly.
- Fix two `namespace_id` drift bugs surfaced by enabling the tests: `Neo4jBackend.get_neighborhoods_batch` (3 call sites) and `StorageCoordinator.get_entity_relationships` (1 call site) both require a keyword-only `namespace_id`; the never-run tests omitted it and would have raised `TypeError`.

Note: local `make dev` exposes Neo4j bolt on `7688`; CI uses `7687` (matching the test-module defaults). Set `KHORA_NEO4J_URL` to override locally. The Postgres service and its `KHORA_PG_REQUIRED` guard are unchanged.

## Test plan
- [x] `ruff check` + `ruff format --check` clean; `ty check src/` clean
- [x] Integration suite collects cleanly; the new conftest guard is a verified no-op when `NEO4J_INTEGRATION_TEST` is unset (280 tests collect, exit 0)
- [x] Loud-abort path verified: flag set + Neo4j unreachable → session exits 1 with a clear message
- [ ] CI `test-integration` job runs the Neo4j-backed tests against the new side-car (first-ever CI execution)
- [ ] Unit job and other jobs unaffected

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -315,10 +315,18 @@ jobs:
     runs-on: ubuntu-latest
     timeout-minutes: 25
     # Job-level env so every step (migrations, pytest, the conftest guard)
-    # inherits the same Postgres DSN and the loud-DB-down flag.
+    # inherits the same Postgres DSN and the loud-DB-down flag. The Neo4j
+    # vars match the contract the integration tests read (KHORA_NEO4J_URL /
+    # _USERNAME / _PASSWORD, defaults bolt://localhost:7687, neo4j, password)
+    # and NEO4J_INTEGRATION_TEST=1 flips the per-module skipif so the Neo4j
+    # graph-backend tests actually run against the side-car.
     env:
       KHORA_DATABASE_URL: postgresql://khora:khora@localhost:5434/khora
       KHORA_PG_REQUIRED: "1"
+      KHORA_NEO4J_URL: bolt://localhost:7687
+      KHORA_NEO4J_USERNAME: neo4j
+      KHORA_NEO4J_PASSWORD: password
+      NEO4J_INTEGRATION_TEST: "1"
     services:
       postgres:
         image: pgvector/pgvector:pg17
@@ -333,6 +341,21 @@ jobs:
           --health-interval 5s
           --health-timeout 5s
           --health-retries 10
+      neo4j:
+        image: neo4j:2025.12.1
+        env:
+          NEO4J_AUTH: neo4j/password
+        ports:
+          - 7687:7687
+        # wget the HTTP endpoint on 7474 — mirrors the compose.yaml
+        # healthcheck for this image. neo4j images take ~20-40s to start,
+        # so a generous retry budget keeps the gate from going green early.
+        options: >-
+          --health-cmd "wget -q --spider http://localhost:7474 || exit 1"
+          --health-interval 10s
+          --health-timeout 5s
+          --health-retries 10
+          --health-start-period 30s
     steps:
       - name: Checkout code
         uses: actions/checkout@v6
@@ -369,6 +392,23 @@ jobs:
             sleep 2
           done
           echo "postgres did not become ready within 60s"
+          exit 1
+
+      - name: Wait for Neo4j readiness
+        # Mirrors the Postgres/Weaviate readiness loops. The services
+        # health-check already gates step start, but neo4j images take
+        # 20-40s to boot, so poll the bolt port (7687) here for a clear,
+        # early failure message if the side-car is somehow not up. Uses a
+        # bash /dev/tcp probe so no extra client (cypher-shell) is needed.
+        run: |
+          for i in $(seq 1 30); do
+            if (echo > /dev/tcp/localhost/7687) >/dev/null 2>&1; then
+              echo "neo4j bolt ready"
+              exit 0
+            fi
+            sleep 2
+          done
+          echo "neo4j did not become ready within 60s"
           exit 1
 
       - name: Apply database migrations
--- a/tests/integration/conftest.py
+++ b/tests/integration/conftest.py
@@ -1,10 +1,14 @@
 """Pytest configuration for the integration test job.
 
-Loud-DB-down guard: when ``KHORA_PG_REQUIRED=1`` is set (the CI integration
+Loud-DB-down guards: when ``KHORA_PG_REQUIRED=1`` is set (the CI integration
 job sets it), a missing or misconfigured Postgres must FAIL the run, not pass
-by skipping the PG-gated tests. Without the env var (a local dev box without
-``make dev``) the guard is a no-op, so the existing per-module ``skipif`` still
-lets those tests skip cleanly.
+by skipping the PG-gated tests. The same applies to Neo4j: when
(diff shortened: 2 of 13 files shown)


filippetrovic:
## Consolidated Review — multi-perspective team + security audit

**Overall: ✅ APPROVE.** Seven reviewers (security, performance, correctness, test, docs, database, devil's advocate) plus a dedicated security audit. One CRITICAL and one LOW finding were surfaced and **already fixed in this PR** (commit `bbe565a`); the remaining items are deliberate design choices or optional, non-blocking enhancements.

### CRITICAL
- **C1 — [Correctness/Test/Devil's-Advocate] `test_neo4j_get_entity_relationships_integration.py:109`** — the call `backend.get_entity_relationships(entity_a.id, direction="outgoing")` omitted the now-required keyword-only `namespace_id` (made mandatory in #769). It would `TypeError` on the first CI run — exactly the never-run-in-CI drift this PR exists to flush out. ✅ **Fixed** (added `namespace_id=namespace_id`). An exhaustive scan of all 13 gated files confirmed this was the only remaining missed site (`retire_orphaned_entities_batch` takes no `namespace_id`; the mock `DualNodeManager` calls pass namespace positionally).

### MEDIUM
- **M1 — [Performance] Redundant readiness double-check** (`ci.yml`, services health-check + "Wait for Neo4j readiness" step, ~60s). *Disposition: intentional* — mirrors the existing Postgres readiness step in the same job (deep health-check gates step start; the bolt-port poll gives a clear early failure message). Kept for consistency with repo convention.
- **M2 — [Performance] `-n 0` serial run not explained.** *Disposition: out of scope* — `-n 0` is pre-existing (added by #1016) because PG tests `DROP SCHEMA public CASCADE`; this PR doesn't change the pytest invocation.
- **M3 — [Devil's-Advocate] `NEO4J_INTEGRATION_TEST` serves dual role** (test-activation flag + loud-abort gate) vs a dedicated `KHORA_NEO4J_REQUIRED`. *Disposition: deliberate, won't-fix.* For Neo4j the per-module `skipif` already keys on this single flag, so it is the natural opt-in: "if you asked to run the Neo4j suite, an unreachable Neo4j fails loudly." A second variable adds surface for no functional gain, and the Correctness + Test reviewers both confirmed the guard fires correctly and is a true no-op when unset. (The Postgres split exists only because PG tests have no activation flag.)
- **M4 — [Devil's-Advocate] Fold Neo4j into `test-integration` vs a separate job** (weaviate pattern). *Disposition: deliberate* — ~5 PG+Neo4j tests need both services in one job, and provisioning here is the explicit intent of the change.

### LOW / INFO
- **L1 — [Docs] Five test-module docstrings still claimed "CI does not provision Neo4j."** ✅ **Fixed** (commit `bbe565a`) — refreshed to reflect that the integration job now provisions Neo4j and sets the flag.
- **L2 — [Security/Devil's-Advocate] Hardcoded `password` in CI YAML** (`NEO4J_AUTH=neo4j/password`, `KHORA_NEO4J_PASSWORD=password`). *Acceptable* — a throwaway credential for an ephemeral CI sidecar with no real data, consistent with the existing `POSTGRES_PASSWORD: khora`. Optional follow-up: centralize both PG and Neo4j CI credentials into GitHub Secrets together.
- **L3 — [Performance] Health-check retry budget generous** (30s start + 100s retries). *Acceptable* — matches `compose.yaml`; prevents false negatives on Neo4j's 20–40s cold start.
- **L4 — [Database] Optional one-time `MATCH (n) DETACH DELETE n` purge** between migrations and tests. *Nice-to-have, not required* — tests are namespace-isolated (unique `uuid4()` per test, `finally` cleanup), and the CI Neo4j has no persistence volume, so orphaned nodes are inert.

### Security Audit (Step 6)
Scanned the 91 added lines across 9 files. No dependency/manifest changes; no new endpoints, redirects, injection sinks, SSRF, or deserialization. Only the readiness probe (`/dev/tcp/localhost/7687`, constant host:port — no user input) and the LOW test-credential item above. **Verdict: PASS.**

### What's solid
Conftest loud-abort guard correctly mirrors the Postgres `KHORA_PG_REQUIRED` tripwire and is a verified no-op when the flag is unset; env-var names match across `ci.yml`, the test modules, and the guard; Neo4j `2025.12.1` service config (auth, bolt port, health-check) is correct; test isolation (unique namespace per test + `finally` cleanup) is sound for serial shared-instance execution.

*Note: the Neo4j-backed suite runs for the first time ever in CI on this PR — the remaining empirical triage is the CI `test-integration` job here.*

probablyrobot (approved):

filippetrovic:
## CI triage update — first-ever Neo4j suite run

The `test-integration` job ran the Neo4j-backed suite for the first time: **251 passed**, confirming provisioning + the loud-abort guard work. Triaged the 18 failures (commit `7f8995e`):

**Fixed**
- `test_replace_document_extraction.py` / `test_forget_cascade_orphans_integration.py` — the async `kb` fixtures were `scope="class"`; under `asyncio_mode=auto` (per-function event loops) the class-scoped asyncpg pool was bound to a closed loop, erroring every test at setup (`got Future … attached to a different loop`). Function-scoped them. *(Note: `@pytest.mark.xfail` cannot absorb fixture-setup errors, so this fix was mandatory, not optional.)*

**Quarantined as tracked `xfail(strict=False)` → #1033** (they run, not skip; visibly xfail/xpass — no silent skips):
- PG namespace FK violations: `test_coordinator_replace_document_extraction.py` (2), `test_graph_namespace_isolation_integration.py` (2)
- Source-doc-id remap count mismatch: `test_neo4j_remap_source_document_ids_integration.py` (2)
- `prefer_current` future-`valid_until` filtering: `test_neo4j_neighborhood_valid_until_integration.py` (1)
- Cross-window entity counting: `test_windowed_processing_submit_batch.py` (1)
- The replace/forget classes are also xfailed pending confirmation their downstream bodies pass once the loop-scope fix lands.

These are real pre-existing issues in never-run-in-CI tests, not regressions from this PR. **#1033** tracks fixing them against a live PG+Neo4j stack (this sandbox has no Docker, so empirical iteration belongs in that follow-up). This satisfies the ticket's AC: the Neo4j files now **execute** in CI and either pass or are tracked-xfail with an issue.

PR #1032, opened 2026-06-08
Make Postgres migration integration tests shared-DB-safe and fix stale namespace API

### LINKED ISSUE #1020: Triage Postgres-integration test failures surfaced by the new test-integration CI job

filippetrovic:
## Summary

The new integration CI job runs the previously self-skipping Postgres-backed tests for the first time, against a shared service database already migrated to head. That surfaced two clusters of pre-existing test-harness failures that had been quarantined as `xfail`. This clears them.

**Cluster 1 — migration tests assumed a non-shared DB.** Five migration tests reset the alembic version pointer and dropped a single table, which collided with the up-front `alembic upgrade head` on the shared DB (`DuplicateTableError`). Each test's Postgres setup is now self-contained: it drops public-schema enum types, runs `DROP SCHEMA public CASCADE`, recreates the schema + `vector` extension, pre-creates the version table, and re-migrates — mirroring the existing bitemporal migration test. Downgrade checks pin explicit predecessor revisions rather than a `-1` step count so they stay correct as the chain grows.

**Cluster 2 — stale backend API.** The chunk namespace-isolation test called `create_namespace` on the vector backend, which has no such method. Namespace and document creation now route through the relational backend via a dedicated fixture, while the chunk getters under test stay on the vector backend so SQL-level namespace isolation is still exercised. Stale `INSERT` columns and a couple of asyncpg type-handling issues (timestamptz binding, `"char"` cast) that only surfaced once the tests actually ran against Postgres are corrected.

All `xfail` markers for both clusters are removed; the tests run green, not xfailed.

Fixes #1020

## Test plan
- [x] `make lint` passes (ruff + ty), exit 0
- [x] The 6 affected test files pass against Postgres (33 passed, 0 failed, 0 xfailed), verified repeatedly
- [ ] CI `test-integration` job green on the shared service Postgres
- [ ] CI `test-unit`, lint, security, install, examples-smoke jobs green

### DIFF
--- a/tests/integration/db/test_migration_032_dream_runs.py
+++ b/tests/integration/db/test_migration_032_dream_runs.py
@@ -32,7 +32,7 @@ import pytest
 import sqlalchemy as sa
 from alembic import command
 from alembic.config import Config
-from sqlalchemy.ext.asyncio import create_async_engine
+from sqlalchemy.ext.asyncio import AsyncEngine, create_async_engine
 
 from khora.db.session import run_migrations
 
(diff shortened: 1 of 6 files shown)


filippetrovic:

Reviewed by an agent team across security, performance, correctness, tests, database, docs, and a devil's-advocate pass, followed by a dedicated security audit of the diff. This is a **test-only** change (6 integration test files; no `src/` or docs changes), verified green against Postgres (33 passed / 0 failed / 0 xfailed, repeatedly).

**Overall assessment: APPROVE** — no blocking findings. Items below are non-blocking (hygiene + a suggested follow-up).

### CRITICAL
_None._

### HIGH
_None._

### MEDIUM
- **M1 — [Performance] Full schema-wipe + full-chain re-migration per test.** Each migration test's setup runs `DROP SCHEMA public CASCADE` + re-migrates the full chain (function-scoped); the namespace-isolation test does it once (module-scoped). Estimated ~40s added to the serial integration job. **Accepted trade-off:** a narrower reset is what caused the original `DuplicateTableError` on the shared DB, and full re-migration is the established pattern (mirrors the existing bitemporal migration test). Function scope is required for the migration tests — each upgrade/downgrade/skip-ahead case needs a clean baseline. No change requested.

### LOW
- **L1 — [Security] Dynamic SQL via f-string** (`DROP TYPE IF EXISTS public.{typname} CASCADE`, all reset helpers). The interpolated `typname` comes from the `pg_type` **system catalog** (enum type names), never user input, in a test fixture against a local DB — so real injection risk is zero. Optional hardening: quote the identifier (`public."{typname}"`). One reviewer rated this MEDIUM on pattern grounds; downgraded to LOW given the fully-trusted source and test-only scope.
- **L2 — [Tests] No explicit fixture teardown.** Isolation is achieved by the setup-time schema wipe (standard function-scoped behavior; matches the existing bitemporal test). Correct as-is; optionally add a one-line docstring noting the reset-on-setup contract.
- **L3 — [Architecture] Cluster-2 dual-backend fixture.** Namespace/document creation routes through the relational backend while chunk getters stay on the vector backend. This is intentional and exactly the right shape for this test — it exercises the vector backend's namespace filter at the SQL boundary directly (not via a coordinator proxy). Noted as a design question by the devil's advocate; confirmed sound by the correctness/test/database reviewers.
- **L4 — [Coverage] Migrations 028/029 are exercised only indirectly** (via the full-chain replay), with no dedicated test. Pre-existing gap, not introduced here. A separate, pre-existing migration-infra fragility (alembic version-stamping across the autocommit `CREATE INDEX CONCURRENTLY` migration on repeated back-to-back full-chain replay) is **tracked as a separate follow-up** — it does not affect CI, which runs the suite single-pass on a dedicated database (the pre-existing bitemporal test already replays the same chain there and is green). Suggested follow-up: add dedicated coverage for that migration.

### INFO
- **[Docs]** No documentation impact — test-only change; existing "Test Infrastructure Isolation" guidance already covers the shared-DB pattern.
- **[Security]** `DATABASE_URL` test-fixture default embeds `khora:khora@localhost` credentials but is env-overridable and test-only — acceptable for test infrastructure, not a production secret.

## Security Audit
Diff scanned across the OWASP-aligned categories (added/modified lines only). Only `tests/` touched — no new endpoints, trust boundaries, secrets in production code, weak crypto, SSRF/file/deserialization/redirect sinks. Single flagged pattern is the f-string DDL in **L1**, mitigated by its trusted system-catalog source. **Verdict: PASS.**

---
_Findings are advisory. The change is correct, scoped, and verified; the LOW items are optional hygiene and the coverage suggestion is a follow-up, none blocking merge._

sale87 (approved):

PR #1035, opened 2026-06-08
Match array-valued metadata fields in Postgres JSONB containment filter

### LINKED ISSUE #1027: compile_postgres: @> containment does not match nested array membership on real Postgres
## Summary
The PostgreSQL recall-filter compiler's array-aware containment (`_Builder._md_containment` in `src/khora/filter/compilers/postgres.py`) emits `metadata @> '{"<key>": <value>}'` for scalar `$eq`, `$in`, and (negated) `$ne`. On **real Postgres**, this does **not** match a row whose `<key>` holds an **array** that contains `<value>`.

Example: with `metadata = {"tier": ["gold","silver"]}`, the predicate `metadata @> '{"tier":"gold"}'` returns **false** — the JSONB array-contains-primitive exception (`'["gold"]'::jsonb @> '"gold"'` → true) does **not** apply at the nested-object key level (`'{"tier":["gold"]}'::jsonb` is required to match an array value).

## Impact
The field-match contract documents scalar `$eq` / `$in` as *array-aware* (a scalar operand should match both a scalar field equal to it and an array field containing it). The current `@>` form only matches the scalar shape, so array-valued metadata fields are missed (and `$ne` wrongly includes them). Affected ops: scalar `$eq`, `$in`, `$nin`/`$ne` array membership.

## Likely fix
`_md_containment` should OR the scalar-doc form with an array-wrapped form, e.g. `metadata @> {k: v}  OR  metadata @> {k: [v]}`, so both `tier = "gold"` and `tier = ["gold", ...]` match. This changes the emitted SQL, so the `@>` unit-test assertions in `tests/unit/filter/test_compile_postgres.py` need updating, and the change touches the documented field-match contract — warrants review.

## Discovery
Surfaced while un-quarantining `tests/integration/test_compile_postgres_rowset.py` in the `#> varchar` path-operand fix (PR #1026). That PR fixes a **distinct** root cause (`jsonb #> character varying`). These 4 `@>`-containment cases fail for this separate reason and are re-quarantined under a clearly-labeled marker pointing here, pending this fix:

- test_metadata_scalar_eq_matches_array_membership
- test_metadata_in_matches_scalar_and_array_membership
- test_metadata_in_multiple_values
- test_metadata_ne_includes_missing_key

filippetrovic:
## Summary
- The Postgres recall-filter compiler's metadata containment (`_md_containment`) emitted only `metadata @> '{"k": v}'` for a scalar operand. On real Postgres, JSONB containment does **not** treat `{"k": "x"}` as contained by a row whose `k` holds an array `["x", …]` — the array-contains-primitive exception (`'["x"]'::jsonb @> '"x"'`) applies only when the operand itself is an array, not at a nested object key. So scalar `$eq`/`$in` (and the negated `$ne`/`$nin`) silently missed array-valued metadata fields — e.g. `metadata.tags $in [...]` against the common `tags: list[str]` shape matched nothing.
- **Fix:** OR the scalar-doc containment with an array-wrapped form whose leaf is a one-element list — `metadata @> '{"k": v}' OR metadata @> '{"k": [v]}'` — so a scalar operand matches both a scalar field equal to `v` and an array field containing `v`. Nested paths wrap the one-key doc at each segment. Both arms are `@>` probes, so a GIN index on `metadata` still serves the OR (the planner bitmap-ORs the two index scans), and each arm is a total boolean, so the `$ne`/`$nin` negation still admits absent-key rows.
- Strengthened the compiler unit tests to assert **both** `@>` doc forms are emitted for scalar `$eq` and for each `$in` value; left the exact-array `$eq` (`#> = jsonb`) and `$ne` negation assertions intact.
- Removed the integration `xfail` markers that quarantined the four array-membership row-set cases.
- **No Cypher change needed:** the Cypher backend post-filters metadata predicates in Python (Neo4j stores metadata as a serialized JSON string property, not a nested map), so array membership is already handled there; only scalar system keys compile to Cypher. The pgvector skeleton backend reuses this compiler directly and inherits the fix.

## Test plan
- [x] Unit tests pass — `uv run pytest tests/unit/filter/` → 111 passed; rendered SQL verified to emit both containment forms for `$eq`/`$in`/`$ne`/`$nin`; exact-array `$eq` still routes to `#> = jsonb`
- [x] Lint + format clean — `make lint`, `make format`
- [ ] Integration tests pass on real Postgres — the four previously-xfailed array-membership row-set cases run in the `test-integration` CI job (Postgres not reachable in the local dev environment)

Fixes #1027

### DIFF
--- a/src/khora/filter/compilers/postgres.py
+++ b/src/khora/filter/compilers/postgres.py
@@ -362,16 +362,36 @@ class _Builder:
     def _md_containment(self, segs: tuple[str, ...], value: Any) -> ColumnElement[bool]:
         """Array-aware ``@>`` containment rooted at the metadata column.
 
-        For a single-segment path ``("tags",)`` this is ``metadata @> '{"tags":
-        v}'`` — which matches both a scalar field equal to ``v`` and an array
-        field containing ``v``. For a nested path the one-key doc is wrapped at
-        each segment (``{"a": {"tags": v}}``). ``@>`` returns a total boolean
-        (``FALSE`` on absent/mismatch, never ``NULL``), so it is negation-safe.
+        JSONB ``@>`` does NOT treat ``'{"tags": "x"}'`` as contained by
+        ``'{"tags": ["x"]}'`` — the array-contains-primitive exception applies
+        only when the operand itself is an array, not at a nested object key. So
+        a single ``metadata @> '{"tags": v}'`` misses an array-valued field. To
+        match BOTH a scalar field equal to ``v`` and an array field containing
+        ``v``, OR the scalar-doc form with an array-wrapped form whose leaf value
+        is a one-element list:
+
+            ``metadata @> '{"tags": v}'  OR  metadata @> '{"tags": [v]}'``
+
+        For a nested path the one-key doc is wrapped at each segment; the array
+        form wraps the LEAF value (``{"a": {"tags": v}}`` / ``{"a": {"tags":
+        [v]}}``). Each ``@>`` is a total boolean (``FALSE`` on absent/mismatch,
+        never ``NULL``), so the OR of two totals is total and negation-safe.
         """
-        doc: Any = _date_to_jsonable(value)
+        leaf = _date_to_jsonable(value)
+        scalar_doc: Any = leaf
+        array_doc: Any = [leaf]
         for seg in reversed(segs):
-            doc = {seg: doc}
-        return self._md.op("@>")(_jsonb_literal(doc))
+            scalar_doc = {seg: scalar_doc}
+            array_doc = {seg: array_doc}
+        # Both arms are `@>` probes, so a GIN index on `metadata` still serves
+        # this OR (the planner bitmap-ORs the two index scans) — do not collapse
+        # it into a non-`@>` form that loses the index. ``value is None`` only
+        # reaches here from `{k: {$in: [null, ...]}}` (a `{k: null}` $eq is routed
+        # to `_md_null` before this), where the array arm just adds `@> {k:[null]}`.
+        return sa.or_(
+            self._md.op("@>")(_jsonb_literal(scalar_doc)),
+            self._md.op("@>")(_jsonb_literal(array_doc)),
+        )
 
     def _md_exists(self, segs: tuple[str, ...]) -> ColumnElement[bool]:
         """Presence test for a metadata path — total boolean.
--- a/tests/integration/test_compile_postgres_rowset.py
+++ b/tests/integration/test_compile_postgres_rowset.py
@@ -80,23 +80,6 @@ pytestmark = [
     pytest.mark.skipif(not _pg_reachable(), reason="PostgreSQL not reachable (run `make dev`)"),
 ]
 
-# These cases assert array-aware ``@>`` containment: a scalar operand (``$eq`` /
-# ``$in`` / negated ``$ne``) should match BOTH a scalar field and an array field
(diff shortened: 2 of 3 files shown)


filippetrovic:

**Team:** security · performance · correctness · database · test · docs · devil's-advocate (7 reviewers) + dedicated security audit.
**Overall assessment: ✅ APPROVE** — no CRITICAL or HIGH findings. The fix is correct, minimal, well-documented, GIN-indexable, and covered by unit + (now-unmarked) integration tests. Remaining items are non-blocking coverage/documentation follow-ups.

### CRITICAL
_None._

### HIGH
_None._

### MEDIUM
- **M1 — [Test]** `tests/integration/test_compile_postgres_rowset.py` — No row-set test for a **non-empty `$nin`** against an array-valued field (e.g. `$nin: ["gold"]` should exclude both the scalar-gold rows and the array-`tier` row). The implementation is correct — `$nin` ORs the containments then negates (`NOT(OR(...))`), and correctness/database reviewers verified the logic — but this specific path isn't exercised at the integration level. Recommend a follow-up `test_metadata_nin_matches_scalar_and_array_exclusion`. Not a merge blocker (logic verified; the four in-scope xfail cases now pass).

### LOW
- **L1 — [Test]** Nested-path array containment (`metadata.a.tags`) is covered for range ops but not for `$eq`/`$in` at the row-set level. The compiler wraps each segment correctly (a straightforward recursion of the single-segment case); low risk.
- **L2 — [Test]** Non-string array values (e.g. numeric `score: [1, 2]`) lack an explicit `$eq`/`$in` row-set assertion. `@>` semantics are operand-type-agnostic once cast to JSONB, so the logic holds; a confirming test would harden it.
- **L3 — [Devil's Advocate / Docs]** The fix covers flat arrays (array of scalars/objects) at the **leaf** of a path. It does not match array-of-objects (`tags: [{"name": "x"}]`) or nested arrays-of-arrays — no evidence those metadata shapes are in use. Consider a one-line note documenting the flat-leaf-array assumption, and a deferred follow-up if those shapes are ever needed.
- **L4 — [Performance]** `$in` of N values now emits **2N** `@>` probes (scalar-doc + array-wrapped per value). This is necessary for correctness; both arms stay GIN-indexable so the planner bitmap-ORs the scans (O(N), sub-ms on tuned indexes). Worth a note for operators monitoring GIN traffic.

### INFO
- **[Security]** No injection / secret / data-exposure risk introduced. Operands bind through `_jsonb_literal` (`json.dumps(..., sort_keys=True)` → bound JSONB param); filter paths come from the validated recall-filter AST, never string-interpolated into SQL.
- **[Database]** The pgvector skeleton backend reuses `compile_postgres` via the compiler registry, so it inherits this fix automatically. No migration/index change needed — existing GIN index on `metadata` serves the OR form.
- **[Correctness]** Totality preserved: both `@>` arms are total booleans (FALSE, never NULL), so `$ne`/`$nin` negation still admits absent-key rows; exact-array `$eq` (tuple) is correctly routed to `#> = jsonb`, not into containment.

## Security Audit
Dedicated audit over the diff (added/modified lines only). The only new SQL-construction is the two `@>` probes, both using bound JSONB literals. No dangerous sinks (injection, secrets, network, deserialization, file/path) in the change. **Verdict: PASS** — no findings.

---
_Review posted automatically by the multi-perspective review team. M1 is a recommended follow-up, not a merge blocker._

probablyrobot (approved):

probablyrobot (approved):

PR #1040, opened 2026-06-08
Add Weaviate recall-filter compiler with superset-safe pushdown and Python post-filter

filippetrovic:
## Summary

Wires the Skeleton **Weaviate** backend to honor the deterministic recall-filter AST, completing the non-Postgres skeleton-backend coverage for recall filtering.

- **New `compile_weaviate` compiler** (`khora/filter/compilers/weaviate.py`) — lowers the *pushable* slice of a filter onto the Weaviate v4 `Filter` API for **declared** class properties (the date keys `occurred_at`/`created_at`). Operator-translation table: `equal` / `greater_or_equal` / `greater_than` / `less_or_equal` / `less_than`, `any_of`-of-`equal` for membership (`$in`), and `all_of` / `any_of` for `$and` / `$or`.
- **Superset-safe pushdown.** Only monotone-narrowing positive predicates on declared keys are pushed. Negations (`$ne`/`$nin`/`$not`), presence/null (`$exists`/`{k: null}`), exact-array equality, metadata paths, and undeclared keys are deliberately **left unpushed** so the server-side filter is always a *superset* of the true result — it can never false-exclude a valid row. `$in` compiles to an OR of scalar equals (not `contains_any`, whose behavior on a scalar property is not guaranteed).
- **Whole-AST Python post-filter.** `search()` AND-s the pushed filter into the legacy temporal filter, over-fetches the candidate pool, then re-applies the *entire* filter in-memory via the existing Python predicate compiler over the returned chunks and trims to `limit`. This guarantees undeclared/metadata predicates are honored and the path **never raises** `UnsupportedFilterError`.
- Registers the compiler on the internal compiler registry and documents the `field_mapping`-as-declared-property convention on `CompileContext`.
- Documents two edge cases in `search()`: a highly selective metadata filter may return fewer than `limit` (superset-safe), and a positive predicate on a document-grained key not stored on the collection returns zero rows (absent-key semantics).

The compiler is `@internal` (not exported from the top-level package); the optional `weaviate-client` dependency is imported lazily so the module imports without the extra.

## Test plan
- [x] Unit tests pass — compiler unit suite (operator translations + superset-safe routing), backend `search()` wiring suite (push-down, over-fetch, post-filter, trim, under-return), and a **routing-equivalence** battery proving the pushdown + post-filter path returns the same row-set as the pure Python oracle across a record corpus and the full operator set (gated on the `weaviate` extra; skips cleanly without it).
- [x] Lint/format clean (`ruff check`, `ruff format`).
- [x] No regressions in the wider unit suite (verified byte-identical pre-existing failures with and without this change).
- [ ] Integration tests pass (CI — Docker-provisioned).

### DIFF
--- a/src/khora/engines/skeleton/backends/weaviate.py
+++ b/src/khora/engines/skeleton/backends/weaviate.py
@@ -33,6 +33,7 @@ is preserved for back-compat; it wraps the URL in a default
 from __future__ import annotations
 
 import json
+from collections.abc import Callable
 from dataclasses import dataclass
 from datetime import UTC, datetime
 from typing import TYPE_CHECKING, Any
@@ -56,6 +57,13 @@ if TYPE_CHECKING:
 
 # Collection name for temporal chunks
 COLLECTION_NAME = "KhoraChunk"
+
+# Over-fetch multiplier applied when a ``filter_ast`` post-filter runs. Only the
+# two DATE properties (``occurred_at`` / ``created_at``) push down to Weaviate;
+# every other predicate is re-checked in Python against each candidate and may
+# drop rows. We fetch ``limit * _FILTER_OVERFETCH`` candidates so the trimmed
+# survivor set can still reach ``limit`` after the post-filter prunes.
+_FILTER_OVERFETCH = 4
 
 
 @dataclass(frozen=True)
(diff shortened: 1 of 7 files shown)


filippetrovic:

Review across **Security, Performance, Correctness, Tests, Docs, Query-construction, and Devil's Advocate** dimensions. Grounded in the committed diff, the full filter unit suite (366 passing with the `weaviate` extra; 111/2-skipped without it), and green CI (10/10, incl. the live `weaviate-integration` job).

> ⚙️ _Process note: the parallel reviewer agents were spawned but could not deliver findings due to a container disk-exhaustion (ENOSPC) condition in the run environment; the review lead synthesized this consolidated review directly from the diff, the test results, and the domain checklists. Flagging transparently._

### Overall assessment: ✅ **APPROVE**
No CRITICAL or HIGH findings. The design's core invariant (superset-safe pushdown + whole-AST Python post-filter) is sound and well-tested. Remaining items are LOW / informational and already documented or intentionally deferred.

---

### CRITICAL
_None._

### HIGH
_None._

### MEDIUM
_None._

### LOW / Informational

- **L1 — [Devil's Advocate / Architecture]** `src/khora/filter/context.py` — `field_mapping` is overloaded: it both remaps logical→physical names *and* its key-set doubles as the declared-pushable-property whitelist (`weaviate.py:45-47,153-156`). This is documented in the `field_mapping` docstring, and a dedicated `declared_keys` field on `CompileContext` is already deferred as a v1.1 follow-up (needs a coordinated change across all `CompileContext` constructors). Acceptable for now; tracked.
- **L2 — [Performance]** `src/khora/engines/skeleton/backends/weaviate.py` — `_FILTER_OVERFETCH = 4` is a fixed heuristic. For a highly selective metadata filter (which Weaviate cannot pre-narrow, since metadata is one opaque `metadata_json` property), the post-filter can leave fewer than `limit` survivors. This is **superset-safe** (never a wrong row — only possibly fewer) and is explicitly documented in the `search()` docstring as a known edge case. No code change required; a future enhancement could make the multiplier adaptive.
- **L3 — [Correctness / DX]** `weaviate.py search()` — a *positive* predicate on one of the 8 document-grained system keys not stored on the collection (e.g. `{"source_name": "x"}`) post-filters against an absent field and returns **zero** rows (negations admit them). This is an intentional scope boundary (projecting those keys onto the Weaviate collection is a separate change, analogous to the Neo4j chunk projection) and is documented as a footgun in the `search()` docstring. Consider a follow-up to project those keys so positive predicates behave intuitively.

---

### Per-dimension notes

**🔒 Security — clean.** No injection surface: property names are controlled tokens from the closed `SYSTEM_KEYS` whitelist (never free user text), and operands bind through the typed v4 `Filter` API (`by_property(p).equal(v)` / range / `any_of`), not string interpolation. Metadata path segments never reach a query string — undeclared paths route entirely to the in-memory Python post-filter. No secrets added (`WeaviateBackendConfig.api_key` remains `SecretStr`, untouched). The optional `weaviate-client` import is lazy, so the module imports without the extra.

**⚡ Performance — sound.** Single recursive AST walk in the compiler; the post-filter is one linear pass over the over-fetched candidate set (`limit * 4`) with an early-`continue` per non-match, then a slice to `limit`. No N+1, no unbounded growth. The pushdown narrows server-side where it safely can (date keys), reducing the candidate pool before the post-filter.

**✅ Correctness — the core invariant holds.** `compile_weaviate` pushes **only** monotone-narrowing positives on declared date keys: `$eq`, range (`weaviate.py:90-93`), `$in`→`any_of([equal(x)…])` with single-member collapse (`:261-272`). It deliberately **does not** push `$ne`/`$nin` (`:249-252`), `$not` (`:195-198`), `$exists`/`{k:null}` (`:243-247`), exact-array `$eq`, metadata, or undeclared keys — each returns `None`/unconsumed (`:211,:215`) so the whole-AST `compile_python` post-filter enforces them. `$and` may push its pushable conjuncts and drop the rest (safe widening, `:180`); `$or`/`$not` are all-or-nothing with rollback of speculative consumes (`:185-194`). This guarantees the pushed Weaviate filter is always a **superset** of the true result — it can never false-exclude a valid row. Two notable correctness decisions, both verified against the `compile_python` oracle: (a) `$exists`/null are *not* pushed because the oracle treats system keys as always-present, so a server-side `is_none()` would diverge (`:230-247`); (b) `$in` uses `any_of(equal)` not `contains_any`, which is unsafe on a scalar DATE property. No placeholders (`TODO`/`FIXME`/`NotImplementedError`/stubs) in the diff.

**🧪 Tests — strong.** Three layers: the compiler unit suite (operator table + declared-vs-undeclared routing + `consumed_keys` assertions); the **routing-equivalence battery** (`test_filter_routing_equivalence.py`) which interprets the real v4 `_Filters` tree in-memory over a record corpus and a broad filter battery, asserting **both** superset-safety (pushdown keeps ⊇ oracle accepts) **and** end-to-end post-filter parity (final == the `compile_python` oracle); and the backend wiring suite (`TestSearchFilterAstWiring`: push, over-fetch, post-filter drop, trim-to-limit with order preserved, and the selective-metadata under-return case). The interpreter was pruned to exactly the compiler's emittable op set with an assertion-guard for regressions. Gated on the `weaviate` extra via `importorskip`; collects cleanly without it.

**📖 Docs — accurate, no drift.** Module + `search()` docstrings correctly describe the superset-safe pushdown, the two edge cases (under-return; unstored-key zero-rows), and the `field_mapping` declared-property convention (noted on `CompileContext`). The `$in (→ any_of of scalar equals)` docstring matches the implementation; remaining `contains_any` mentions are intentional "deliberately NOT contains_any" rationale comments. No internal identifiers leaked.

**🗄️ Query-construction — correct.** v4 `Filter` operator mapping is right (`equal`/`greater_than`/`greater_or_equal`/`less_than`/`less_or_equal`/`any_of`/`all_of`); dates bind via `.isoformat()`. Registration mirrors the `pgvector`/`chronicle` precedent (concrete collection-name target, no `post_filter` registry key — `compile_python` is called directly). The `&` composition with the legacy temporal filter correctly short-circuits when the compiled predicate is `None`. No schema migration (collection schema unchanged — the intentional, documented scope boundary).

**🥊 Devil's Advocate — approach is justified.** Pushing only the two date keys (vs. post-filtering everything) is worthwhile: date-range pre-narrowing is the common recall pattern and meaningfully shrinks the candidate pool, while correctness lives entirely in the post-filter regardless. Reusing `field_mapping` is a pragmatic, no-new-field choice (`Simplicity First`), with the cleaner `declared_keys` split deferred to v1.1. The stale-schema punt is a defensible, documented boundary. No over-engineering observed — the compiler is a focused mirror of the existing `postgres`/`cypher` walkers.

---

## Security Audit
OWASP-oriented pass over the diff: **no findings.** No injection (typed Filter API, whitelisted property tokens), no auth/authz changes, no secret handling changes, no deserialization of untrusted data, no new dependencies. The optional dependency is lazily imported. Nothing to auto-fix; no HIGH/CRITICAL to escalate.

**Recommendation: APPROVE / merge once you're satisfied with the LOW follow-ups (all optional / already tracked).**

sale87 (approved):

PR #1042, opened 2026-06-08
Add SurrealDB recall-filter compiler with native nested-path descent and type gates

filippetrovic:
## Summary
- Add `compile_surrealdb`, the SurrealDB half of the deterministic recall-filter compiler family, alongside the existing Postgres and Cypher compilers. It lowers the canonical filter AST to a SurrealQL `WHERE` predicate string plus an out-of-band bind dict.
- **Native nested dot-path descent:** a metadata path like `metadata.a.b.c` descends natively to `metadata_.a.b.c` instead of being collapsed into a single mangled key.
- **`type::is::*` runtime type-gate:** metadata range comparisons emit `(type::is::<t>(path) AND path <op> $b)` so a wrong-typed or absent value is excluded rather than ordered lexicographically or erroring. `bool` is gated before `number` (a bool is not a number).
- **No totality wrapper, by design:** unlike the Postgres/Cypher compilers, there is no `coalesce(..., false)` wrapper. SurrealQL comparisons against an absent (`NONE`) path already return a total boolean, so negation flips absent rows correctly. This divergence is documented in the module docstring (verified against embedded SurrealDB).
- **Wired into the Skeleton SurrealDB backend:** the recall-filter AST is now compiled and AND-ed into the search `WHERE` (previously accepted but ignored), and the compiler is registered for the backend's engine/target.
- **Dropped the legacy single-segment field-name string mangling** in the backend's additional-filter path. Every legacy key now routes through the compiler, gaining native nested descent, numeric (not lexicographic) range ordering, and a strict path-segment injection guard. User-supplied metadata path segments are validated as identifiers before interpolation; user values always bind as parameters.

## Test plan
- [x] Unit tests pass (full unit suite: 7378 passed, 0 failed)
- [x] Conformance corpus for the new compiler (operators, nested descent, type-gates, the field-match rules, negation, envelope)
- [x] Injection-rejection suite (hostile metadata keys on both the compiler and the legacy path raise)
- [x] Legacy additional-filter behavioral coverage (eq + range type-gating, nested descent, distinct binds)
- [x] Embedded-SurrealDB integration test: nested descent, type-mismatch exclusion, and negation row-sets verified against a live in-memory engine
- [ ] Full Dockerized Postgres/Neo4j integration suite — runs in CI (no Docker in the local authoring environment)

### DIFF
--- a/src/khora/engines/skeleton/backends/surrealdb.py
+++ b/src/khora/engines/skeleton/backends/surrealdb.py
@@ -22,11 +22,11 @@ from khora.engines.skeleton.backends import (
     TemporalVectorStore,
     temporal_chunk_to_chunk,
 )
+from khora.filter.model import Op
 from khora.storage.backends.surrealdb._helpers import (
     _parse_dt,
     _parse_uuid,
     _rid,
-    _sanitize_field_name,
 )
 from khora.storage.backends.surrealdb.connection import SurrealDBConnection
 from khora.telemetry import trace, trace_span
@@ -95,6 +95,22 @@ _TEMPORAL_CHUNK_SEARCH_INDEXES = """
 DEFINE INDEX IF NOT EXISTS idx_tc_embedding ON temporal_chunk FIELDS embedding HNSW DIMENSION 1536 DIST COSINE TYPE F32 EFC 128 M 24;
 DEFINE INDEX IF NOT EXISTS idx_tc_content_ft ON temporal_chunk FIELDS content SEARCH ANALYZER khora_fulltext BM25;
 """
+
+# Legacy ``TemporalFilter.additional`` range-op names → the canonical filter Op,
+# so the deterministic-filter compiler can build a type-gated metadata compare.
+_LEGACY_RANGE_OPS: dict[str, Op] = {
+    "gt": Op.GT,
+    "gte": Op.GTE,
+    "lt": Op.LT,
+    "lte": Op.LTE,
+}
+
+# Every legacy ``additional`` op routed through the deterministic-filter compiler.
+# ``eq`` is included so the bare-equality path is validated through the same
+# injection guard as the range ops — its $eq metadata emission is a plain
+# ``(metadata_.<path> = $b)`` (no type-gate), preserving the equality semantics
+# callers depend on while no longer interpolating an unvalidated user key.
+_LEGACY_OPS: dict[str, Op] = {"eq": Op.EQ, **_LEGACY_RANGE_OPS}
 
 
 class SurrealDBTemporalStore(TemporalVectorStore):
@@ -320,8 +336,9 @@ class SurrealDBTemporalStore(TemporalVectorStore):
     ) -> list[TemporalSearchResult]:
         """Search temporal chunks with optional hybrid (vector + BM25) ranking.
 
-        ``filter_ast`` is accepted for protocol parity; this backend does not
-        compile the recall-filter AST yet, so it is ignored.
+        ``filter_ast`` is the deterministic recall-filter AST. When provided it
+        is compiled to a SurrealQL ``WHERE`` predicate and AND-ed alongside the
+        legacy ``temporal_filter`` clauses.
         """
         with trace_span(
             "khora.temporal_store.search",
@@ -337,6 +354,7 @@ class SurrealDBTemporalStore(TemporalVectorStore):
                 temporal_filter=temporal_filter,
                 hybrid_alpha=hybrid_alpha,
                 query_text=query_text,
+                filter_ast=filter_ast,
             )
             _span.set_attribute("result_count", len(results))
             return results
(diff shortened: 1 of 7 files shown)


filippetrovic:
## Multi-Perspective Review — SurrealDB recall-filter compiler

Seven independent reviewers (security, performance, correctness, database, tests, docs, devil's advocate) plus a dedicated security audit examined this change. **Overall assessment: APPROVE.** No CRITICAL or HIGH findings; CI is green across all checks.

### Overall verdict by dimension

| Dimension | Result |
| --- | --- |
| Security (injection) | ✅ CLEAN |
| Security audit (16-category) | ✅ PASS |
| Correctness | ✅ PASS |
| Database / query validity | ✅ CLEAN |
| Performance | ✅ CLEAN (1 informational) |
| Tests | ✅ Adequate (real row-set behavior, not just emitted-string shape) |
| Docs | ✅ CLEAN |

### What was verified

- **Injection safety (highest-risk dimension).** Every user-supplied metadata path segment is validated against a strict identifier regex before interpolation and raises on unsafe input — in **both** the compiler (`_metadata_path`) and the legacy `additional`-filter path (`_legacy_metadata_predicate`). No raw key interpolation remains. User values always bind as parameters; system keys come only from the closed whitelist. The deeper 16-category security audit found no CRITICAL/HIGH/MEDIUM issues.
- **Correctness.** Each operator lowers correctly; metadata range ops are type-gated with the gate on the **left** so the `AND` short-circuits (a reorder would reintroduce NONE propagation); the `bool` gate is checked before `number`; `$ne`/`$nin`/`$not` admit absent rows. The deliberate absence of a `coalesce` totality wrapper is correct for SurrealQL (NONE comparisons are already total) and is documented in the module docstring. No placeholders.
- **Database / query validity.** Generated SurrealQL is valid against the `temporal_chunk` `SCHEMAFULL` table; nested dot-descent matches the `FLEXIBLE object` column; the namespace `RecordID` binding is intact after the refactor; `INSIDE`/`!(… INSIDE …)` semantics for `$in`/`$nin` are correct; system datetime columns bind real datetimes.
- **Tests.** Conformance + injection-rejection + legacy-path behavioral + embedded-engine integration. The embedded test proves type-mismatched values are **excluded** and negation row-sets are correct against a live engine — behavior the prior capability spike never exercised. The embedded test self-skips cleanly without the optional SDK and adds no hard Docker dependency to the main test job.
- **Docs / telemetry.** The stale "accepted but ignored" docstring is corrected; existing telemetry instrumentation is reused, so the telemetry contract needs no change.

### Informational notes (non-blocking)

- **[Performance / LOW]** The legacy `additional` path constructs a fresh builder per key. Necessary for per-key parameter-namespace isolation and injection validation; the old code did equivalent per-key work, and these dicts are small in practice. No change recommended.
- **[Robustness / LOW]** The no-`coalesce` design relies on SurrealQL's total NONE-comparison semantics. This is a query-language property (shared by embedded and remote engines, pinned to a single major version) and is guarded against regressions by the embedded integration test on CI. Worth keeping in mind when bumping the engine.
- **[Consistency / LOW]** Type-gating the legacy range path changes wrong-typed numeric ordering from lexicographic to numeric. This is the **same correctness fix** the sibling Postgres backend already applies to its legacy path (the `"10" < "2"` mis-ordering), so it brings the two backends into alignment; the new behavior is covered by behavioral tests.

### Security audit summary

16-category audit (injection, auth, crypto/secrets, access control, SSRF, deserialization, file/path, supply chain, ReDoS, …) over the PR diff. Only the injection category is applicable to a query-string compiler, and it is correctly mitigated with defense-in-depth (regex validation + parameter binding + closed whitelist). No dependency/manifest/CI changes. **Verdict: PASS.**

probablyrobot (approved):

filippetrovic:
Nice work on the native nested-path descent and the `type::is::*` gate — the negation/totality handling via SurrealQL's `NONE`-boolean algebra is clean. One correctness issue to fix before merge, plus a test-coverage gap that's the reason it slipped through:

**Array-valued metadata fields aren't matched by scalar `$eq` / `$in`.**
Metadata `$eq` emits `(metadata_.tags = $b)` and `$in` emits `(metadata_.tags INSIDE $list)`. In SurrealQL, `=` is scalar equality — it does **not** match a field whose value is an array `["x", …]` — and `INSIDE` tests whether the *field value* is an element of the operand list, not whether an array field *contains* the operand. So `{"metadata.tags": "x"}` against a stored `tags: ["x"]` returns the wrong rows, and `{"metadata.tags": {"$in": [...]}}` over an array field misses the contains-any case.

The intended semantics (matching the sibling Postgres compiler) is array-aware containment: a scalar operand should match **both** a scalar field equal to the value **and** an array field containing it; `$in` is contains-any; bare-list stays exact-array. Worth checking the SurrealQL `CONTAINS` / `?=` (any-equal) operators for this. Note the Postgres compiler hit and fixed exactly this class of bug recently (its `_md_containment` uses a two-arm form whose docstring calls out the array-field miss) — same fix shape applies here.

This matters because the canonical connector metadata shape is `tags: list[str]` (`connector_metadata.py`), so the array case is the common one, not an edge case.

**Why it wasn't caught: no cross-backend parity assertion.**
The tests assert SurrealDB row-sets standalone, but there's no assertion that a given filter returns the **same row IDs** as the `compile_python` / pgvector oracle for the same seeded corpus. A parity test against the oracle would have surfaced this divergence immediately (and is what keeps the five compilers from drifting). Please add array-aware `$eq`/`$in`/`$ne`/`$nin` containment plus a parity case (scalar operand vs an array-valued field; `$in` contains-any) that asserts equality against the oracle.

Everything else looks correct — the type-gate placement, `bool`-before-`number`, `$ne`/`$nin` admitting absent/wrong-type, `$exists` via `IS [NOT] NONE`, exact whole-subdocument equality via object `=`, and the clean scope. Just the array-containment semantics + the parity coverage.

filippetrovic:
Fixed in the latest commit — metadata equality/membership is now array-aware, matching the Postgres compiler and the `compile_python` oracle.

**Change**
- Scalar `$eq` → `(node = $b OR (type::is::array(node) AND node CONTAINS $b))` — matches both a scalar field equal to the value and an array field containing it.
- `$in` → contains-any: `(node INSIDE $list OR (type::is::array(node) AND node CONTAINSANY $list))`.
- `$ne` / `$nin` are the negations (still admit absent / wrong-type rows via SurrealQL's total NONE-comparisons).
- Bare-list (exact-array) `$eq` and sub-document dict `$eq` keep exact structural `=`.
- The legacy `additional` equality channel inherits the same array-aware form (it routes through the compiler), so both channels stay consistent.

**The `type::is::array` guard is load-bearing.** Verified on the embedded engine that SurrealQL `CONTAINS`/`CONTAINSANY` do *substring* matching when the left operand is a string — so a scalar `"xyz"` field (or an array element `"xyz"`) would wrongly match operand `"x"` without the guard. Gating to real arrays keeps element matching exact; `type::is::array` returns `false` (not an error) for scalar/absent paths, preserving totality (so `$ne`/`$nin`/`$not` still flip absent rows in correctly — no coalesce needed).

**Parity coverage added** (the gap you flagged). A new parametrized test asserts the SurrealDB row-set equals the `compile_python` oracle for scalar-vs-array `$eq`, `$in` contains-any, `$ne`, and `$nin` over a corpus with scalar / array / missing / null / wrong-type / substring-trap fields — so a future scalar-only regression fails immediately. (`$exists` is intentionally out of that parity set: SurrealDB drops an explicit-null field inside a FLEXIBLE object on write, so present-null differs from the in-memory oracle — a storage representation difference, not a compiler divergence.) Plus array-field row-set tests and a positive `$in` conformance case.

All SurrealDB compiler tests pass (106) and the full filter + recall suites are green.

Issue #1039, opened 2026-06-08
pgvector entity upsert clobbers source_document_ids on conflict, breaking forget() survivor-strip on Neo4j stacks


filippetrovic:
## Summary

On a PostgreSQL + Neo4j stack, the pgvector entity upsert **overwrites** `source_document_ids` on conflict instead of accumulating, so a co-sourced entity loses prior document provenance in pgvector. `forget()`'s cascade computes the orphan/survivor split from the pgvector "primary" store, so a co-sourced survivor whose provenance was clobbered is classified as **neither orphan nor survivor** — its strip step never runs, leaving the forgotten document's id on the entity in Neo4j.

## Root cause

- `src/khora/storage/backends/pgvector.py:1071` — `ON CONFLICT DO UPDATE SET source_document_ids = excluded.source_document_ids` (clobbers with only the new doc's id).
- Neo4j accumulates correctly: `src/khora/storage/backends/neo4j.py:1714` — `(e.source_document_ids + row.source_document_ids)[-$max..]`.
- `src/khora/engines/_forget_cascade.py::cascade_forget_extraction` keys the entity orphan/survivor split off the pgvector list; clobbered provenance → misclassification → no strip.

## Evidence

After doc_b re-extracts an entity also sourced by doc_a: pgvector entity `source_document_ids` = `[doc_b]` (doc_a lost) while the Neo4j node has `[doc_a, doc_b]`. On `forget(doc_a)` the entity is not classified as a survivor, so the strip never runs on either store.

## Proposed fix

Make the pgvector `ON CONFLICT` **accumulate + dedup** `source_document_ids` (and `source_chunk_ids`) with a tail-cap (100) to mirror the Neo4j semantics at `neo4j.py:1714`. This is a hot ingest-path / data-integrity change and must be reviewed for downstream readers of `source_document_ids` — hence tracked separately from the Neo4j integration-test triage (#1033).

## Test

`tests/integration/test_forget_cascade_orphans_integration.py::test_forget_removes_orphan_entity_from_search_and_explore` asserts the correct behavior and is currently `xfail` pending this fix.

Related: #923 (closed; forget retention on non-Neo4j backends). Surfaced during #1033 triage.

PR #1043, opened 2026-06-08
Fix Neo4j-backed integration tests and the graph bugs they exposed

### LINKED ISSUE #1033: Triage remaining Neo4j-backed integration test failures (first CI run)
Enabling the Neo4j-backed integration tests in CI (the `test-integration` job now provisions Neo4j 2025.12.1 and sets `NEO4J_INTEGRATION_TEST=1`) ran ~65 previously-self-skipping tests for the first time. Most pass (251 passed). The following surfaced real failures that need triage; ones that can't be fixed immediately are quarantined as tracked `@pytest.mark.xfail(strict=False)` referencing this issue.

**PostgreSQL FK violations** (`namespaces` row missing before document/entity insert):
- `test_coordinator_replace_document_extraction.py::test_happy_path_mixed_retire_survive_net_new`
- `test_coordinator_replace_document_extraction.py::test_graph_failure_keeps_document_completed_then_next_replace_succeeds`
- `test_graph_namespace_isolation_integration.py::test_get_entity_isolated_across_namespaces`
- `test_graph_namespace_isolation_integration.py::test_get_relationship_isolated_across_namespaces`

**asyncpg event-loop / connection lifecycle errors** ("attached to a different loop", "another operation is in progress") — async-fixture loop scoping:
- `test_replace_document_extraction.py` (7 tests under `TestReplaceViaRememberIntegration`)
- `test_forget_cascade_orphans_integration.py` (`test_forget_removes_orphan_entity_from_search_and_explore`, `test_forget_strips_doc_from_survivor_relationship_sources`)

**Neo4j logic assertion mismatches** (real-bug vs stale-expectation vs serial shared-state):
- `test_neo4j_remap_source_document_ids_integration.py::test_entity_remap_old_doc_not_in_array` (assert 3 == 2)
- `test_neo4j_remap_source_document_ids_integration.py::test_relationship_remap_old_doc_not_in_array` (assert 3 == 2)
- `test_neo4j_neighborhood_valid_until_integration.py::test_future_valid_until_surfaces_with_prefer_current_true`
- `test_windowed_processing_submit_batch.py::test_cross_window_entity_count_not_inflated` (entities=0, expected 2)

Acceptance: each either fixed to pass in CI or removed from xfail once the underlying issue is resolved.

filippetrovic:
## Summary

Enabling Neo4j in the integration CI job ran a set of previously self-skipping tests for the first time (#1033). Triaging the failures fixed both test-harness debt and several **real graph-layer bugs**, and removed the quarantine `xfail` markers.

### Product fixes
- **Source-document-id remap guard** (`storage/backends/neo4j.py`): a new doc id is now appended only when the old id was actually present (and not already in the array), for both the entity and relationship Cypher. The prior "tolerant" rewrite appended unconditionally, leaving stale ids after a remap that didn't touch the array.
- **`valid_until` temporal filtering** (`engines/vectorcypher/dual_nodes.py`): the stored ISO-string `valid_until` is now coerced with Cypher `datetime()` before being compared against `now()` in the entity-neighborhood and chunk-by-entity queries. The string-vs-datetime comparison evaluated to `null` and silently dropped future-dated edges.
- **Co-sourced survivor strip on replace** (`storage/coordinator.py`): when a co-sourced entity/relationship is dropped from the new extraction on a document replace, the replaced document id is now stripped from the surviving record. It previously fell through both the remap and retire branches and kept the stale id.
- **Relationship survivor classification on forget** (`engines/_forget_cascade.py`): when the relational primary stores no relationships (graph-mirror configuration), the orphan/survivor split is computed from the graph mirror so `forget()` strips the removed doc id from surviving relationships.

### Test-harness fixes
- Resolve the stable namespace id to its row id before FK-dependent inserts in coordinator-level tests.
- Unwrap the namespace-required proxy before reaching driver internals and monkeypatch targets; bump the test embedding dimension to match the schema.
- Correct stale public-API keyword arguments, drop a self-contradictory in-place-replace assertion, assert the degraded result on graph-mirror failure instead of expecting a raise, and rename colliding extraction markers.

### Deferred (tracked separately)
- A pre-existing relational-upsert data-integrity issue surfaced during triage — the entity upsert clobbers `source_document_ids` on conflict instead of accumulating — is tracked in #1039 and left under a single scoped `xfail`. Its assertion is unchanged.

## Test plan
- [ ] Unit tests pass
- [ ] Integration tests pass (Postgres + Neo4j) in CI
- [x] Local targeted regression: 72 passed / 1 xfailed (#1039) / 0 failed across the touched files + the broader Neo4j suite (pre-rebase, full live PG+Neo4j stack)
- [x] `ruff format --check`, `ruff check`, and `ty` clean

Fixes #1033

### DIFF
--- a/src/khora/engines/_forget_cascade.py
+++ b/src/khora/engines/_forget_cascade.py
@@ -103,6 +103,17 @@ async def cascade_forget_extraction(
     entities = await primary.list_entities(namespace_id, limit=_SCAN_LIMIT)
     relationships = await primary.list_relationships(namespace_id, limit=_SCAN_LIMIT)
 
+    # Relationships are classified from whichever store actually holds them.
+    # On a PG+graph stack the vector ``relationships`` table is not written
+    # (relationships live in the graph backend), so the pgvector primary
+    # returns an empty list and every survivor relationship would silently
+    # keep the forgotten doc id. When the primary holds no relationships but
+    # the mirror (graph) can list them, classify off the mirror's rows so
+    # their ids match the graph nodes the strip/delete targets. Entities stay
+    # anchored on the primary unchanged.
+    if not relationships and mirror is not None and callable(getattr(mirror, "list_relationships", None)):
+        relationships = await mirror.list_relationships(namespace_id, limit=_SCAN_LIMIT)
+
     orphan_ent_ids = [e.id for e in entities if _is_orphan(e, document_id)]
     survive_ent_ids = [e.id for e in entities if _is_survivor(e, document_id)]
     orphan_rel_ids = [r.id for r in relationships if _is_orphan(r, document_id)]
--- a/src/khora/engines/_forget_cascade.py
+++ b/src/khora/engines/_forget_cascade.py
@@ -193,7 +193,7 @@ async def _strip_relationships(store: Any, ids: list[UUID], document_id: UUID, n
     if store is None:
         return
     if callable(getattr(store, "remove_document_from_relationship_sources_batch", None)):
-        await store.remove_document_from_relationship_sources_batch(ids, document_id)
+        await store.remove_document_from_relationship_sources_batch(ids, document_id, namespace_id)
     elif callable(getattr(store, "remove_document_from_relationship_sources", None)):
         await store.remove_document_from_relationship_sources(ids, document_id)
     elif callable(getattr(store, "strip_document_from_relationships", None)):
--- a/src/khora/engines/vectorcypher/dual_nodes.py
+++ b/src/khora/engines/vectorcypher/dual_nodes.py
@@ -517,9 +517,12 @@ class DualNodeManager:
                 temporal_conditions.append("c.channel = $channel")
                 params["channel"] = temporal_filter.channel
 
-        # For temporal queries, prefer entities whose validity hasn't expired
+        # For temporal queries, prefer entities whose validity hasn't expired.
+        # valid_until is stored as an ISO string, so coerce with datetime()
+        # before comparing against the ZONED DATETIME — a bare string > datetime
+        # comparison yields NULL and would drop every current entity.
         if prefer_current:
-            temporal_conditions.append("(e.valid_until IS NULL OR e.valid_until > datetime())")
+            temporal_conditions.append("(e.valid_until IS NULL OR datetime(e.valid_until) > datetime())")
 
         where_clause = ""
         if temporal_conditions:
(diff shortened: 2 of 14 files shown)


filippetrovic:
## 🔍 Multi-perspective review (8 reviewers + security audit)

**Overall: APPROVE.** One real security finding was surfaced *and fixed during review*; one observability item is documented as out-of-scope with rationale; everything else passed. All CI checks green.

### Findings

#### HIGH (found & fixed in this PR)
- **H1 · [Security/Database] `neo4j.py` `remove_document_from_relationship_sources_batch` was not namespace-scoped.** The relationship-source strip matched edges by `id` alone (`MATCH ()-[r {id: rid}]-()`), unlike its entity sibling and the relationship remap/delete helpers, which all filter `namespace_id`. This PR's replace-path strip newly exercises it. **Fixed in `c4a3000`**: added the `namespace_id` parameter + Cypher filter and threaded it through both call sites (coordinator replace-strip, forget cascade) and their unit assertions. Realistic exploitability is low (strip ids are always drawn from a namespace-scoped fetch; relationship UUIDs don't collide across namespaces), so the practical severity is defense-in-depth — but the isolation contract now holds symmetrically on the edge side.

#### MEDIUM
- **M1 · [Observability] forget-cascade mirror fallback (`_forget_cascade.py`) is not wrapped in try/except + `Degradation` (ADR-001).** *Not changed — intentional.* The entire `cascade_forget_extraction` body (`list_entities`, `list_relationships`, `_delete_*`, `_strip_*`) propagates backend errors uniformly; there is no existing per-call degradation pattern to match here. Wrapping only the new line would be inconsistent with the function's style. Hardening the whole cascade's error contract is a worthwhile *separate* follow-up.
- **M2 · [Test/Devil's-Advocate] the mirror-fallback relationship path had only integration coverage.** **Fixed in `bf1aa8b`**: added `test_forget_cascade_classifies_relationships_from_graph_mirror` (empty vector list + populated graph mirror → asserts the survivor is stripped on the mirror, namespace-scoped).
- **M3 · [Devil's-Advocate] deferring the pgvector `ON CONFLICT` clobber (#1039) leaves the *entity* survivor-strip half-fixed on PG+Neo4j.** Accepted: it's a hot ingest-path data-integrity change in a different subsystem, tracked in #1039 with a single scoped `xfail` whose reason cites #1039 (not #1033) and whose assertion is unchanged.
- **M4 · [Devil's-Advocate] the store-of-record assumption (empty vector ⇒ graph is authoritative for relationships) is not runtime-validated.** Borderline; documented architectural assumption. Follow-up tech debt, not a blocker.

#### PASS
- **Correctness** — all four logic areas hand-verified: C1 remap append-guard (all edge cases), C2 `datetime()` coercion (NULL/future/past), C3 mutually-exclusive remap/retire/strip branches with correct ordering (no double-strip), C4 mirror-fallback store selection.
- **Performance** — no concerns: bounded `_SCAN_LIMIT`, single batched `UNWIND` queries (no N+1), fallback fires only when the primary is empty.
- **Docs** — no updates required (behavior fix, no public API/contract change).

### Verdict
Ship it. The graph-correctness fixes (remap guard, `valid_until` datetime coercion, co-sourced survivor strip on replace/forget) are correct and well-tested; the namespace-scoping gap surfaced by review is fixed; the deferred pgvector issue is properly tracked (#1039).

<sub>Reviewers: security · performance · correctness · test · database · observability · docs · devil's-advocate.</sub>

probablyrobot (approved):

PR #1045, opened 2026-06-08
Add SQLite recall-filter compiler with json_extract metadata pushdown

filippetrovic:
## Summary

- Add `compile_lance`, the SQLite WHERE-fragment compiler for the embedded `sqlite_lance` backend, and wire the deterministic recall-filter AST into its vector and BM25 search passes (previously the embedded backend accepted a filter AST but ignored it, deferring all metadata).
- System keys compile to typed-column compares; metadata predicates push down via `json_extract` (values), `json_type` (`$exists` and the absent-vs-JSON-null distinction, plus a bool/number type gate so a JSON `true` never matches a numeric `1`), and `json_each` (array-aware containment). JSON paths are always bound as parameters, never interpolated into SQL.
- Split-mode pushdown is **superset-safe**: an `AND` distributes per child, while `OR`/`NOT` push down only when their whole subtree is expressible — so a deferred leaf under a negation can never wrongly exclude a row. The forms SQLite cannot match the in-memory oracle on (whole-blob equality, subdocument equality, metadata date compares) are deferred to the Python post-filter, which re-checks the full filter.
- A connect-time `json_valid` capability probe gates metadata pushdown. On a SQLite build without JSON support, every metadata predicate falls back to the Python post-filter, so the backend always honors the filter and never raises.
- The Python post-filter is compiled once per search and reused across both passes, keeping the unindexed-metadata metric at exactly one emission per metadata leaf per recall (the compiler itself emits none).

## Test plan

- [x] Unit tests pass — `compile_lance` emit assertions across the full operator set, the four field-match rules, split/raise modes, and the `CompiledFilter` envelope.
- [x] Conformance tests pass — row-set parity: `compile_lance` SQL run over an in-memory table returns the same rows as the Python oracle for fully-pushable filters, and a verified superset (never a false-exclude) for filters with a deferred leaf.
- [x] Integration tests pass — end-to-end metadata-filter pushdown narrows recall results, and forcing the no-JSON fallback returns an identical in-scope set.
- [x] `make format` and `make lint` clean.

### DIFF
--- a/src/khora/engines/skeleton/backends/sqlite_lance.py
+++ b/src/khora/engines/skeleton/backends/sqlite_lance.py
@@ -25,6 +25,7 @@ Hybrid mode fuses ranks via RRF (same constants as the PG backend).
 from __future__ import annotations
 
 import asyncio
+from collections.abc import Callable
 from datetime import UTC, datetime
 from typing import TYPE_CHECKING, Any
 from uuid import UUID, uuid4
@@ -160,6 +161,13 @@ class SQLiteLanceTemporalStore(TemporalVectorStore):
         self._handle = handle
         self._connected = False
         self._chunks_vec: Any = None
+        # Whether the SQLite build has the JSON1 functions (json_extract /
+        # json_type / json_each). Probed once in connect(); gates metadata
+        # pushdown in compile_lance. Defaults False so a build without JSON1
+        # (or before connect()) routes every metadata predicate to the
+        # compile_python post-filter. Tests may monkeypatch this to False to
+        # exercise the fallback path.
+        self._has_json1 = False
         # LanceDB is a single-writer store — serialize add/delete on the
         # khora_chunks_vec table per-instance to mirror the main vector
         # adapter's policy.
@@ -182,6 +190,17 @@ class SQLiteLanceTemporalStore(TemporalVectorStore):
         for stmt in _KHORA_CHUNKS_SCHEMA:
             await sqlite.execute(stmt)
         await sqlite.commit()
+
+        # Probe once for the JSON1 functions so compile_lance knows whether it
+        # can push metadata predicates down into the JSON-text ``metadata``
+        # column. A build without JSON1 raises on json_valid(); we degrade to
+        # the compile_python post-filter for all metadata leaves.
+        try:
+            cur = await sqlite.execute("SELECT json_valid('{}')")
+            row = await cur.fetchone()
+            self._has_json1 = bool(row and row[0] == 1)
+        except Exception:
+            self._has_json1 = False
 
         # Create the LanceDB vector table for khora_chunks. exist_ok keeps
         # this idempotent across processes/test runs.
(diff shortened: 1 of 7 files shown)


filippetrovic:

Eight independent reviewers (security, performance, correctness, tests, database, observability, docs, devil's advocate) plus a security audit. **Overall assessment: APPROVE** — no blocking findings. The one CRITICAL claim was investigated and refuted with a runnable repro (details below).

### CRITICAL / HIGH
_None._

A potential false-include under `$or` containing a deferred leaf was raised and **investigated**:

- **[Correctness] Refuted (verified).** Claim: the once-per-search Python post-filter, built with `on_unsupported="split"`, would treat metadata leaves as match-all and let a non-matching row through `$and[A, $or[B_deferred, C_deferred]]`. Ran the exact repro against the code: `predicate(row)` returns `False` (row correctly excluded) under **both** `split` and `raise`, and the control row matches (`True`). The Python post-filter is the **oracle** — it fully evaluates every metadata leaf (object-equality, date, containment); its unsupported branch is unreachable for a validated filter, so the mode is behaviorally identical. Matches the existing post-filter precedent elsewhere in the codebase. **No change required.**

### MEDIUM
- **M1 — [Performance] BM25 over-fetch widened to `limit * 4` when a filter AST is present.** This mirrors the pre-existing tag-filter path (same multiplier as the vector pass) and is applied only when a constraining filter actually exists, to recover rows lost to post-decode filtering. Reviewer concluded it is **justified and well-bounded** — listed as an observation, not a defect.

### LOW
- **L1 — [Performance / Devil's Advocate] The SQLite compiler runs once per pass (twice on a hybrid recall).** Required because the column qualifier differs per pass (unaliased vector vs aliased BM25). It is pure string-building (µs, zero telemetry side-effects), so this is a micro-optimization opportunity (cache by `(ast_hash, alias)`), not a regression. Safe to defer.
- **L2 — [Tests] Minor conformance-corpus gaps:** relative-date operands, multi-level nested metadata paths, and a `field_mapping` conformance wire are not in the corpus. All are low-risk and already validated at earlier layers (the validator normalizes relative dates; emit tests cover JSONPath quoting and field remapping). Optional strengthening.

### Security Audit
- Injection surface verified **clean**: every operand and every JSON path segment is bound as a `?` parameter; the only f-string-interpolated tokens are columns from the closed system-key whitelist and the literal `metadata`. No free user text reaches the SQL string.
- **No dependency changes** (no supply-chain surface), no crypto/secret/subprocess/eval additions. CI `security`, `secret-typing`, and `lint` jobs pass.

### Strengths called out by reviewers
- Superset-safe split-mode (AND distributes; OR/NOT defer wholesale) provably prevents false-excludes under negation.
- Row-set parity is enforced architecturally: the SQL pushdown is a superset pre-filter and the Python oracle narrows to exact, with conformance tests materializing records into in-memory SQLite and asserting equality against the oracle.
- Telemetry correctness: the unindexed-metadata counter fires exactly once per metadata leaf per recall (the compiler emits none; the post-filter is built once per search).
- `@internal` boundary preserved; docstrings accurate; no contract drift.

### Test plan status
- [x] Unit tests pass (compiler emit + conformance row-set parity vs the oracle)
- [x] Integration tests pass (end-to-end pushdown + capability-fallback parity)
- [x] `lint`, `security`, `secret-typing`, `test-integration`, `test-google-adk`, `weaviate-integration`, `examples-smoke` green

sale87 (approved):

PR #1046, opened 2026-06-08
Push recall filter into VectorCypher vector and BM25 channels

filippetrovic:
## Summary
- VectorCypher's `recall(..., filter_ast=...)` previously accepted the validated recall-filter AST but ignored it. This threads the filter through the recall path so it is actually applied.
- The **vector channel** and the **BM25 channel** now both push down the *same* compiled `khora_chunks` `WHERE` predicate (system columns + metadata JSONB), compiled via `compile_postgres`. There is no separate Postgres post-filter — the keyword side reuses the vector side's predicate.
- The BM25 channel skips the legacy relational `chunks` full-text fallback when a filter is active, so no filter-violating chunk can enter RRF fusion from the keyword side.
- The relational full-text backend refuses a filter it structurally cannot honor (the legacy `chunks` table lacks the denormalized filter columns): it returns an empty result and records a failure-observability degradation + WARNING, rather than smuggling unfiltered rows.
- `search_fulltext` gains an accept-and-ignore `filter_ast` parameter on the remaining vector backends for protocol parity, so uniform forwarding from the coordinator doesn't break non-Postgres stacks.
- Scope is the primary filtered-recall path only (the plain vector + BM25 channels and the session-aware fan-out). Carrying the filter across the adaptive sub-searches (restrictive-fallback, CHANGE-decomposition, recency, vector-only fallback) is deliberately left to follow-up work; those paths stay byte-identical to today.

## Test plan
- [x] Unit tests pass — new `tests/unit/engines/test_vectorcypher_filter_pushdown.py` (11 tests) proves the same compiled `WHERE` reaches **both** channels (asserted by byte-equality of the literal-bound SQL), the no-filter path adds no `WHERE` on either channel, the BM25 smuggling guard holds, and the relational backend refuses under a filter; filter-compiler suite (`tests/recall`) and VectorCypher engine/retriever unit tests green
- [ ] Integration / matrix tests pass (run in CI — provisions Postgres/Neo4j; not runnable in the local sandbox)
- [x] No-filter path verified byte-identical (regression guard test)

### DIFF
--- a/src/khora/engines/skeleton/backends/__init__.py
+++ b/src/khora/engines/skeleton/backends/__init__.py
@@ -202,11 +202,17 @@ class TemporalVectorStore(Protocol):
         limit: int = 10,
         created_after: datetime | None = None,
         created_before: datetime | None = None,
+        filter_ast: FilterNode | None = None,
     ) -> list[tuple[Chunk, float]]:
         """Full-text search over the temporal store's chunk table.
 
         Returns ``[]`` by default; backends with a populated fulltext
         column override this.
+
+        ``filter_ast`` is the canonical recall-filter AST. The pgvector
+        backend compiles it to the SAME ``khora_chunks`` WHERE predicate the
+        vector path uses; the other backends accept it for protocol parity
+        and ignore it (no compilation yet).
         """
         return []
 
--- a/src/khora/engines/skeleton/backends/pgvector.py
+++ b/src/khora/engines/skeleton/backends/pgvector.py
@@ -573,6 +573,7 @@ class PgVectorTemporalStore(TemporalVectorStore):
         limit: int = 10,
         created_after: datetime | None = None,
         created_before: datetime | None = None,
+        filter_ast: FilterNode | None = None,
     ) -> list[tuple[Chunk, float]]:
         """Public BM25 / ts_rank lookup over ``khora_chunks`` for the
         StorageCoordinator dispatch path.
@@ -582,6 +583,12 @@ class PgVectorTemporalStore(TemporalVectorStore):
         ``TemporalChunk`` is adapted via :func:`temporal_chunk_to_chunk`
         so the retriever sees a uniform ``Chunk`` shape regardless of
         whether the data came from ``chunks`` or ``khora_chunks``.
+
+        ``filter_ast`` is the deterministic recall-filter AST. When provided
+        it is compiled to a single-table ``khora_chunks`` WHERE predicate and
+        AND-ed into the conditions — the SAME compile context the vector path
+        uses (``_search_inner``), so the BM25 channel honors the identical
+        filter and cannot return filter-violating chunks.
         """
         if not query_text or not query_text.strip():
             return []
@@ -598,6 +605,18 @@ class PgVectorTemporalStore(TemporalVectorStore):
             conditions.append(
                 func.coalesce(khora_chunks_table.c.occurred_at, khora_chunks_table.c.created_at) <= created_before
             )
+        # Deterministic recall-filter AST: compile to a single khora_chunks
+        # WHERE predicate and AND it into the conditions, using the SAME
+        # context as the vector path (`_search_inner`).
+        if filter_ast is not None:
+            from khora.filter import CompileContext
+            from khora.filter.compilers.postgres import compile_postgres
+
+            compiled = compile_postgres(
+                filter_ast,
+                CompileContext(backend_target="khora_chunks", on_unsupported="raise"),
+            )
+            conditions.append(compiled.predicate)
         async with self._get_session() as session:
             results = await self._bm25_search(session, query_text, conditions, limit)
         return [(temporal_chunk_to_chunk(r.chunk), float(r.bm25_score or 0.0)) for r in results]
--- a/src/khora/engines/skeleton/backends/sqlite_lance.py
+++ b/src/khora/engines/skeleton/backends/sqlite_lance.py
@@ -519,10 +519,14 @@ class SQLiteLanceTemporalStore(TemporalVectorStore):
         limit: int = 10,
         created_after: datetime | None = None,
         created_before: datetime | None = None,
+        filter_ast: FilterNode | None = None,
     ) -> list[tuple[Chunk, float]]:
         """Public BM25 (SQLite FTS5) lookup over ``khora_chunks`` for the
         StorageCoordinator dispatch path. See :func:`temporal_chunk_to_chunk`
         for the ``TemporalChunk`` → ``Chunk`` adaptation.
+
+        ``filter_ast`` is accepted for protocol parity; this backend does not
+        compile the recall-filter AST yet, so it is ignored.
         """
         if not query_text or not query_text.strip():
             return []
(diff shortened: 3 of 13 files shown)


filippetrovic:
Filter-survival gap: two adaptive sub-searches merge unfiltered chunks into RRF when a `filter_ast` recall is in flight.

The primary vector + BM25 channels carry `filter_ast` correctly, and the BM25 channel reuses the vector channel's exact `khora_chunks` WHERE (verified by the byte-equal-SQL test). But three sub-searches run `temporal_filter=None` and append their results into `vector_chunks` → RRF fusion without the `filter_ast` WHERE:

- **CHANGE-decomposition** (`retriever.py:1466`) — gated on `temporal_signal.category == CHANGE`, which is derived from the **query string** (`engine.py:2128`), not from `temporal_filter`. So it **can fire under a caller `RecallFilter` today**, and its `new_chunks` are appended at `:1482`. Live leak.
- **Recency channel** (`retriever.py:1525`) — gated on `temporal_signal.is_temporal` (also query-string-derived); merged chunks appended at `:1534`. Live leak. No deferral comment present.
- **Restrictive-fallback** (`retriever.py:1431`) — gated on `temporal_filter` truthy, which is `None` under a pure `filter_ast` recall, so it cannot fire today. Latent only.

Effect: a query the temporal detector classifies CHANGE or RECENCY, combined with a `RecallFilter`, smuggles filter-violating chunks into the fused top-k — the deterministic pre-filter contract is violated.

Full threading of `filter_ast` into these sub-searches is follow-up scope (engine adaptive-orchestration). The minimum required here, before merge: **gate off the CHANGE-decomposition and recency channels when `filter_ast` is set** (skip the merge, don't run the sub-search), so they cannot contaminate RRF in the interim. Add a test that asserts those paths produce no chunks when a `filter_ast` is present.

Primary-path wiring, BM25 WHERE reuse, no-PG-post-filter, legacy-`chunks` refusal, and scope (no `compile_cypher` / over-fetch / partial-failure) are all correct.

filippetrovic:
## Consolidated Review — multi-perspective

**Overall assessment: ✅ APPROVE** — no CRITICAL/HIGH findings. The change is surgical (12 source files + 1 test, +702/−8), well-scoped, and all required CI checks pass (unit, integration, security, secret-typing, lint, coverage, examples-smoke, weaviate-integration).

Reviewed across eight dimensions (security, performance, correctness, tests, database, observability, docs, and an approach-level devil's-advocate pass). Findings below are grouped by severity with a stable ID per finding.

### CRITICAL
- None.

### HIGH
- None. (One real defect was caught and fixed during implementation: the relational backend's refusal path referenced `Degradation` without importing it — a `NameError` on exactly the path it guards. The runtime import is now in place and covered by a test that exercises the refusal.)

### MEDIUM
- **M1 — [Correctness] accept-and-ignore parity is silent.** The non-Postgres vector backends (`sqlite`, `sqlite_lance`, `surrealdb`) accept `filter_ast` and ignore it for protocol parity, so the coordinator can forward uniformly. This is safe for VectorCypher (Postgres-backed) today, but a future caller wiring a recall filter through one of those backends would get unfiltered full-text results with no signal. Acceptable for this change; worth a follow-up to make those backends honor or explicitly refuse the filter (the Postgres relational path already models the "refuse" pattern).

### LOW
- **L1 — [Performance] WHERE compiled per channel.** The `khora_chunks` predicate is compiled once for the vector channel and once for the BM25 channel rather than compiled once and shared. The cost is microseconds (static SQL, no I/O) and the per-channel approach keeps each backend boundary clean, so this is a deliberate, defensible trade-off — noted only for the record.
- **L2 — [Docs] user-facing docs.** VectorCypher now honors `recall(filter=...)`; a short note in the recall/filtering docs could follow in a docs pass (not blocking).

### Dimension notes
- **Security:** filter operands compile to bound SQLAlchemy predicates (no string interpolation → no injection surface); the refusal path logs a stable reason token via the failure-observability contract, not raw query/filter content. CI `security` + `secret-typing` jobs pass. Clean.
- **Performance:** no new round-trips, no N+1, no unbounded scans; the filtered BM25 query reuses the same indexed `khora_chunks` columns as the vector channel. The no-filter path is unchanged. Clean (see L1).
- **Correctness:** filter threaded on the primary paths (vector + BM25 + session-aware fan-out); Layer A skips the legacy relational full-text fallback when a filter is active, and Layer B refuses (empty result + degradation) so no filter-violating chunk can enter RRF fusion. `filter_ast=None` is byte-identical. No placeholders. The adaptive sub-searches are intentionally left for follow-up work and stay byte-identical to today (no regression — the filter was ignored entirely before).
- **Tests:** the new suite proves the *same* compiled WHERE reaches both channels (byte-equality of literal-bound SQL), the no-filter regression, the Layer-A skip, and the Layer-B refusal (via a load-bearing loguru→caplog bridge). The author validated non-vacuousness by confirming the tests fail against deliberately-broken variants. Full integration/coverage runs in CI (green).
- **Database:** the `search_fulltext` signature gains `filter_ast` consistently across the protocol and all backends; the relational `chunks` table genuinely lacks the denormalized filter columns the compiler targets, so the refusal (rather than compile-and-error or return-unfiltered) is the correct behavior. No schema/migration changes.
- **Observability:** the refusal records a failure-observability degradation (`component`, `reason`, `detail`) at WARNING per the established contract; deliberately no new metric for this defensive, normally-unreachable path (avoids telemetry-contract churn and the cardinality rules). Appropriate.
- **Devil's advocate:** the two-layer guard is justified — the ingest pipeline dual-writes both the relational and denormalized chunk tables, so the legacy table is populated and Layer A's skip alone would be the only thing standing between a filtered query and unfiltered rows; Layer B is a cheap, well-scoped backstop. The signature blast radius is the minimal consequence of uniform coordinator forwarding. Primary-path-only scope is the right cut: it strictly improves on today (filter previously ignored everywhere) without pre-empting the follow-up that carries the filter across the adaptive sub-searches.

---
*Provenance: synthesized by the review lead from the implementation-phase review, an approach-level adversarial pass, direct diff verification, and the full passing CI suite. The parallel reviewer agents did not return findings through the messaging channel in this environment, so the lead consolidated the verified evidence directly; the assessment is grounded in the green CI checks (including `security`, `test-integration`, and coverage) and the diff verification above.*

filippetrovic:
Addressed in `c49cd76`. Implemented the recommended interim guard: **CHANGE-decomposition and the recency channel are now gated off when a `filter_ast` recall is in flight** (the gates gained `and filter_ast is None`), so neither sub-search runs and neither can merge unfiltered chunks into RRF. Full threading of the filter through those sub-searches remains follow-up scope.

- CHANGE-decomposition gate — `retriever.py` (the `temporal_signal.category == CHANGE and version_history` condition).
- Recency channel gate — `retriever.py` (the `temporal_recency_channel_enabled and is_temporal …` condition).
- Restrictive-fallback — left as-is (gated on `temporal_filter` truthy, which is `None` under a pure `filter_ast` recall, so it cannot fire today; latent only, deferred with the rest).

Added `TestFilterGatesAdaptiveSubSearches` (2 tests): asserts both `_decompose_change_query` and `_recency_channel_chunks` do **not** run when a `filter_ast` is present, with a no-filter control (same CHANGE signal) proving the gate — not an unrelated condition — is what suppresses them, so the skip-assertion can't pass vacuously.

probablyrobot (approved):

PR #1048, opened 2026-06-08
Fix pgvector entity upsert to accumulate source provenance instead of clobbering (#1039)

filippetrovic:
## Summary

Fixes the pgvector entity upsert so it **accumulates** `source_document_ids` and `source_chunk_ids` on conflict instead of clobbering them, mirroring the Neo4j `ON MATCH` semantics so the two backends stay in sync.

**Problem:** When re-ingesting a document that shares an entity with a prior document, the pgvector `ON CONFLICT DO UPDATE` replaced the entire source-ids array with only the new write's ids. On a PostgreSQL + Neo4j stack this made `forget()` misclassify multi-sourced *survivor* entities (still referenced by another document): the clobbered entity was neither an orphan nor a recognized survivor, so its provenance-strip step never ran — leaving the forgotten document's id stranded on the entity in Neo4j.

**Solution:** The upsert now uses a scalar subquery to
1. concatenate existing + incoming arrays (existing first, so incoming ids carry the higher/most-recent ordinals),
2. de-duplicate by element, preserving each id's most-recent position,
3. keep the 100 most-recent **distinct** ids, and
4. re-aggregate newest-last (matching Neo4j's tail-cap order).

The dedup is what guards the core regression: re-extracting the *same* document many times no longer burns through the 100-id cap and evicts prior **distinct** provenance.

## Changes

- Add `_accumulate_source_ids_sql()` helper that builds the `ON CONFLICT DO UPDATE` set-expression (union + dedup + recency tail-cap). The cap is an inlined int constant and the column name is one of two hard-coded literals — no bind param (avoids an executemany VALUES collision) and no injection surface.
- Use the helper for both `source_document_ids` and `source_chunk_ids` in **both** upsert paths: `_upsert_entity()` (single) and `upsert_entities_batch()` (batch).
- Remove the `xfail` from `test_forget_removes_orphan_entity_from_search_and_explore` (it now passes).
- Add a focused regression test (`test_pgvector_accumulate_source_ids.py`) covering both paths, dedup/no-duplicate, and the same-document-repeat eviction guard.
- CHANGELOG `### Fixed` entry.

Scope is deliberately narrow: Neo4j (already accumulates), the relationship upsert path (relationships live in the graph backend on a PG+Neo4j stack), and the reserved bi-temporal columns are all untouched.

## Test plan

- [x] Unit suite green (`make test-unit`: 7746 passed, 43 skipped; 83% coverage)
- [x] New regression test added (gated on a live Postgres via `make dev`)
- [ ] Integration suite (`test-integration`, incl. the un-xfailed forget-cascade test) — runs in CI on a Docker-capable runner


probablyrobot (approved):

filippetrovic:
## 🔍 Multi-perspective review — PR #1048

7 independent reviewers (Security, Performance, Correctness, Database, Test, Docs, Devil's Advocate) plus a focused security audit. **CI is green: 11/11 checks pass, including `test-integration` (the un-xfailed forget-cascade test + the new accumulate test ran against real Postgres).**

**Overall assessment: APPROVE — one MEDIUM to address (cap fidelity).** Six of seven reviewers approve as-is; the core fix is correct, performant, well-tested, and well-documented. The one substantive finding is a chunk-id cap mismatch with Neo4j (below), which will be fixed in a follow-up commit on this PR.

### MEDIUM
- **M1 — [Performance / Database] `src/khora/storage/backends/pgvector.py:51`** — `_SOURCE_IDS_CAP = 100` is applied to **both** `source_document_ids` and `source_chunk_ids`, but Neo4j caps them **differently**: `entity_source_document_ids_max = 100` and `entity_source_chunk_ids_max = 250` (`neo4j.py:398-399` / `:765-766`). The helper's comment claims "the two backends must agree," which holds for document ids (100 = 100) but **not** chunk ids (100 vs 250) — pgvector under-retains chunk provenance. Note this does **not** break the `forget()` fix this PR targets: `forget()` classifies orphan/survivor off `source_document_ids` (which agree at 100). The divergence is in chunk-provenance retention only. **Fix:** split into per-column caps (`doc = 100`, `chunk = 250`) wired to mirror Neo4j, and make the comment accurate.

### LOW
- **L1 — [Devil's Advocate] `pgvector.py:54-82` vs `neo4j.py:1714-1715`** — pgvector now **dedups**; Neo4j does **not** (it concatenates + tail-slices, so it can hold duplicate ids). This is safe for `forget()` (which only checks membership + length via `_is_orphan`/`_is_survivor`, `_forget_cascade.py:149-156`), but the intentional asymmetry should be noted in a code comment.
- **L2 — [Test] `tests/integration/test_pgvector_accumulate_source_ids.py:177-211`** — the same-doc-repeat eviction test asserts only `source_document_ids`. Add a `source_chunk_ids` symmetry assertion (both columns share the helper).
- **L3 — [Test]** — no exact cap-boundary case (insert exactly the cap of distinct ids, then one more, assert the oldest is evicted). Strengthens confidence in the `LIMIT` logic.
- **L4 — [Test]** — tests assert set-equality but not newest-at-tail recency ordering; an ordering assertion would lock in the tail-cap contract.
- **L5 — [Devil's Advocate] PR description** — make the entities-only scope explicit: relationships are not upserted into pgvector on a PG+Neo4j stack (they live in the graph backend), so their path is correctly out of scope. *(Verified correct.)*

### Cleared / validated
- **Security: CLEAN** — the raw-SQL helper has no injection surface: `column` is only ever called with the two hard-coded literals (4 call sites verified), the cap is an inlined int constant, no user input reaches the string, and the `# noqa: S608` suppression is justified. No secrets; only new import is `sqlalchemy`.
- **Correctness: APPROVED** — accumulate (`existing || incoming`), dedup (`GROUP BY elem`, `MAX(ord)`), tail-cap (`ORDER BY MAX(ord) DESC LIMIT`), newest-at-tail (`array_agg ORDER BY last_ord`) all verified; NULL/empty handling, same-doc-repeat eviction guard, and batch-row conflict semantics correct; `(xmax = 0)` is_new detection unaffected; both upsert paths fixed.
- **Database: APPROVED (high confidence)** — `entities.<col>` / `excluded.<col>` references valid in `DO UPDATE SET`; `uuid[]` type-safe; scalar subquery returns a single array; advisory-lock + executemany composition sound; no migration; constraint intact.
- **Performance: APPROVED** — unnest input bounded (≤ cap + incoming), `LIMIT` pushed into sort, no quadratic blow-up at batch scale, negligible lock-hold impact.
- **Docs: CLEAN, no drift** — CHANGELOG entry and helper docstring accurate; no existing doc describes the old clobber semantics.
- **Test: APPROVED** — batch + single paths, dedup/no-duplicate, and the eviction guard covered; xfail removal justified; integration test is strong end-to-end cross-backend proof.

---
*Posted by the `/team:review` multi-agent reviewer team.*

filippetrovic:
### ✅ Addressed review finding M1 (+ L1, L2, L4) — commit `992ec51`

- **M1 (chunk-cap fidelity):** replaced the single `_SOURCE_IDS_CAP = 100` with per-column constants `_SOURCE_DOCUMENT_IDS_CAP = 100` and `_SOURCE_CHUNK_IDS_CAP = 250`, each mirroring the corresponding Neo4j default (`entity_source_document_ids_max` / `entity_source_chunk_ids_max`, `neo4j.py:398-399`). `_accumulate_source_ids_sql` now selects the cap per column (still an inlined int constant — no bind param, no injection surface). The helper comment is corrected to claim per-column agreement.
- **L1:** documented the intentional dedup asymmetry (Neo4j concatenates + tail-slices without dedup; pgvector dedups) and why `forget()` is resilient (it only checks membership + length).
- **L2:** the eviction regression test now exercises and asserts `source_chunk_ids` parity (stressed past the larger 250 cap).
- **L4:** added a newest-at-tail ordering assertion (`source_document_ids[-1] == doc_b`) to lock the tail-cap contract.

CHANGELOG updated to state the per-column caps. `make format` / `make lint` clean. CI re-running on the new HEAD.

L3 (exact cap-boundary test) and L5 (PR-description scope note) were judged optional; L5 is covered below — relationships are not upserted into pgvector on a PG+Neo4j stack (they live in the graph backend), so their path is correctly out of scope.

PR #1049, opened 2026-06-08
Add filter-conformance CI job with per-backend parallelism and read-only Postgres execution

filippetrovic:
## Summary
- Add a new `.github/workflows/filter-conformance.yml` GitHub Actions job that runs the recall-filter conformance corpus across a parallel backend matrix (python, postgres, chronicle, sqlite_lance, cypher, surrealdb, weaviate), separate from the main test job so it gates PRs without slowing the main suite. Each leg provisions only its own database — only the Postgres leg starts one (via an `if`-guarded `docker run`), and `fail-fast: false` lets every backend report independently.
- The Postgres leg seeds the conformance store **once** in a dedicated pre-step that also emits a seed-id→chunk-id JSON map; the pytest step then runs **read-only** under `-n auto`, reading the map instead of re-seeding, so cases fan across xdist workers with no write contention.
- Register a `filter_conformance` pytest marker and exclude it from the main test/integration jobs (the default `addopts` selector **and** the integration job's `-m`) so the corpus never double-runs.
- Land the conformance test module, the live-Postgres seam (skeleton `khora_chunks` coordinator + read-only predicate runner), the one-time seeder, a no-DB seed-map round-trip test, and a compiler-registry drift guard that asserts every registered `(engine_id, storage_target)` compiler key resolves and fails loudly (`UnknownCompilerError`) on drift.
- Prune the string-key operator cases off the Postgres backend at the source (the seeder denormalizes only date keys onto the chunk row); the python/chronicle oracles still validate the full corpus.

## Test plan
- [x] `make format` + `make lint` clean (0 errors)
- [x] Conformance corpus green locally without Docker: python and chronicle legs **67 passed** each (60 cases + 5 registry-guard + 2 seed-map); guard resolves all five registered compilers
- [x] Marker exclusion verified: the main test/integration selectors collect **zero** conformance cases (no double-run)
- [ ] Postgres leg (live `khora_chunks` execution) runs in CI — requires the Postgres service


filippetrovic:
## Consolidated Review — multi-perspective team + security audit

**Overall: APPROVE.** No correctness or security defects across 7 reviewer perspectives + a dedicated security audit. All four acceptance goals hold (separate gating workflow, per-leg DB isolation, one-time seed + read-only cases, marker excluded from the main job). The findings below are efficiency / maintainability / design improvements suitable for follow-up — none block merge. CI note: the live-Postgres leg initially failed on a namespace FK violation; fixed in this PR (child rows seed under the row-level `memory_namespaces.id`, not the stable namespace_id) and re-running.

### MEDIUM
- **M1 [Performance]** `tests/integration/matrix/_conformance_pg.py:187` — `run_predicate` calls `create_async_engine(...)` per case, so each conformance case spins up (and disposes) a fresh connection pool. Reusing one engine across the leg (e.g. a session-scoped fixture) would cut dozens of pool allocations per run. Functionally correct today (every engine is disposed in a `finally`); the Database review rated it "acceptable", Performance rated it the top efficiency win.
- **M2 [Maintainability]** `tests/integration/matrix/test_compiler_registry_keys.py` — the guard is rename/removal-tight but **add-blind**: a newly-registered compiler that isn't added to `EXPECTED_KEYS` still passes (the test only checks the listed keys resolve). The comment documents the "keep in lockstep" contract, but consider asserting the registry's full key set **equals** `EXPECTED_KEYS` so a new unlisted registration fails loudly instead of relying on review. (Raised independently in QA + Devil's Advocate.)

### LOW
- **L1 [Docs]** `.github/workflows/filter-conformance.yml` (seed step) — add a one-line hint: if the postgres leg fails with "missing from seed map", conformance cases were added and the seeder must be re-run (`python -m tests.integration.matrix._conformance_seed`).
- **L2 [Docs]** `src/khora/filter/conformance.py` (`seed_case` / `_STRING_OP_BACKENDS`) — add a forward-looking note that string document keys aren't yet denormalized onto `khora_chunks`, which is why string-key F-OP cases are pruned off postgres; denormalizing them would re-enable those cases on the postgres leg.
- **L3 [Design]** The matrix includes four guard-only legs (cypher/surrealdb/weaviate/sqlite_lance) that currently run zero conformance cases (no harness executor yet) — intentional per "the matrix grows as compilers land", and they run the registry/compile guard cheaply in parallel. A reader may expect cases on those legs; a short note or splitting "guard" vs "matrix" jobs would make the intent explicit. (Devil's Advocate.)
- **L4 [Design]** The main-job exclusion relies on every conformance test also carrying the `integration` marker (so `-m "integration and not filter_conformance"` catches it). A conformance test that omitted `integration` could leak into the main job. Consider a meta-test asserting the main selection collects zero `filter_conformance` tests. (Devil's Advocate.)

### INFO
- **[Scope, Devil's Advocate]** This PR touches the merged conformance harness twice — the string-key prune and the namespace-FK fix. Both are defensible here (the FK fix was forced by this PR being the first live-Postgres exercise of `seed_case`, and the prune is the per-backend tuning the harness explicitly delegates to the CI ticket), but a reviewer preferring tighter scope could land the FK fix separately.
- **[Security]** PASS — SQL is fully parameterized (`id.in_` + compiled predicate, no interpolation); workflow uses only the standard `khora/khora` test fixture credentials (matches ci.yml); GitHub Actions pinned to major versions; no secrets, no injection, no unsafe deserialization, no path traversal (the seed-map path is CI-controlled). Security audit verdict: **PASS**, no auto-fixes required.
- **[Correctness/Database]** The namespace-FK fix was independently verified correct against the dual-id convention (`use id for child-table FK lookups`); the shared-engine coordinator wiring avoids pool exhaustion; the read-only `id = ANY(:ids)` scoping precludes cross-case aliasing.

_Reviewers: security, performance, correctness, test, database, docs, devil's-advocate (parallel) + security audit._

probablyrobot (approved):

PR #1050, opened 2026-06-08
Add Chronicle recall-filter integration tests over a real sqlite_lance store

filippetrovic:
## Summary
- Add `tests/recall/test_filter_chronicle.py`: full-stack integration tests for the recall-filter on the **Chronicle** engine over a real embedded **SQLite + LanceDB** store (hermetic — no Docker, self-skips when the embedded extra is absent).
- Mirrors the sibling skeleton-pgvector integration test's shared corpus and its exact **5-in-scope / 5-out-of-scope** split (each out-of-scope row violates exactly one predicate; one in-scope row sits exactly at the date bound) and asserts the live `recall(filter=...)` narrows to exactly the same in-scope set.
- Because Chronicle's recall DTO does not hydrate the document-projection keys, and the system date-key compare currently raises on the embedded store's timezone-naive datetimes, the three-predicate AND is expressed over the **metadata axes** the engine carries on the real store (channel / when / tag) — standing in for the document key and the system date key while preserving an **identical in-scope set**. Both substitutions are documented in the module docstring and the corpus comment.
- Adds a **strict-xfail guard** pinning the system `occurred_at` date-key timezone-comparison limitation on the embedded store: it flips to a hard failure (XPASS) the moment the post-filter compiler normalizes timezone, signalling it is time to drop the workaround.
- Asserts the post-filter increments the `unindexed_metadata` telemetry counter on the metadata path (carrying the leaf op), with two negative controls (a non-metadata filter and a no-filter recall both leave it silent).
- **Tests only** — no `src/` changes.

## Test plan
- [x] New integration test passes in isolation: `uv run pytest tests/recall/test_filter_chronicle.py -m integration -v --no-cov` → **5 passed, 1 xfailed**
- [x] `make format` clean; `make lint` exit 0
- [x] No `src/` changes (`git diff --stat -- src/` empty)
- [ ] CI integration job (Docker-backed Postgres/Neo4j) green

### DIFF
--- a/tests/recall/test_filter_chronicle.py
+++ b/tests/recall/test_filter_chronicle.py
@@ -1,0 +1,527 @@
+"""Full-stack integration tests for the recall-filter on Chronicle + sqlite_lance.
+
+This is the Chronicle counterpart to ``test_filter_skeleton_pgvector.py`` (V1):
+where V1 drives the public ``Khora.recall(filter=...)`` API end-to-end through
+the Skeleton engine into a real pgvector ``khora_chunks`` lake, this file drives
+the SAME contract through the Chronicle engine over a real embedded SQLite +
+LanceDB store. The engine→backend wiring threads ``filter_ast`` from the facade
+through Chronicle's recall path into the deterministic post-filter, so row
+filtering is effective end-to-end — these tests pin that contract on the
+embedded stack.
+
+Unlike ``test_chronicle_filter_composition.py`` (which is the MOCKED, unit-level
+proof of how Chronicle composes the two compiled filter halves on a
+``MagicMock`` storage), this file's value-add is the LIVE corpus isolation: real
+``Khora.remember()`` writes, real ``Khora.recall(filter=...)`` reads, real
+sqlite_lance roundtrip. It does NOT re-mock what the composition suite already
+proves; it proves the same logical narrowing survives the live write/read path.
+
+EQUIVALENCE TO V1 — same corpus + split, Chronicle-carried axes
+===============================================================
+V1's filter is ``source_name == "linear" AND occurred_at >= 2026-04-05 AND
+metadata.tag IN {"urgent", "release"}``. This file reuses V1's shared corpus and
+its exact 5-in-scope / 5-out-of-scope split (each out-of-scope row violating
+EXACTLY ONE predicate, one in-scope row sitting EXACTLY at the date bound). Two
+of V1's three predicate KEYS, however, do not function on the REAL Chronicle
+sqlite_lance recall path, so the three-predicate AND is expressed over the axes
+Chronicle actually carries on the real store — metadata keys — standing in for
+V1's ``source_name`` (a denormalized document key) and ``occurred_at`` (a system
+date key). The in-scope SET is identical; the contract proven here is "a
+real-store multi-predicate AND narrows to exactly V1's in-scope set". The system
+``occurred_at`` axis is pinned separately by a strict-xfail (below).
+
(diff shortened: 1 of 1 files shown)


filippetrovic:

Six independent reviewers (security, performance, correctness, test-quality, docs, devil's-advocate) reviewed this PR. Net assessment: **APPROVE — no blocking findings.** The change is a single hermetic, additive test file (no `src/` changes); reviewers converged on quality/clarity/efficiency improvements suitable as pre-merge polish or fast-follow.

> Note on severity: two findings were submitted as CRITICAL/HIGH but are self-downgraded in their own write-ups (no crash, data-loss, or security risk); they are reclassified below to match impact.

### MEDIUM
- **M1 [Docs]** module docstring "EQUIVALENCE" section — the "in-scope SET is identical" framing should state explicitly that the **system `occurred_at` key** equivalence is *not* proven on the real store; only the `metadata.when` substitute is. Reword so a skimmer can't read it as all three axes being proven end-to-end. (docs + devil's-advocate)
- **M2 [Test]** `test_non_metadata_filter_silent_for_unindexed_metadata` asserts only counter-silence; add a liveness assertion (the recall actually ran/returned) so a swallowed exception can't false-pass the "does not crash" claim the docstring makes. (test-quality)
- **M3 [Performance]** redundant corpus re-seeding: 4 tests each re-run the full real-ingestion seed (~50s apiece — the dominant cost of the file's ~5min runtime, ~30× higher than the reviewer's first estimate). A module/session-scoped seeded-corpus fixture (shared corpus, per-test telemetry counter kept fresh) would cut runtime substantially and ease xdist memory pressure. (performance)
- **M4 [Coverage]** on the real store the test exercises only the **metadata** axis: document-key predicates resolve absent (DTO doesn't hydrate `source_document`) and the system date key is xfail'd. A follow-up that exercises doc-key / system-date-key behaviour end-to-end is warranted once the engine hydrates the projection / the date-compare is tz-normalized. Accepted, documented scope limitation. (devil's-advocate + test-quality)

### LOW
- **L1 [Performance]** no per-test timeout marker; a hang would block a worker until the job-level timeout. If added, use a **generous** value (~180s) — the suggested 30s would fail legitimately ~50s tests.
- **L2 [Test]** reranking disabled diverges from the pgvector reference fixture (justified — reranking only reorders, never adds/drops rows; kept off for hermeticity/determinism).
- **L3 [Correctness]** `_content_for` uses a SHA1[:8] content/label join key; collision probability is negligible and would fail loudly (never false-pass) if it ever occurred.
- **L4 [Test/Scope]** when the tracked timezone-normalization fix lands, the strict-xfail flips to XPASS but the green test still uses `metadata.when`; a code comment recording the "fold the date axis back to the system key" rework step would de-risk silent drift.

### Security Audit
**Clean.** The diff is one hermetic test file with no security surface — no secrets/credentials, no dependency or crypto changes, no input-handling or service boundaries, no leaked internal identifiers. The dedicated Security Reviewer's OWASP-oriented pass returned zero findings.

### Strengths (consensus)
Surgical corpus design (5-in/5-out, each out-row a single-predicate violation, one in-row exactly at the inclusive bound); expected set encoded by construction (no tautology); strict-xfail correctly pins the system date-key crash and will XPASS the moment it's fixed; telemetry assertion checks the leaf op with two real negative controls; cleanly complementary to (not duplicative of) the mocked composition suite.

**Overall: APPROVE.** No CRITICAL/HIGH. MEDIUM items are clarity/efficiency/coverage polish; none block correctness or merge.

probablyrobot (approved):

probablyrobot (approved):

PR #1051, opened 2026-06-09
Compose recall filter across VectorCypher adaptive sub-searches

filippetrovic:
## Summary

Thread the caller-supplied recall filter through **every** adaptive sub-search in the VectorCypher engine so no retrieval channel silently drops or mis-applies it. Previously several adaptive paths (change-decomposition, the recency channel) bailed out when a filter was present, and the graph chunk channel applied no caller filter at all.

- **Conditional graph over-fetch** — new `query.metadata_overfetch_multiplier` (int, default 3, 2–10). The graph chunk channel over-fetches **only** when a residual metadata predicate needs an in-memory post-filter (detected by comparing the Cypher compiler's consumed keys against the filter's leaf keys); capped at `min(limit * multiplier, 200)`. No over-fetch for an absent filter or a system-key-only filter.
- **Graph-channel filtering** — push system keys into the graph `Chunk` channel via the Cypher compiler, and run a full-AST Python post-filter over graph candidates (covering the PPR path too) so a metadata leaf nested inside an `$or` (which the graph predicate collapses to a non-constraining superset) is correctly re-constrained.
- **Date-gated recency synthesis** — the EXPLICIT temporal category is synthesized only when the filter constrains a date key (`occurred_at` / `created_at`). A non-date filter (e.g. a metadata field) runs the normal temporal detector and no longer triggers recency decay.
- **Restrictive-fallback skip** — the unfiltered fallback re-run is skipped whenever a caller filter is present (sparse-but-correct beats complete-but-wrong).
- **Session fan-out intersect** — when the caller constrains the channel, fan-out is intersected with that constraint (skipped when disjoint); the caller's constraint is never overwritten. Best-effort efficiency guard; correctness rides on filter AND-composition.
- **Carry across sub-searches** — the change-decomposition and recency sub-searches now carry the filter (the recency channel reads the relational table and post-filters in memory).
- **Unsupported-predicate surfacing** — a predicate the Cypher channel cannot express surfaces to the caller instead of silently degrading to a vector-only fallback (which stays reserved for transient graph errors).
- **Observability** — when the graph channel empties under a filter while the SQL channels return rows, a degradation is recorded and a new `*.graph_channel.empty_total` counter increments (telemetry contract updated).

New `filter/execute.py` helpers (single AST walk) back the leaf-key, date-key, residual-metadata, and channel-constraint detection.

## Test plan
- [x] Unit + pushdown-spy tests pass (`make test-unit`): new behavior covered for every item above, including mutation-verified non-vacuity; touched-module coverage 80–100%. Pre-existing failures in this sandbox are limited to optional extras not installed here (`langgraph`, `lancedb`/sqlite_lance) — unrelated to this change.
- [ ] Integration leg + 77% coverage gate (`make test-integration`, requires Docker) — runs in CI.
- [ ] Filter-conformance CI job green.

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -259,6 +259,7 @@ Prefix: `KHORA_QUERY_`. See [query-engine/retrieval-tuning.md](query-engine/retr
 | `KHORA_QUERY_HYDE_CYPHER_LIMIT` | `20` | Max entities returned per HyDE-Cypher template execution. |
 | `KHORA_QUERY_ENABLE_RERANKING` | `true` | Cross-encoder reranking of top candidates. |
 | `KHORA_QUERY_TEMPORAL_SQL_PUSHDOWN` | `true` | Push relative-date filters into SQL WHERE clauses. |
+| `KHORA_QUERY_METADATA_OVERFETCH_MULTIPLIER` | `3` | Over-fetch multiplier (2–10) for the VectorCypher graph chunk channel when a caller filter has a metadata predicate that can't push down to Cypher and must be post-filtered in memory. The graph fetch widens to `min(limit * multiplier, 200)` so fusion still has candidates after filtering; no effect for no-filter or system-key-only filters. |
 
 ## Tenancy
 
--- a/docs/telemetry-contract.json
+++ b/docs/telemetry-contract.json
@@ -1046,6 +1046,17 @@
       "description": "Issue #904 (ADR-001). VectorCypher relationship-fetch silent fallback - incremented when the parallel rels_task raises and the retrieve path continues without relationships. The same event is also appended to RecallResult.engine_info['degradations']. Asymmetric prior art: _cypher_expand failure already sets engine_info['graph_fallback'] / 'graph_error'; this counter closes the rel-fetch gap. NO namespace_id label - cardinality rule. See docs/architecture/failure-observability-contract.md."
     },
     {
+      "name": "khora.vectorcypher.graph_channel.empty_total",
+      "type": "counter",
+      "stability": "public",
+      "unit": "1",
+      "owner": "vectorcypher",
+      "labels": [
+        {"name": "reason", "values": ["empty_under_filter"], "_comment": "Low-cardinality enum; extend when adding new graph-channel-empty paths."}
+      ],
+      "description": "ADR-001. VectorCypher graph chunk channel emptied by the in-memory metadata post-filter while the SQL-pushed vector/BM25 channels returned filtered rows - i.e. the graph side under-recalled relative to the completeness backstop (metadata is not pushable to Cypher, so a metadata predicate is enforced in memory after the graph fetch). The same event is also appended to RecallResult.engine_info['degradations']. NO namespace_id label - cardinality rule. See docs/architecture/failure-observability-contract.md."
+    },
+    {
       "name": "khora.vectorcypher.chunk_mirror.degraded_total",
       "type": "counter",
       "stability": "public",
--- a/docs/telemetry-contract.json
+++ b/docs/telemetry-contract.json
@@ -1046,17 +1046,6 @@
       "description": "Issue #904 (ADR-001). VectorCypher relationship-fetch silent fallback - incremented when the parallel rels_task raises and the retrieve path continues without relationships. The same event is also appended to RecallResult.engine_info['degradations']. Asymmetric prior art: _cypher_expand failure already sets engine_info['graph_fallback'] / 'graph_error'; this counter closes the rel-fetch gap. NO namespace_id label - cardinality rule. See docs/architecture/failure-observability-contract.md."
     },
     {
-      "name": "khora.vectorcypher.graph_channel.empty_total",
-      "type": "counter",
-      "stability": "public",
-      "unit": "1",
-      "owner": "vectorcypher",
-      "labels": [
-        {"name": "reason", "values": ["empty_under_filter"], "_comment": "Low-cardinality enum; extend when adding new graph-channel-empty paths."}
-      ],
-      "description": "VectorCypher graph chunk channel emptied by the in-memory metadata post-filter while the SQL-pushed vector/BM25 channels returned filtered rows - i.e. the graph side under-recalled relative to the completeness backstop (metadata is not pushable to Cypher, so a metadata predicate is enforced in memory after the graph fetch). The same event is also appended to RecallResult.engine_info['degradations']. NO namespace_id label - cardinality rule. See docs/architecture/failure-observability-contract.md."
-    },
-    {
       "name": "khora.vectorcypher.chunk_mirror.degraded_total",
       "type": "counter",
       "stability": "public",
--- a/docs/telemetry-contract.json
+++ b/docs/telemetry-contract.json
@@ -1054,7 +1054,7 @@
       "labels": [
         {"name": "reason", "values": ["empty_under_filter"], "_comment": "Low-cardinality enum; extend when adding new graph-channel-empty paths."}
       ],
-      "description": "ADR-001. VectorCypher graph chunk channel emptied by the in-memory metadata post-filter while the SQL-pushed vector/BM25 channels returned filtered rows - i.e. the graph side under-recalled relative to the completeness backstop (metadata is not pushable to Cypher, so a metadata predicate is enforced in memory after the graph fetch). The same event is also appended to RecallResult.engine_info['degradations']. NO namespace_id label - cardinality rule. See docs/architecture/failure-observability-contract.md."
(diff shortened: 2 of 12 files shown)


filippetrovic:

8 reviewers (security, performance, correctness, test, database, observability, docs, devil's advocate) plus a dedicated security audit. Overall: **strong APPROVE pending one small observability fix.** The change threads the caller filter deterministically across every adaptive sub-search (vector, BM25, graph, recency, change-decomposition) while respecting the SQL/Cypher/Python pushdown boundaries.

### CRITICAL
_None._

### HIGH
- **[Observability]** `engines/vectorcypher/retriever.py` (graph-channel-empty degradation) — the new degradation path records the `Degradation` and increments the empty-channel counter but does **not** emit the convention-mandated `WARNING` log (the failure-observability contract / ADR-001 requires `WARNING` for a degradation/fallback path; the sibling relationship-fetch degradation a few lines above logs correctly). No `exc_info` needed (not exception-triggered). **Fix:** add a `logger.warning(...)` alongside the counter. _(Being fixed in a follow-up commit on this branch.)_

### MEDIUM
- **[Docs]** `config/schema.py` / `docs/configuration.md` — the new `metadata_overfetch_multiplier` query setting (`KHORA_QUERY_METADATA_OVERFETCH_MULTIPLIER`) is user-tunable but absent from the configuration reference. **Fix:** add a row to the Query section. _(Being fixed.)_
- **[Maintainability]** `engines/vectorcypher/retriever.py` (graph fetch limit) — the `min(limit * multiplier, 200)` cap uses a tacit `200` constant. **Fix:** add a one-line comment explaining it guards against a misconfigured multiplier. _(Being fixed.)_
- **[Performance]** `engines/vectorcypher/retriever.py` — `compile_python(filter_ast, ...)` is compiled separately on the graph and recency post-filter paths. Optional micro-opt: compile once and reuse (<1ms, only when both paths fire). Not blocking.
- **[Process/Docs]** The EXPLICIT temporal synthesis is now also gated on a date-constraining caller filter (`occurred_at`/`created_at`), unifying caller-filter date constraints with API `temporal_filter`. Worth calling out explicitly. _(PR description updated.)_

### LOW
- **[Docs]** Optional CLAUDE.md gotcha entry for graph-channel metadata post-filtering / over-fetch tuning (operator visibility).

### Security Audit
**PASS — no findings.** Cypher and Postgres predicates are parameterized (driver `$param` binds, no string interpolation); the in-memory post-filters are pure lambda closures (no `eval`/`exec`/attribute injection); inputs are bounded (`metadata_overfetch_multiplier` `ge=2 le=10`, fetch capped at 200); the new counter is cardinality-safe (`reason` label only, no `namespace_id`); no secrets in the diff. Pattern hits were all in test code (mock values, test method names, safe `getattr(..., default)` spies).

### Test coverage
All acceptance criteria are covered by spy-style unit tests, **mutation-verified** for non-vacuity (each path disabled in turn, matching test confirmed to fail), with negative controls. Touched-module coverage 80–100%.

### Overall assessment
**APPROVE** once the HIGH observability log line lands. All other items are minor/optional. No correctness, security, or data-safety issues found.

filippetrovic:
Four items to address before merge. Filter-survival threading across the channels is correct — the prior interim skip-gates are replaced by real threading, the BM25 channel reuses the vector channel's exact `khora_chunks` WHERE, and the legacy `chunks` coordinator is refused under a filter. The issues are observability-contract and coverage, not a filter leak.

**1. Graph-channel-empty counter is renamed away from the declared contract metric.**
`docs/telemetry-contract.json` already declares `khora.recall.filter.graph_channel_empty` (line 1491) for this exact condition. This PR adds a second counter `khora.vectorcypher.graph_channel.empty_total` and increments that one instead, so the declared `graph_channel_empty` is never incremented — a dead contract metric plus two names for one condition. Increment the existing `khora.recall.filter.graph_channel_empty`, or drop the contract entry — not both.

**2. `under_filled` counter is not wired.**
`khora.recall.filter.under_filled` is declared in the contract (line 1483) but appears nowhere in the retriever. The graph over-fetch path records the empty-channel degradation but never emits `under_filled` when the post-filter eliminates more than `limit` candidates. Wire it on the over-fetch elimination path, or state explicitly that it is deferred.

**3. The partial-failure raise contract is documented but cannot fire.**
`dual_nodes.py` comments that the `compile_cypher` call "deliberately" surfaces `RecallFilterUnsupportedError` outside the ClientError/timeout handler, but every call passes `on_unsupported="split"`, which never raises — it returns `"true"` and omits the key from `consumed_keys`. So a genuinely unsupported predicate is silently split to a non-constraining `true` + post-filter rather than raising. The tests asserting the raise force it via `side_effect=RecallFilterUnsupportedError`, not the real split-mode path. Either use `on_unsupported="raise"` for the unsupportable case, or correct the comments to say surfacing is deferred and add a residual-after-post-filter check if the contract must hold.

**4. No real-DB coverage for the threaded filter; patch coverage is 22%.**
The composition tests in `test_vectorcypher_filter_pushdown.py` are entirely mock/spy with fabricated candidate sets — fine for the seam-spy assertions, but nothing exercises the threaded `filter_ast` against a real Postgres+Neo4j store end-to-end. Codecov reports 22.2% patch coverage (`retriever.py` 12.1%, `execute.py` 28.5%, `dual_nodes.py` 0%). Add a real-DB integration case covering the graph / recency / CHANGE channels under a filter, or confirm the Docker-gated integration job covers them.

Non-blocking: `filter_constrains_date_key` (`execute.py`) walks every leaf, so a date key nested under `$or`/`$not` triggers EXPLICIT recency synthesis even though it is not a hard conjunctive constraint — inconsistent with `caller_channel_constraint`, which returns `None` for the same nesting. Ranking-only (the predicate is still enforced on every channel), but worth aligning.

filippetrovic:
Thanks — all four addressed in the latest push, plus the non-blocking alignment.

**1. Graph-channel-empty counter → declared contract metric.** Removed the duplicate `khora.vectorcypher.graph_channel.empty_total` (counter + contract entry). The retriever now calls a new `record_graph_channel_empty()` helper that increments the already-declared `khora.recall.filter.graph_channel_empty`. It fires whenever a filter narrows the graph channel to empty (the contract semantics); the richer degradation record stays gated on the "vector/BM25 still returned rows" backstop condition.

**2. `under_filled` wired.** Added `record_under_filled()` and emit it once per recall from the engine when a caller filter is present and the returned candidate set came back under the requested `k`. (Wired to the contract definition — "returned fewer results than the requested limit" — rather than only the over-fetch elimination path, so it captures every channel.)

**3. Partial-failure comment corrected.** The graph-fetch compile stays `on_unsupported="split"`, so it does not raise for metadata leaves — they're deferred to the full-AST in-memory post-filter, which enforces every leaf. Reworded the comment to say exactly that: split mode does not raise in normal operation; the call sits outside the ClientError/timeout handler only so that a genuine compiler error surfaces rather than being masked as a transient Neo4j failure. No artificial raise added (the post-filter is the universal backstop, so there is no residual-unsupported case at the graph layer).

**4. Coverage.** Added unit coverage for the previously-0% graph-fetch filter branch (`get_chunks_by_entities`): a system-key predicate is spliced into the Cypher WHERE as a bound `$param`, and a metadata-only filter consumes nothing (deferred to post-filter). Also added tests that both wired counters fire and that the realigned date-key detector behaves. End-to-end real-DB coverage across the graph / recency / CHANGE channels remains the explicit scope of the dedicated follow-up integration slice this PR blocks; the Docker-gated conformance matrix (postgres / cypher / etc.) already gates per-backend compiler correctness.

**Non-blocking — date-key alignment.** `filter_constrains_date_key` now inspects only the root conjunction (mirroring `caller_channel_constraint`), so a date predicate buried under `$or` / `$not` no longer triggers EXPLICIT recency synthesis. A bare single predicate parses to `AND`-with-one-child, so the common top-level date case still flags; added tests for the `$or` / `$not` cases.

probablyrobot (approved):

filippetrovic:
Re-review of the latest push.

- Graph-channel-empty counter, `under_filled` wiring, and the date-key conjunctive gate: confirmed fixed. One counter for the empty condition using the declared contract name; `under_filled` fires on the filtered-recall-returns-fewer-than-`limit` path, which matches the declared metric semantics; the date-key gate now fires only on top-level conjunctive constraints.
- `on_unsupported` partial-failure comment: softened but still rationalizes guarding against a `RecallFilterUnsupportedError` that split-mode cannot produce — the sole raise site is gated on `on_unsupported="raise"`, and every call here uses `"split"`. Drop or further soften that clause so the comment doesn't describe an unreachable path.
- Mock-only coverage of the threaded filter: the real Postgres+Neo4j row-set validation across the graph / recency / CHANGE channels is owned by the dedicated VectorCypher integration test (result-set parity against the skeleton-pgvector baseline). Deferring end-to-end validation to that test is fine; no real-DB test is required in this PR provided that one exercises these channels under a filter.

With the comment clause addressed, this is mergeable.

filippetrovic:
Done — dropped the clause that rationalized guarding against a `RecallFilterUnsupportedError` the split-mode compile cannot produce. The comment now only describes the reachable behavior: split mode omits an inexpressible leaf from `consumed_keys` and defers it to the full-AST in-memory post-filter, so the call does not raise. (`e460819`)

PR #1053, opened 2026-06-09
Make Alembic version-stamping atomic per migration with a session-scoped advisory lock

filippetrovic:
## Summary
- Fixes intermittent migration version-stamping failures on repeated back-to-back full-chain replays (e.g. `DROP SCHEMA public CASCADE` + migrate-to-head, run twice). The symptom was an Alembic `CommandError: ... expected to match one row when updating '028...' to '029...'; 0 found`.
- Root cause: the whole chain ran inside a single transaction, but the autocommit `CREATE INDEX CONCURRENTLY` migrations (run inside `op.get_context().autocommit_block()`) commit that outer transaction mid-chain, so a version-stamp `UPDATE ... WHERE version_num = '<prev>'` could find the table not exactly at the expected revision.
- Set `transaction_per_migration=True` so each migration commits its DDL **and** its version stamp atomically — the next stamp always reads a durably-committed prior revision.
- Switch the migration mutex from the transaction-scoped `pg_advisory_xact_lock` to a **session-scoped** `pg_advisory_lock`, acquired before the migration transaction and released in a `finally` on the same (NullPool) connection. This is required because per-migration commits would otherwise release a transaction-scoped lock mid-chain; it also closes a latent gap where the first autocommit block already dropped the transaction-scoped lock partway through the run.
- The explicit post-run `connection.commit()` is kept (now load-bearing on Postgres, since the outer `begin_transaction()` is a no-op context manager under `transaction_per_migration`) and its comment corrected.
- SQLite is unaffected: the flag is a no-op on a non-transactional-DDL dialect, and the advisory-lock paths are Postgres-only.
- Updates the migration docs and the affected unit tests (advisory-lock SQL assertions, a guard that the flag is configured), and adds a Postgres regression test that replays the full chain back-to-back and asserts every run stamps cleanly.

## Test plan
- [x] Unit tests pass (`make test-unit` — migration-infra suites green; the new flag/lock unit assertions pass)
- [x] `make lint` clean; `make format` clean
- [x] SQLite full-chain migration still runs cleanly under the new flag
- [ ] Integration tests pass in CI (the new `test_migration_replay_idempotent` Postgres test runs in the `test-integration` job; it self-skips where Postgres is unavailable and reproduces the back-to-back-replay scenario where it runs)
- [x] Manual verification: confirmed the advisory lock is acquired before the migration transaction and released in a `finally` on every path, and that ahead-detection / fresh-DB / version-column-widen behaviors are preserved


filippetrovic:
## 🤖 Multi-perspective review

Seven independent reviewers (security, performance, correctness, database, tests, docs, devil's advocate) plus a focused security audit reviewed this change against the committed diff.

**Overall assessment: ✅ APPROVE** — no CRITICAL or HIGH findings. The fix (per-migration atomic stamping via `transaction_per_migration=True` + a session-scoped `pg_advisory_lock` released in a `finally`) correctly resolves the repeated-replay version-stamping flake while preserving every invariant (fresh-DB, ahead-DB skip, `version_num` widen, DSN redaction, SQLite full-chain).

### CRITICAL
- None.

### HIGH
- None.

### MEDIUM
- None blocking. The devil's advocate raised two points and concluded both are acceptable:
  - **[Correctness/DB]** Lock release under asyncio cancellation / connection death — mitigated by `NullPool` (fresh connection per run) + the `finally` block + session-scoped locks auto-clearing on connection close. The `_release_advisory_lock` best-effort swallow with a generic warning is the right pattern.
  - **[DB]** With per-migration commits, a process killed mid-migration leaves the version table and schema consistent at the last *completed* migration (each migration commits its DDL + stamp together). This is strictly better than the previous whole-chain transaction and matches Alembic's resumable model.

### LOW (optional polish — not blocking)
- **[Tests]** `tests/integration/db/test_migration_replay_idempotent.py` — consider a one-line note on why 6 iterations is sufficient (≈ `1 − 0.5^6` ≈ 98% catch rate for a ~1-in-2 flake).
- **[Correctness/DB]** Optional one-line comment at the lock-acquire site noting the lifecycle invariant (acquired and released on the same `NullPool` connection within the `finally` before context exit).
- **[Tests]** Integration suite exercises single-process replay only (not concurrent-startup lock contention) — acceptable boundary; the acquisition backoff is unit-mockable if desired later.

### Security Audit
- **No findings.** Both advisory-lock calls use bound parameters (`:lock_id`); `LOCK_ID` is a non-secret deterministic constant. The new warning log contains no secrets/DSN. Existing `redact_dsn` coverage on connection-error paths is untouched. No new dependencies, no crypto, no user-controlled input, no auth surface.

### Notable strengths
- Regression test genuinely reproduces the flake (full `DROP SCHEMA` + re-migrate each iteration — not a no-op replay), and a unit guard asserts `transaction_per_migration=True` so the core fix can't be silently reverted.
- The session-scoped lock also closes a latent gap where the previous transaction-scoped lock was released at the first `autocommit_block` migration, leaving the rest of the chain unprotected.
- Docs (`CLAUDE.md`, `docs/migrations.md`), the stale-comment fix, and the drift-guard test were all updated consistently; remaining `pg_advisory_xact_lock` references are the *separate* dream-phase locks (intentionally transaction-scoped), not drift.

_No blocking changes requested. LOW items are optional and can land in a follow-up._

probablyrobot (approved):

PR #1054, opened 2026-06-09
Persist and read occurred_at on the embedded chunk write path

filippetrovic:
## Summary
- The embedded sqlite_lance chunk write path persisted `source_timestamp` but dropped `occurred_at`: a chunk written with a distinct event time read back with `occurred_at = NULL`, so an `occurred_at` recall filter silently evaluated against `source_timestamp` instead of the real event time.
- Add a nullable `occurred_at` column to the `chunks` table via a new dialect-portable migration. It materializes on both PostgreSQL and SQLite (the embedded test fixtures run the full Alembic chain against SQLite); existing rows land at `NULL`.
- Declare `occurred_at` on the `ChunkModel` ORM for parity with the migration and the sibling timestamp columns. PostgreSQL read/write population of the column is intentionally left for a separate change — this PR only adds the column for schema parity.
- Persist `occurred_at` in `create_chunks_batch` and read it back in `_row_to_chunk`, so a distinct event time now round-trips through the embedded store and is honored by the Chronicle event-time filter (`COALESCE(occurred_at, source_timestamp)`). The LanceDB rows are unchanged — SQLite remains the source of truth for chunk timestamps.

## Test plan
- [x] Unit tests pass (`make test-unit`: full unit + recall suite green, including migration-drift, migration-bundling, and SQLite-migration head checks)
- [x] Embedded recall-filter e2e + round-trip tests pass (`tests/integration/test_chronicle_filter_embedded.py`, `tests/unit/storage/backends/sqlite_lance/test_vector.py`)
- [ ] Full integration suite (PostgreSQL + Neo4j) — runs in CI
- Regression guard: the strengthened e2e test asserts a chunk with a distinct in-range `occurred_at` and an out-of-range `source_timestamp` now survives the `occurred_at` filter; verified to fail when the write path drops the column (mutation-checked) and pass with the fix. A direct write→read round-trip test asserts the stored `occurred_at` reads back equal to the written value and distinct from `source_timestamp`.

### DIFF
--- a/src/khora/db/migrations/versions/046_chunks_occurred_at.py
+++ b/src/khora/db/migrations/versions/046_chunks_occurred_at.py
@@ -1,0 +1,31 @@
+"""Add nullable occurred_at column to chunks.
+
+Revision ID: 046_chunks_occurred_at
+Revises: 045_khora_try_timestamptz
+Create Date: 2026-06-09
+
+Each chunk gains a nullable ``occurred_at`` TIMESTAMPTZ column recording the
+real-world event time the chunk's content refers to, distinct from its
+ingestion ``created_at``. The column is nullable - existing chunks land at
+NULL until a value is supplied. The plain ``op.add_column`` is
+dialect-portable, so it materializes on both PostgreSQL and SQLite (the
+sqlite_lance test fixtures run the full Alembic chain against SQLite).
+"""
+
+from collections.abc import Sequence
+
+from alembic import op
+from sqlalchemy import Column, DateTime
+
+revision: str = "046_chunks_occurred_at"
+down_revision: str | Sequence[str] | None = "045_khora_try_timestamptz"
+branch_labels: str | Sequence[str] | None = None
+depends_on: str | Sequence[str] | None = None
+
+
+def upgrade() -> None:
+    op.add_column("chunks", Column("occurred_at", DateTime(timezone=True), nullable=True))
+
+
+def downgrade() -> None:
+    op.drop_column("chunks", "occurred_at")
--- a/src/khora/db/models.py
+++ b/src/khora/db/models.py
@@ -253,6 +253,8 @@ class ChunkModel(Base):
     # decay path treats ``max(source_timestamp, last_accessed_at)`` as the
     # effective event time so frequently-recalled chunks stay fresh.
     last_accessed_at: Mapped[datetime | None] = mapped_column(DateTime(timezone=True), nullable=True)
+    # Real-world event time the chunk's content refers to, distinct from created_at.
+    occurred_at: Mapped[datetime | None] = mapped_column(DateTime(timezone=True), nullable=True)
 
     # Session attribution for agentic-framework adapters (#620).
     # Inherited from the parent document at chunking time.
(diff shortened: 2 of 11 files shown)


filippetrovic:
## 🔍 Multi-Perspective Review — PR #1054

Reviewed by a 7-reviewer team (security, performance, correctness, test, database, docs, devil's advocate) plus a dedicated security audit. **Consolidated verdict: ✅ APPROVE.** No CRITICAL or HIGH findings. The MEDIUM findings concern traceability of intentionally-deferred follow-up work, not the correctness of this change.

### CRITICAL
- None.

### HIGH
- None.

### MEDIUM
- **M1 — [Devil's Advocate / Test] Half-wired column on the server-side backend.** `ChunkModel.occurred_at` is declared and the migration adds the column on both dialects, but the PostgreSQL/pgvector write+read mappers are not wired to populate it, so PG chunks read back `occurred_at = NULL`. This is **intentional, in-scope deferral** — it parallels the pre-existing `source_timestamp` DTO denormalization gap on the PG `chunks` path, and is benign because the event-time post-filter uses `COALESCE(occurred_at, source_timestamp)` (no false-empty). The embedded backend (the actual bug being fixed) round-trips a distinct `occurred_at` correctly. **Resolution:** the PG population work is now tracked as a separate follow-up so it isn't rediscovered later.
- **M2 — [Devil's Advocate] Deferred scope not linked to a tracked item.** The PR body noted the deferral in prose only. **Resolution:** a follow-up issue has been filed for the PostgreSQL `occurred_at` write/read population (kept out of the public history per repo policy).

### LOW
- **L1 — [Correctness] Read-path style inconsistency.** `_row_to_chunk` uses the defensive `_parse_dt(_row_get("occurred_at"))` while `_temporal_row_to_chunk` uses `_parse_dt(row["occurred_at"])`. Both are NULL-safe; the difference is stylistic, not a bug. No change required.
- **L2 — [Devil's Advocate] Migration docstring scope.** The docstring justifies the column primarily in SQLite terms; a one-line note that PG write-population is *deferred, not eliminated* would aid future maintainers. Optional.
- **L3 — [Test] Optional PG-NULL guard.** A test asserting PG chunks currently read `occurred_at = NULL` was considered but not added (it would lock in a known-incomplete behavior); deferred to the follow-up instead.

### Security Audit
**PASS — no findings.** Scanned all 153 added lines across the 16 OWASP-aligned categories. The only SQL change is a static, fully parameterized `?`-placeholder INSERT (column-list/values alignment verified, 16 == 16); namespace scoping on the read paths is intact; no secrets, IDOR, SSRF, deserialization, file/path handling, or dependency changes. No auto-fixes were needed.

### Domain highlights
- **Correctness:** INSERT column-list ↔ values-tuple alignment verified (count and order); `_dt_to_str`-on-write / `_parse_dt`-on-read symmetry; COALESCE event-time semantics correct end-to-end; no behavior change to the temporal/`khora_chunks` path; no placeholders.
- **Database:** Nullable `ADD COLUMN` is metadata-only (no table rewrite / lock) on Postgres; `down_revision` chains correctly and `downgrade()` drops the column; ORM↔migration drift gate passes; dialect-portable.
- **Performance:** ~1–2 µs/chunk serialization overhead; no new indexes, query-plan changes, or N+1.
- **Test:** Regression guard is genuine (mutation-checked: a distinct in-range `occurred_at` with an out-of-range `source_timestamp` survives only because the column round-trips) — not a tautology. NULL/default, fallback, and negative cases all covered; migration count/head assertions updated.
- **Docs:** Updated module docstring and in-test comments accurately reflect the new behavior; no stale "embedded path drops occurred_at" claims remain.

**Overall: APPROVE.** The embedded bug fix is correct, well-tested, and surgical; the deferred PostgreSQL population is appropriately scoped and now tracked.

probablyrobot (approved):

PR #1058, opened 2026-06-09
Add recall-filter conformance corpus across fourteen filter families

filippetrovic:
## Summary

Authors the hand-written conformance case catalog into the recall-filter corpus, filling the family-generator stubs the harness shipped with empty bodies. The corpus proves every backend compiler returns the **same row set** as the in-memory Python oracle for the same filter.

- **`src/khora/filter/conformance.py`** — 202 hand-authored cases across 14 families: `f_op` (87), `f_logic` (18), `f_coerce` (13), `f_dates` (10), `f_nullval` (10), `f_polarity` (9), `f_array` (9), `f_sugar` (9), `f_objeq` (8), `f_exists` (8), `f_impossible` (7), `f_sel` (6), `f_dotkey` (5), `f_unsup` (3). `f_op_cases()` is reconciled to the richer 5-record-seed form (NULL row, empty-set `$in`/`$nin` variants, bare-list constant-false short-circuit, Mongo-faithful negation including the NULL/absent row). Every new generator is exported in `__all__`.
- **`tests/unit/filter/test_conformance_catalog.py`** (new) — runs every family through the Python oracle (asserting against hand-declared `expected_ids`, never oracle-derived) and, for chronicle-targeting cases, the Chronicle plan/run executor with agreement asserted.
- **`tests/unit/filter/test_exists_primitive.py`** (new) — asserts the **primitive** each compiler emits for `$exists`: SQLite `json_type` (not `json_extract`, which would collapse absent vs present-JSON-null), Postgres has-key `?` / nested-path existence (not `IS NOT NULL` on metadata), SurrealDB `IS NOT NONE`.
- **`tests/recall/test_filter_validator.py`** — validator coverage for out-of-scope operators (`$regex`/`$elemMatch`/`$size`/`$all`/`$type`/`$mod`/`$expr`/`$where`), typed-key operand type-gate raises, empty logical-array raises, and documented accept-not-raise guards (dotted-literal key, `$`-literal key, duplicate-key last-wins).

Every `expected_ids` is hand-authored from the seed and cross-checked by the oracle; the oracle stays falsifiable (the assertion target is always the declared set). The corpus introduces no compiler behavior changes — only test/corpus code.

## Authoring notes — corpus intent reconciled to shipped grammar

Where the working catalog's pre-implementation expectations diverged from the implemented validator/oracle, cases were authored to the **shipped** behavior and the divergence noted here (no production code was changed):

- Explicit `{"$eq": [list]}` lowers to array **containment**, while a **bare** list lowers to an exact-array operand — yet both canonicalize to the same `canonical_hash`. Same hash, different row sets: a latent cache-correctness smell worth a follow-up (cases authored on the bare-list exact-array form).
- A few "impossible" shapes actually raise at validation (e.g. a list operand under a scalar-typed `$eq`, a bare list on a date key, numeric `$in` elements against a string key) and are covered as validator raises rather than empty-result conformance cases.
- `{"$gte": null}` on a date key and a `$regex`-shaped object inside `$in` are **accepted** by the shipped validator (unset-bound sentinel / opaque exact-array literal) — covered as acceptance guards.
- The `$date` typed literal is a metadata-grammar form; system date keys take plain ISO strings.
- `$or`/`$in` and De Morgan pairs are row-equivalent but hash-distinct, so they assert row equality only; hash stability is asserted only on the true desugar identities.

## Test plan
- [x] Unit tests pass — filter + validator suite: 1316 passed
- [x] Conformance matrix (python leg): 87 passed
- [x] Lint clean (ruff check + ruff format + ty)
- [ ] Integration / live-backend conformance matrix (Postgres + Chronicle) — runs in the dedicated conformance CI job

### DIFF
--- a/.github/workflows/filter-conformance.yml
+++ b/.github/workflows/filter-conformance.yml
@@ -32,31 +32,50 @@ env:
 
 jobs:
   conformance:
+    # Per-leg check name: the three legs that actually execute and assert row-sets
+    # (the in-memory python oracle, live Postgres, and Chronicle) report as
+    # `conformance (<backend>)`; the four legs that today run ONLY the import/compile
+    # guard report as `compile-guard (<backend>)`. This keeps a green check from
+    # advertising backend conformance it did not run — a guard leg flips back to
+    # `conformance` only once it seeds a store and asserts row-set equality vs the
+    # python oracle.
+    name: ${{ matrix.kind }} (${{ matrix.backend }})
     runs-on: ubuntu-latest
     timeout-minutes: 25
     strategy:
       fail-fast: false
       matrix:
-        # `include` so each leg carries its own provisioning metadata:
+        # `include` so each leg carries its own provisioning + naming metadata:
+        #   kind:     `conformance` (executes + asserts row-sets) vs `compile-guard`
+        #             (import-only registry/compile check, no live execution yet).
         #   postgres: leg needs a live Postgres container + one-time seed.
-        # Legs without that flag run the import-only registry/compile guard,
+        # Legs without the postgres flag run the import-only registry/compile guard,
         # which needs no live database.
         include:
           # In-memory oracle — no service, no seed.
           - backend: python
+            kind: conformance
           # Live Postgres executor — the only leg that provisions a DB.
           - backend: postgres
             postgres: true
+            kind: conformance
           # Chronicle date-bound pushdown + in-memory post-filter — no DB.
           - backend: chronicle
+            kind: conformance
           # The four legs below currently run the registry/compile guard
           # (resolve the compiler from the registry + compile the AST, no live
-          # execution). They grow into live conformance execution once their
-          # harness executor lands in src/khora/filter/conformance.py.
+          # execution) — hence `kind: compile-guard`, so their green check does not
+          # claim executed backend conformance. They grow into live conformance
+          # execution (and flip to `kind: conformance`) once their harness executor
+          # lands in src/khora/filter/conformance.py.
           - backend: sqlite_lance
+            kind: compile-guard
           - backend: cypher
+            kind: compile-guard
           - backend: surrealdb
+            kind: compile-guard
           - backend: weaviate
+            kind: compile-guard
     env:
       KHORA_CONFORMANCE_BACKEND: ${{ matrix.backend }}
     steps:
--- a/src/khora/filter/conformance.py
+++ b/src/khora/filter/conformance.py
@@ -73,8 +73,20 @@ __all__ = [
     "PythonExecutor",
     "SeedRecord",
     "assert_case",
+    "f_array_cases",
+    "f_coerce_cases",
+    "f_dates_cases",
+    "f_dotkey_cases",
+    "f_exists_cases",
+    "f_impossible_cases",
+    "f_logic_cases",
+    "f_nullval_cases",
     "f_objeq_cases",
     "f_op_cases",
+    "f_polarity_cases",
+    "f_sel_cases",
+    "f_sugar_cases",
+    "f_unsup_cases",
     "oracle_survivors",
     "run_case_for_backend",
     "seed_case",
@@ -483,17 +495,15 @@ async def seed_case(coord: StorageCoordinator, case: ConformanceCase) -> dict[st
 # python-oracle cross-check can confirm — never define — them.
 
 
-# Fixed seed anchors for the date F-OP cases. ``_HIT`` is in range of every
-# generated date bound, ``_MISS`` out of range; the third record carries no date
-# at all (the absent-value record). UTC, matching the validator's normalization.
+# Fixed seed anchors for the date F-OP cases. ``_DATE_HIT`` is in range of every
+# generated date bound, ``_DATE_MISS`` out of range; ``_DATE_MID`` sits between the
+# two bounds so the boundary ops separate cleanly, and the ``-5`` record carries no
+# date at all (the NULL row, exercised by F1). UTC, matching the validator's
+# normalization. The bound ``_DATE_BOUND`` is below the hit and above the miss.
 _DATE_HIT = datetime(2026, 6, 1, tzinfo=UTC)
+_DATE_MID = datetime(2026, 3, 15, tzinfo=UTC)
 _DATE_MISS = datetime(2020, 1, 1, tzinfo=UTC)
 _DATE_BOUND = "2026-01-01T00:00:00Z"
-
-# The string F-OP cases compare against this value; ``_STR_OTHER`` is a distinct
-# value and the third record leaves the key unset (the absent-value record).
-_STR_HIT = "match-me"
-_STR_OTHER = "other-value"
 
 # All three backends are oracle-comparable for the F-OP family on the embedded
 # date/metadata surface; the catalog/CI tickets prune per-backend as needed (the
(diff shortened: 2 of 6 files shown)


filippetrovic:
## Consolidated Multi-Perspective Review — PR #1058

Seven independent reviewers (correctness, test-quality, database, security, performance, docs, devil's-advocate) plus a dedicated security audit. **Overall: APPROVE** — no blockers. One doc-accuracy fix was applied and pushed during review; remaining items are non-blocking hardening and one documented deferred follow-up.

### MEDIUM

- **M1 [Docs] — FIXED** `src/khora/filter/conformance.py` — the `f_op_cases()` docstring and the `_NON_NULL_STRING_KEY` comment described the non-null string key as staying consistent "across the in-memory oracle and live Postgres", but string-key F-OP cases run on python + chronicle only (`_STRING_OP_BACKENDS` prunes postgres, since the seven string document keys aren't denormalized onto the seeded chunk). Corrected to reflect the actual backend set and the forward-looking SQL-consistency rationale.
- **M2 [Devil's Advocate / Correctness] — DEFERRED (documented)** A bare list (exact-array `$eq`) and an explicit `{"$eq": [list]}` (array containment) compile to different predicates but canonicalize to the same `canonical_hash` (root: `_canonicalize_operand` treats tuple and list identically). Same hash, different rows is a latent cache-correctness hazard *if* a filter cache ever keys on the hash and both forms are mixed. This PR does not introduce the behavior — it authors the corpus on the bare-list form and surfaces the gap. Documented in the PR body and tracked as a follow-up; not fixed here because it lives in production AST code, outside this test/corpus PR's scope.
- **M3 [Test] — non-blocking hardening** `tests/unit/filter/test_exists_primitive.py` — the `$exists` primitive assertions use substring checks (`"json_type" in sql`, `"?" / "#>" in sql`, `"is not none"`); they could miss future compiler-output wrapping. The **actual emitted primitives were independently verified correct against the compiler source** by the database reviewer (Lance `json_type` not `json_extract`; Postgres `?`/`#> IS NOT NULL` not `->>`; SurrealDB `IS NOT NONE`), so this is assertion-tightening, not a correctness gap.
- **M4 [Test] — non-blocking hardening** `tests/recall/test_filter_validator.py` — the accept-not-raise guards (dotted-literal key, `$`-literal key, duplicate-key last-wins) assert acceptance but don't pair each with a near-variant that *does* raise. Adding the negative twin would make them stronger regression guards.
- **M5 [Test] — non-blocking hardening** `tests/unit/filter/test_conformance_catalog.py` — case-id uniqueness is keyed on `id` rather than `(family, id)`; the coerce/polarity shared-seed layout subtleties aren't asserted directly; metadata-operator coverage is a single representative case. All are precision improvements; correctness was hand-verified clean.

### LOW

- **L1 [Test]** `test_conformance_catalog.py` — the family-generator list is hand-maintained (must be updated when a new family is added) and the chronicle-agreement assertion message could include `case.filter` for better diagnostics.

### Clean dimensions

- **Correctness — CLEAN.** Hand-counted F-OP date/string, F-OBJEQ (superset excluded), dotkey-3 (whole-blob equality, descent excluded), F1 negation (NULL/absent included), F-EXISTS s4/s7 (present-JSON-null distinguished). Falsifiability intact — tests assert against hand-declared `expected_ids`, never oracle-derived.
- **Database — CLEAN.** Backend-pruning (string keys + `created_at` drop postgres), FK seeding via `ns.id`, and the three `$exists` primitives all verified against compiler source. No compiler behavior changed.
- **Security — CLEAN / Security Audit PASS.** Test/corpus-only diff: no injection/secret/deserialization/network/file sinks, no dependency or CI changes, no trust boundaries, and no internal identifiers leaked.
- **Performance — CLEAN.** Negligible cost (~<20ms collection, ~4s unit runtime); generators are pure with static seeds, no O(n²).
- **Devil's Advocate — SOUND.** The 202-vs-catalog reduction collapses near-duplicates only; every named F-LOGIC pattern + equivalence-pair (both spellings) is preserved; F-UNSUP/F-SEL cuts are harness-limitation-driven. The catalog-vs-shipped-grammar divergence calls (author to the implemented validator/oracle) are correct.

### Verdict: **APPROVE.** No blocking findings. M1 fixed and pushed; M2 deferred with documentation; M3–M5/L1 are optional test-precision hardening with the underlying behavior independently verified correct.

probablyrobot (approved):

filippetrovic:
The case authoring is thorough and the falsifiability discipline (assert against hand-authored `expected_ids`, never oracle output) is intact. F-VALIDATE is correctly placed in the validator suite as a single-run family rather than a backend-parametrized one — not a missing family.

The structural issue is that the corpus, as wired, does not execute cross-backend: 13 of the 14 families run only against `compile_python`.

**1. The live-DB matrix runner is hardcoded to one family.**
`tests/integration/matrix/test_filter_conformance.py:64` returns `list(f_op_cases())`, so against Postgres only F-OP runs. The 13 new families never reach a real database.

**2. Four backend legs report green without executing anything.**
The `cypher` / `surrealdb` / `weaviate` / `sqlite_lance` legs run an import/compile guard with no row-set execution, but surface as green checks named `conformance (cypher)` etc. A green check there is read as "backend conformance verified" when no rows were asserted.

**3. The Chronicle executor is `compile_python`.**
`src/khora/filter/execute.py:110` applies only `compile_python(...).predicate`, so `test_chronicle_agrees_with_python_oracle` compares `compile_python` against itself — no independent-backend signal.

Net effect: every known divergence class has an authored case — array-aware containment, object_equal exactness, `$exists` type-aware routing, `$ne`/`$nin` null-inclusion, impossible→false — but for the metadata-shaped families where these manifest, the case runs the oracle against itself. The PG `varchar = text[]`, jsonb `@>`-vs-`=`, and `->>`-vs-`json_type` divergences this corpus exists to catch are not executed against the compilers that diverge.

Required before the corpus can be relied on as the determinism guard:

- Replace the hardcoded `f_op_cases()` with the full family generator set so all 14 families execute against Postgres (the PG compiler/executor already exist).
- Rename or gate the four non-executing CI legs (e.g. `compile-guard (cypher)`) so a green check is not read as backend conformance until that leg executes row-sets.
- Build the per-backend executors (cypher / surrealdb / weaviate / sqlite_lance) that seed each store and assert row-set equality vs the python oracle.

Two smaller items:

- F-EXISTS dropped 16→8 cases, losing the per-backend primitive-routing rows (the `s1`–`s8` × pushdown-vs-post-filter matrix). Restore them as executed cases once a real metadata backend runs.
- The bare-list vs explicit-`$eq`-list canonicalization (same filter hash, different row semantics) cache hazard noted in review needs a tracked follow-up; it lives in the production AST canonicalizer, not in this corpus.

If P1-c is intentionally scoped to "author the cases + the F-OP-on-Postgres leg," with the backend executor wiring as a separate downstream change, that is a defensible slice — but items 1 and 2 above should still land or be gated here, because the misleading-green legs ship false confidence into the merge queue in the meantime.

filippetrovic:
Requesting one change before merge.

The `cypher` / `surrealdb` / `weaviate` / `sqlite_lance` legs of `filter-conformance.yml` run an import/compile guard with no row-set execution, but surface as green checks named `conformance (cypher)` etc. As-is, merging this makes the merge queue read "backend conformance verified" for four backends that asserted no rows.

Before merge: gate or rename those four legs so a green check is not mistaken for executed backend coverage — e.g. rename to `compile-guard (<backend>)`, or mark them non-required — and flip back to `conformance (<backend>)` only when the leg seeds a store and asserts row-set equality vs the python oracle.

Building those executors, and wiring the full family set into the Postgres leg (still F-OP-only at `tests/integration/matrix/test_filter_conformance.py:64`), is fine to defer to a separate follow-up — not required here. The only change requested for this PR is that no CI leg advertise backend conformance it did not run.

filippetrovic:
Addressed the required change in `fa04f74`.

**Done — no CI leg advertises backend conformance it did not run:**
- `.github/workflows/filter-conformance.yml` — each matrix leg now carries an explicit `kind`, and the job display name is `${{ matrix.kind }} (${{ matrix.backend }})`. The four guard-only legs report as **`compile-guard (cypher)` / `compile-guard (surrealdb)` / `compile-guard (weaviate)` / `compile-guard (sqlite_lance)`**; the three that actually seed/compile and assert row-sets keep **`conformance (python)` / `conformance (postgres)` / `conformance (chronicle)`**. A guard leg flips back to `conformance` only when it seeds a store and asserts row-set equality vs the python oracle.
- `tests/integration/matrix/test_filter_conformance.py` — corrected the now-stale `_all_cases()` docstring (it claimed the catalog families "return `[]` until filled"; this PR filled them). It now states plainly that the metadata-shaped families are authored + validated against the python oracle + Chronicle in the unit suite, and are deliberately kept off the live-store leg until the known `compile_postgres` metadata divergences are resolved.

**Deferred to follow-up (per your second comment — "fine to defer"):**
- Wiring the full 14-family set into the live Postgres leg. Confirming your read: those metadata families exercise `compile_postgres` paths with **known open divergences** — sub-path object-equality (`@>` vs exact `=`), array-valued containment, and nested `jsonb #> varchar` typing — already tracked as separate issues. Running them against live Postgres today would fail on those compiler bugs, not on the corpus, so they stay oracle/Chronicle-validated until those land.
- The per-backend executors (cypher / surrealdb / weaviate / sqlite_lance) that seed each store and assert row-set equality.
- Restoring the F-EXISTS `s1`–`s8` × pushdown-vs-post-filter primitive-routing rows as **executed** cases — depends on a real metadata backend executor, so it rides with the executor work above.
- The bare-list vs explicit-`$eq`-list canonicalization cache hazard (production AST `_canonicalize_operand`, not this corpus) — tracked as a follow-up.

The CI run on `fa04f74` will show the renamed legs.

PR #1061, opened 2026-06-09
Normalize tz-naive datetimes at the recall-filter comparison boundary

filippetrovic:
## Summary
- System date-key recall filters (`occurred_at` / `created_at` / `source_timestamp`) raised `TypeError: can't compare offset-naive and offset-aware datetimes` instead of narrowing when the stored datetime came back tz-naive (e.g. the embedded SQLite store) and the filter operand was tz-aware. This violated the compiler's contract of excluding incomparable values rather than raising.
- Added `_to_utc` / `_align_dt` helpers and applied them to the **values actually compared** in the shared system-key ops (`_system_eq` / `_system_ne` / `_system_in` / `_system_nin` / `_system_range`), mirroring the existing metadata-date normalization. Naive datetimes are read as UTC; aware datetimes are converted to UTC, so instants compare correctly regardless of the source offset.
- The type gate (`_comparable`) is unchanged — genuinely incomparable cross-family pairs (e.g. `str` vs `datetime`) are still excluded without raising (no behavior change on the previously-working aware/aware path). The fix lives in the shared post-filter, so every backend that can return naive stored datetimes is covered, not just the embedded store.
- Dropped the strict `xfail` that pinned the crash; that test now passes as a plain guard. Added regression coverage across all three system date keys on the real embedded store, plus unit coverage for naive/aware/mixed-offset comparisons and the incomparable-cross-family exclusion rule.

## Known limitation (pre-existing, out of scope)
- On the embedded store, a `source_timestamp` **upper bound** (`$lte` / range) does not narrow: the recency window pre-filters on the ingest-time `created_at` column, so an upper bound before "now" clamps the candidate set. The lower bound (`$gte`) narrows correctly. This is a separate recency-pushdown characteristic, not the datetime-comparison fix here; the regression test scopes `source_timestamp` accordingly and documents it.

## Test plan
- [x] Unit tests pass (`tests/unit/filter`, `tests/recall/test_compile_python.py`)
- [x] Integration tests pass (`tests/recall/test_filter_chronicle.py` on the real embedded sqlite_lance store)
- [x] Cross-backend filter conformance green (python-oracle / chronicle / live postgres)
- [x] `make format` + `make lint` clean
- [ ] Full CI suite green

### DIFF
--- a/src/khora/filter/compilers/python.py
+++ b/src/khora/filter/compilers/python.py
@@ -56,7 +56,7 @@ khora's own engines; not re-exported from :mod:`khora.__init__`.
 from __future__ import annotations
 
 from collections.abc import Callable, Mapping
-from datetime import datetime
+from datetime import UTC, datetime
 from typing import Any
 
 from khora.filter import (
@@ -437,6 +437,14 @@ def _comparable(stored: Any, operand: Any) -> bool:
     strings with strings, datetimes with datetimes. Any cross-family pair is
     NOT comparable — the positive op excludes it (Rule 1: never abort, exclude
     instead of raising a ``TypeError`` on ``str < int``).
+
+    Two datetimes are comparable regardless of tz-awareness: a naive stored
+    value (some backends — e.g. the embedded sqlite store, whose SQLAlchemy
+    ``DateTime`` column is tz-naive — return naive datetimes) is normalized to
+    UTC at the comparison boundary by :func:`_align_dt`, so the pair both
+    compares and orders without a ``TypeError``. The gate stays a same-family
+    test; the tz alignment is applied to the *values* before ``==`` / ``<`` /
+    membership, never here.
     """
     if _is_number(operand):
         return _is_number(stored)
@@ -451,6 +459,27 @@ def _comparable(stored: Any, operand: Any) -> bool:
     return type(stored) is type(operand)
 
 
+def _to_utc(value: datetime) -> datetime:
+    """Normalize a datetime to UTC-aware — naive is read as UTC (mirrors :func:`_try_parse_utc`)."""
+    if value.tzinfo is None:
+        return value.replace(tzinfo=UTC)
+    return value.astimezone(UTC)
+
+
+def _align_dt(stored: Any, value: Any) -> tuple[Any, Any]:
+    """Align a comparable datetime pair to UTC-aware so ``==`` / ``<`` never raise.
+
+    Only rewrites when BOTH operands are datetimes (the sole family where a
+    naive/aware mix raises ``TypeError`` on compare). A naive operand on either
+    side — stored (e.g. the embedded sqlite store) or the filter value — is read
+    as UTC, mirroring :func:`_try_parse_utc` / :func:`_md_date_compare`. Any
+    other pair (already-handled by :func:`_comparable`) passes through unchanged.
+    """
+    if isinstance(stored, datetime) and isinstance(value, datetime):
+        return _to_utc(stored), _to_utc(value)
+    return stored, value
+
+
 # ----- system-key value comparisons (None / wrong-type aware) -------------- #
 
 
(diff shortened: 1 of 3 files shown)


filippetrovic:
## 🧑‍⚖️ Multi-perspective review — **APPROVE**

A 7-reviewer panel (security, performance, correctness, test, database/cross-backend, docs, devil's advocate) plus a dedicated security audit reviewed this change. No blocking defects: correctness, security, and cross-backend result-set parity are all confirmed. Remaining items are non-blocking follow-ups.

**Panel verdicts:** Correctness ✅ · Security ✅ none · Cross-backend parity ✅ · Tests ✅ · Docs ✅ clean · Performance ⚠️ MEDIUM/LOW · Devil's Advocate ⚠️ seam dissent (lead: non-blocking) · Security Audit ✅ PASS

### CRITICAL / HIGH
None.

### MEDIUM
- **M1 — [Performance]** `src/khora/filter/compilers/python.py` (`_to_utc`/`_align_dt`, applied in `_system_eq`/`_ne`/`_in`/`_nin`/`_range`). Alignment runs **per candidate row** on the post-filter hot path: two `isinstance` checks + `astimezone` per comparison, with extra tuple/datetime allocation in `$in`/`$nin` via `_eq_aligned` (once per list member per row). Functionally correct (no regression across 643 affected-suite tests + conformance), but the filter operands are fixed and known at predicate-compile time, so the operand-side alignment can be **hoisted out of the per-row loop**. Recommended follow-up: pre-align operands when the predicate closure is built + an aware-value fast-path (`value.tzinfo is UTC → return as-is`). Not a merge blocker.
- **M2 — [Design / Devil's Advocate]** Seam choice. The devil's advocate argues the fix belongs in the embedded reader (`storage/backends/sqlite_lance/vector.py::_parse_dt`, return tz-aware) rather than the shared Python compiler, to avoid a global "naive ⇒ UTC" policy and the (hypothetical) future backend that stores naive *local* times. **Lead assessment: not a blocker.** (1) The change's stated design intent deliberately prefers the compiler seam so *every* backend that can emit naive datetimes is covered, not just the embedded one. (2) The "naive ⇒ UTC" convention is **not new** — the metadata-date path (`_try_parse_utc`) already assumes it. (3) Cross-backend parity was independently verified by the database reviewer: postgres `TIMESTAMPTZ` (session TZ=UTC) and weaviate ISO-8601 string binding both agree with the python oracle's normalized semantics. Hardening `_parse_dt` to return aware is welcome as *additive* defense-in-depth, but is out of scope here. The future-naive-local concern is worth a docstring note, not a redesign.

### LOW
- **L1 — [Docs]** Document the `source_timestamp` upper-bound (`$lte`/range) interaction with the recency window (the pre-filter screens on ingest-time `created_at`, so an upper bound before "now" clamps the candidate set independently of `source_timestamp`) in the engine docs. Already honestly scoped in the test docstring; surfacing it user-side would prevent surprise. (`$gte` lower bounds narrow correctly.)
- **L2 — [Test]** Optional perf-regression guard for large-candidate-set date filters (ties to M1).
- **L3 — [Cleanup]** `_try_parse_utc` re-imports `UTC` locally though it's already a module import — harmless, follow-up tidy.
- **L4 — [Test]** Coverage is already asymmetry-complete; an explicit `$ne`-on-range case and an opposite-sign offset (e.g. `-05:00`) would round it out.

### Security Audit (deep pass)
**PASS.** Added lines scanned across all 16 categories (injection, authn/z, crypto/secrets, access control, SSRF, deserialization, file/path, redirects, mass-assignment, ReDoS, supply chain, boundary validation, …). The diff is pure datetime-comparison logic + tests — no untrusted-input→sink paths, no secrets, no dependency/lockfile changes, stateless pure functions with no tenant-isolation surface. No findings; nothing to auto-fix.

### Why APPROVE
The fix is correct across naive/aware/mixed-offset inputs, preserves the compiler's "exclude, never raise" rule for incomparable cross-family pairs, normalizes the **values actually compared** (not merely the type gate) in all five system ops, and is backed by comprehensive regressions on the real embedded store plus cross-backend conformance. M1 (perf hoist) and M2 (seam debate) are worthwhile follow-ups, not blockers.

probablyrobot (approved):

PR #1062, opened 2026-06-09
Persist and read occurred_at on the PostgreSQL pgvector chunk path

filippetrovic:
## Summary
- The PostgreSQL chunk write/read mappers handled `source_timestamp`, `last_accessed_at`, and `session_id` but omitted `occurred_at`, so chunks written through the pgvector backend always read back `occurred_at = NULL`. A distinct event time could not be honored there — it silently fell back to `source_timestamp` via `COALESCE(occurred_at, source_timestamp)`.
- Populate `occurred_at` in both write paths (single + batch `ChunkModel` construction) and read it back in the `ChunkModel → Chunk` mapper, mirroring the adjacent `source_timestamp` handling. The column, ORM mapping, domain field, and migration already existed; this only wires the relational/pgvector backend to use them (+3 lines).
- Add an end-to-end PG/pgvector test module mirroring the existing embedded sibling: same engine, same filter AST, same recall path — only the storage backend differs.

## Test plan
- [ ] Unit tests pass
- [x] Unit/recall suite passes locally (the new module self-skips when Postgres is unreachable)
- [ ] Integration tests pass (run against live Postgres in CI):
  - `test_occurred_at_round_trips_through_real_store` — a chunk written with an `occurred_at` distinct from both `created_at` and `source_timestamp` reads back with that exact value (fails loudly if the write path drops it).
  - `test_occurred_at_filter_honored_over_out_of_range_source_timestamp` — a chunk whose `occurred_at` is in range but whose `source_timestamp` is out of range is honored by an `occurred_at` recall filter, proving the value is genuinely persisted rather than recovered via `COALESCE`.
- [x] `make format` clean; `make lint` exits 0; change is surgical (+3 production lines)

## Notes
- The SurrealDB chunk backend has the same gap but its `chunk` table is schemafull (silently strips undeclared fields), so a correct fix there also needs a schema `DEFINE FIELD` change. Deferred to a separate follow-up to keep this change surgical; it does not regress today (the event-time post-filter recovers via `source_timestamp`).

### DIFF
--- a/src/khora/storage/backends/pgvector.py
+++ b/src/khora/storage/backends/pgvector.py
@@ -388,6 +388,7 @@ class PgVectorBackend(AsyncSessionMixin):
             created_at=chunk.created_at,
             source_timestamp=getattr(chunk, "source_timestamp", None),
             last_accessed_at=getattr(chunk, "last_accessed_at", None),
+            occurred_at=getattr(chunk, "occurred_at", None),
             session_id=getattr(chunk, "session_id", None),
         )
         session.add(model)
@@ -428,6 +429,7 @@ class PgVectorBackend(AsyncSessionMixin):
                 created_at=chunk.created_at,
                 source_timestamp=getattr(chunk, "source_timestamp", None),
                 last_accessed_at=getattr(chunk, "last_accessed_at", None),
+                occurred_at=getattr(chunk, "occurred_at", None),
                 session_id=getattr(chunk, "session_id", None),
             )
             for chunk in chunks
@@ -795,6 +797,7 @@ class PgVectorBackend(AsyncSessionMixin):
             created_at=model.created_at,
             source_timestamp=getattr(model, "source_timestamp", None),
             last_accessed_at=getattr(model, "last_accessed_at", None),
+            occurred_at=getattr(model, "occurred_at", None),
             session_id=getattr(model, "session_id", None),
         )
 
--- a/tests/integration/test_chronicle_filter_pgvector.py
+++ b/tests/integration/test_chronicle_filter_pgvector.py
@@ -1,0 +1,335 @@
+"""End-to-end occurred_at persistence + recall-filter tests on the real PG/pgvector path.
+
+The PostgreSQL ``chunks`` table carries a distinct ``occurred_at`` TIMESTAMPTZ column
+(the real-world event time the chunk's content refers to, separate from both
+``created_at`` ingestion time and ``source_timestamp``). These tests drive that column
+through the production ``PgVectorBackend`` write/read path against a live, fully-migrated
+Postgres — no storage mocks — and prove two halves of the contract:
+
+* **Round-trip** — a chunk written with an ``occurred_at`` distinct from both
+  ``created_at`` and ``source_timestamp`` reads back with that exact value. This fails
+  loudly if the write path drops ``occurred_at`` (it would read back ``NULL``).
+* **Recall-filter regression guard** — a chunk whose ``occurred_at`` is in range but
+  whose ``source_timestamp`` is out of range is honored by an ``occurred_at`` recall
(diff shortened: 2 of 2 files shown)


filippetrovic:
## Multi-Perspective Review — PR #1062

Reviewers: Security · Performance · Correctness · Test · Database · Docs · Devil's Advocate · Security Audit

**Overall: ✅ APPROVE** — a surgical, well-tested parity fix. No CRITICAL/HIGH findings on the changed lines. One pre-existing cross-backend gap is noted as a recommended follow-up (out of scope here).

### Scope
Wires `occurred_at` through the PostgreSQL pgvector chunk write/read mappers (+3 production lines) — single-chunk write, batch write, and the `ChunkModel → Chunk` read mapper — mirroring the adjacent `source_timestamp` handling. The column, ORM mapping, domain field, and migration already exist; this only connects them. Adds a two-test PG integration module (distinct-value round-trip + recall-filter regression guard).

### Findings

**MEDIUM**
- **M1 — Cross-backend parity (Devil's Advocate).** The SurrealDB vector backend still does not persist/read `occurred_at`, so a distinct `occurred_at` is dropped there. This is **pre-existing** (SurrealDB never persisted it) and **not introduced by this PR** — this change actually *reduces* divergence by bringing PostgreSQL in line with the already-fixed embedded backend. SurrealDB's `chunk` table is schemafull (silently strips undeclared fields), so a correct fix also needs a `DEFINE FIELD occurred_at` schema change plus several code sites — materially larger than this change and outside its stated scope. **Recommendation:** track in a dedicated follow-up issue (consider folding in `last_accessed_at`, which the SurrealDB chunk path also omits). Non-blocking here; no regression today, since recall recovers via `COALESCE(occurred_at, source_timestamp)`.

**LOW**
- **L1 — Style (Devil's Advocate).** The three new lines use `getattr(chunk, "occurred_at", None)` where direct attribute access would also be safe (the `Chunk` dataclass always declares the field). This is intentionally **consistent** with the adjacent `source_timestamp` / `last_accessed_at` / `session_id` lines, so it is uniform rather than a defect. No change required.

**INFO**
- **I1 — Injection (Security Audit).** The new test module's `_reset_public_schema` teardown builds a `DROP TYPE … {typname}` statement via f-string. `typname` is read from the database's own `pg_type` catalog (not untrusted input), in test-only teardown against a local throwaway database. No trust boundary, not exploitable. Noted for completeness.

### Per-domain verdicts
- 🔒 **Security — PASS / CLEAN.** Read paths are namespace-scoped before mapping (no IDOR); the field is ORM-parameterized; no secrets involved.
- ⚡ **Performance — CLEAN.** O(1) per-chunk attribute copy; no extra round-trips or N+1; existing `(namespace_id, occurred_at)` index covers range filters.
- ✅ **Correctness — PASS.** Populated at both write sites + the read mapper; tz-aware (`DateTime(timezone=True)`); `COALESCE` event-time semantics preserved (no false-empty when unset).
- 🧪 **Tests — PASS (strong).** Round-trip uses a genuinely 3-way-distinct `occurred_at`; the filter test fails if `occurred_at` is dropped (COALESCE would fall back to the out-of-range `source_timestamp`); tz-consistent; correct reachability skip-guard; isolated module/function fixtures.
- 🗄 **Database — APPROVED.** Column ↔ mapper aligned; no new migration needed; no schema drift; correct tz handling at the SQLAlchemy boundary.
- 📚 **Docs — CLEAN.** Internal wiring fix; no public API change, no CHANGELOG/docs drift; docstrings accurate and free of internal identifiers.
- 🛡 **Security Audit — PASS.** No dependency/manifest changes; no injection/secret/SSRF/deserialization surface on the changed lines.

### Verdict: ✅ APPROVE
Ready to merge once CI is green. Recommend opening a follow-up issue for SurrealDB `occurred_at` persistence (M1).

probablyrobot (approved):

probablyrobot (approved):

PR #1064, opened 2026-06-09
Add VectorCypher recall-filter integration tests

filippetrovic:
## Summary
- Add `tests/recall/test_filter_vectorcypher.py`, the VectorCypher counterpart to the existing V1 Skeleton-pgvector recall-filter suite.
- **Cross-engine equivalence (live PG + Neo4j):** reuse the identical 5-in-scope / 5-out-of-scope corpus and three-predicate filter (`source_name` + `occurred_at >=` + `metadata.tag IN`) so the VectorCypher result set is pinned to exactly match the Skeleton-pgvector result set for the shared corpus. Includes a no-filter control proving the narrowing is attributable to the filter alone.
- **Engine-info carrier + validation (live):** asserts the honest VectorCypher filter carrier (`engine="vectorcypher"`, `supported=False`, `pushed_down=False`) and that an invalid filter raises `RecallFilterValidationError` through the facade.
- **Backward-compat (live):** the deprecated `start_time`/`end_time` bounds filter rows end-to-end, and supplying both `filter=` and the bounds raises.
- **Filter telemetry (hermetic, runs everywhere):** `under_filled` fires when a filtered recall returns fewer chunks than the requested limit; `graph_channel_empty` fires when the metadata post-filter empties a genuinely non-empty graph channel — proven both on a real-graph path (real `MENTIONED_IN` edges, with a pre-flight non-vacuity guard) and via a mocked-channel path, each with a silent-survivor control.
- **Capability boundary (hermetic):** an unsupported Cypher compilation propagates `RecallFilterUnsupportedError` out of `retrieve()` (proven to *not* degrade to the vector-only fallback), while a transient Neo4j error degrades to vector-only and still returns chunks.

The live (real Postgres + Neo4j) scenarios self-skip cleanly when either backend is unreachable, mirroring the V1 interim-test pattern; the hermetic telemetry/boundary tests run in every job. No production source changed — this is a test-only addition.

## Test plan
- [x] `ruff check` + `ruff format --check` clean
- [x] `make lint` (ruff + ty + optional-imports) passes
- [x] New file: `uv run pytest tests/recall/test_filter_vectorcypher.py` → 6 passed, 7 skipped (live scenarios self-skip without the local stack)
- [x] Unit + recall suites pass serially (`tests/recall/` → 432 passed, 15 skipped)
- [ ] CI: full suite incl. integration job (real Postgres + Neo4j) — the live scenarios run there as the permanent gate


probablyrobot (approved):

filippetrovic:
## Multi-perspective review — test-only change (1 file, +1041)

Six independent reviewer lenses (security, performance, correctness, test-quality, docs, devil's-advocate) + a security audit. **Overall assessment: APPROVE — no blocking findings.**

### CRITICAL
_None._

### HIGH
_None._ (Two HIGH candidates from the test-quality lens were re-rated on synthesis — see notes.)

### MEDIUM
- **M1 — [devil's-advocate / test]** The 7 live (`@_LIVE_DB`) tests self-skip in the main test job and gate only in the CI integration job (which provisions Postgres + Neo4j). This is intentional and mirrors the existing V1 interim-test pattern, but "7 skipped" in the main job can read as missing coverage at a glance. The integration job is the real gate; the conformance suite is the future permanent home (per the module docstring). No code change — documented here for reviewers.

### LOW
- **L1 — [test]** `graph_channel_empty` non-vacuity is proven via an unfiltered pre-flight asserting `engine_info["graph_chunk_count"] > 0` rather than a direct Neo4j `MENTIONED_IN` edge count. `graph_chunk_count` is the live graph channel's returned-chunk count, so `>0` does imply edges existed; a direct edge query would be more explicit but couples the test to graph-backend internals. The hermetic counterpart already proves the counter-fire seam deterministically. Optional enhancement, not a blocker. (The correctness lens independently rated this path CORRECT / non-vacuous.)
- **L2 — [test]** Counter assertions use exact `counter.adds == [(1, {})]`. This matches the established pattern in `tests/unit/filter/test_filter_telemetry.py` and is precise-by-design — the record helpers call `.add(1)` with no attributes, so a future attribute addition is a contract change that *should* update the test. Keep as-is.
- **L3 — [test]** The `_vector_only_fallback.assert_not_awaited()` spy asserts on a private attribute. `pytest.raises` alone proves propagation; the spy is belt-and-suspenders that explicitly documents "no fallback was taken" and guards a catch-then-fallback refactor. Harmless — keep.
- **L4 — [devil's-advocate]** The module docstring's "any divergence is an engine difference" framing is marginally stronger than identical-corpus can prove (VectorCypher adds graph/BM25/extraction channels Skeleton lacks). The docstring already labels this an INTERIM smoke test; no rewrite needed.

### Synthesis notes (re-rated findings)
- The test lens's HIGH **"`compile_cypher` monkeypatch binding may go stale (module-imported)"** is **not valid**: `retriever.py` imports `compile_cypher` *inside* `retrieve()` (≈line 1084), not at module top, so patching `khora.filter.compilers.cypher.compile_cypher` resolves fresh on each call. Verified working — the test raises `RecallFilterUnsupportedError` as asserted. → resolved/invalid.
- The test lens's two CRITICALs (non-vacuity, brittle `.adds`) are re-rated **LOW** (L1, L2); the correctness lens independently validated both paths as correct.

### Security Audit
Test-only diff: no new dependencies, no cryptography, no endpoints/auth, no secrets committed. The raw SQL in `_seed` is parameterized (`WHERE namespace_id = :ns` bind). The fixture's `neo4j`/`password` values are env-overridable local-compose defaults, test-scoped behind `@pytest.mark.integration` + a TCP reachability skip — not production credentials and unable to leak externally. No OWASP Top 10 surface introduced. **No findings.**

---
All four acceptance criteria are covered (cross-engine equivalence vs V1, `under_filled` + `graph_channel_empty` counters, partial-failure-raises vs transient-fallback). 6 hermetic tests run everywhere; 7 live tests gate in the integration job. Approve.

filippetrovic:
The live tests in this file do not execute in any CI job, and the row-set-parity check is in the wrong place architecturally. Recommend re-scoping rather than wiring it as-is.

**1. The live scenarios run in no CI job.**
`tests/recall/` is collected only by `test-unit`, which provisions no Postgres or Neo4j — so the seven `@_LIVE_DB` scenarios `skipif` out silently. The `test-integration` job (which does provision both databases and carries the loud `KHORA_PG_REQUIRED` / `NEO4J_INTEGRATION_TEST` guards via `tests/integration/conftest.py`) collects only `tests/integration/`, not `tests/recall/`. So the headline assertion — VectorCypher returns the same row set as the pgvector baseline — never runs; the PR is green on the mocked tests alone (`test-unit … 50 skipped` includes these).

**2. The parity check belongs in the conformance corpus, not a standalone file.**
The per-engine "same row IDs as the `compile_python` oracle for the same filter" guarantee is the conformance corpus's job: `tests/integration/matrix/`, the `filter_conformance` marker, executed by the dedicated `filter-conformance` job as the `postgres` and `cypher` matrix legs over the shared case set. Those `cypher` / `surrealdb` / `weaviate` / `sqlite_lance` legs are currently `compile-guard` stubs (compile-only, no row-set execution). Wiring them to seed each store and assert row-set equality against the oracle is where this parity belongs — and it yields one shared expected set instead of a second hand-authored `_IN_SCOPE` that can drift from the pgvector one.

**3. Scope distinction.**
The conformance corpus is isolated filter-predicate execution — seed rows, run the compiled `WHERE`, assert which survive — deliberately without embeddings, ranking, RRF, or engine orchestration. The full-engine `kb.recall()` tests in this file (`graph_channel_empty` / `under_filled` counter assertions, compile-error propagation, transient-Neo4j fallback) are a different kind of test: white-box pushdown/enforcement spies that assert the filter *reaches* each retrieval channel unchanged. Those belong in the enforcement-spy suite, not bundled with the row-set parity check.

**Suggested re-scope:**
- Deliver the VectorCypher row-set parity as the `postgres` + `cypher` legs of the conformance corpus (`tests/integration/matrix/`, `filter_conformance`) so it runs in the `filter-conformance` job against the shared oracle.
- Move the counter / transient-fallback / compile-error tests into the pushdown-enforcement spy suite (the white-box "filter carried to each channel" tests).
- Drop the standalone `tests/recall/test_filter_vectorcypher.py` parity file. If an interim full-engine test is kept, relocate it under `tests/integration/` and add it to the `test-integration` invocation so it actually executes against both databases under the loud-skip guard.

filippetrovic:
Re-scoped per the review feedback — thank you, the live-tests-run-in-no-CI-job point was correct. The PR now delivers exactly the one gap nothing else covers (that the two recall-filter service counters fire on their conditions) and drops everything redundant.

**Removed**
- `tests/recall/test_filter_vectorcypher.py` (deleted). Its row-set parity duplicated the conformance `postgres` leg; its pushdown / partial-failure spies duplicated the pushdown-spy suite; and its live scenarios sat in `tests/recall/`, which only the no-DB `test-unit` job collects — so they skipped silently and gated nothing.

**Added**
- `tests/unit/engines/vectorcypher/test_filter_counters.py` — hermetic spies that `graph_channel_empty` and `under_filled` fire on their declared call-site conditions, each paired with a silent-control vacuity guard. Runs in `test-unit`.
- `tests/integration/test_vectorcypher_filter_counters.py` — real PG+Neo4j proof that `graph_channel_empty` fires when a metadata filter empties a *genuine, non-empty* graph channel (entity-seeded `MENTIONED_IN` edges, `SearchMode.GRAPH`, `min_entity_similarity=0.0`, pre-filter `graph_chunk_count > 0` assertion). Carries `integration` + `filter_enforcement`.
- Registered the `filter_enforcement` marker.

**Proof it executes in CI (not silently skipped)** — from the `test-integration` job log:
```
collected 437 items / 101 deselected / 336 selected
tests/integration/test_vectorcypher_filter_counters.py .
= 326 passed, 6 skipped, 101 deselected, 4 xfailed ... in 343.57s =
```
The `.` (not `s`) confirms the live test ran and passed against the provisioned Postgres + Neo4j (the job sets `NEO4J_INTEGRATION_TEST=1` / `KHORA_PG_REQUIRED=1` with the loud conftest abort). All checks are green.

Row-set parity (the postgres + cypher conformance legs) and the channel-threading / partial-failure spies are intentionally left to their own work and not re-added here.

PR #1065, opened 2026-06-09
Hydrate document keys on the Chronicle recall path

filippetrovic:
## Summary
- Hydrate the denormalized document keys (`source_name`, `source_type`, `source_url`, `external_id`, `content_type`, `source`, `title`) on the Chronicle recall path so recall-filter predicates on those keys resolve against the live post-filter record instead of resolving as absent (which previously made any positive document-key predicate match nothing).
- The recall path now batch-fetches the per-document projections for all candidate chunks in a single call, **gated** on the filter actually constraining a document key (`filter_leaf_keys(...) ∩` the document-key set). Recalls with no filter, or a recency-only / event-time / metadata-only filter, short-circuit and pay zero extra query — the hot path is unaffected. No N+1.
- The record builder resolves each document key `chunk → hydrated projection → source_document → absent`, keeping the existing fallback as a tertiary so projection-less paths don't regress. `DocumentSource` is untouched and the record builder stays engine-private.
- A projection fetch failure degrades gracefully: it logs a warning, records an observability degradation on the recall result, and continues with the keys absent so recall never crashes. A missing/orphaned single document simply omits that id from the batch.
- Strengthen the embedded recall-filter integration test to assert the full three-predicate motivating filter (`source_name` + `occurred_at` range + `metadata.tag`) over the real system keys, replacing the earlier metadata stand-ins. Adds positive/sentinel pairs that prove the keys hydrate to real values (not "absent → matches everything"), a previously-absent-key (`external_id`) regression, a happy-path no-degradation assertion, and a short-circuit no-op assertion.

## Test plan
- [x] Unit tests pass (`tests/recall/` — 429 passed, 8 skipped)
- [x] Integration tests pass (embedded recall-filter gate — composition + live-store suites green)
- [x] Lint/format clean (`make format`, `make lint`)
- [ ] CI green on all required checks

### DIFF
--- a/docs/telemetry-contract.json
+++ b/docs/telemetry-contract.json
@@ -1005,8 +1005,8 @@
       "unit": "1",
       "owner": "chronicle",
       "labels": [
-        {"name": "channel", "values": ["bm25", "semantic", "temporal", "entity", "temporal_channel"], "_comment": "Trailing token of the component name."},
-        {"name": "reason", "values": ["fulltext_backend_unavailable", "channel_exception", "events_query_failed", "cosine_batch_failed", "chunk_fetch_failed", "chunk_fallback_failed", "events_no_usable_signal"], "_comment": "Low-cardinality enum; extend when adding new fallback paths."}
+        {"name": "channel", "values": ["bm25", "semantic", "temporal", "entity", "temporal_channel", "doc_hydration"], "_comment": "Trailing token of the component name."},
+        {"name": "reason", "values": ["fulltext_backend_unavailable", "channel_exception", "events_query_failed", "cosine_batch_failed", "chunk_fetch_failed", "chunk_fallback_failed", "events_no_usable_signal", "projection_fetch_failed"], "_comment": "Low-cardinality enum; extend when adding new fallback paths."}
       ],
       "description": "Issues #901, #906. Chronicle retrieval channel degradations recorded under ADR-001 (silent-failure observability). Incremented once per channel-failure or silent-fallback site in chronicle.recall(). The same event is also appended to RecallResult.engine_info['degradations']. NO namespace_id label - cardinality rule. See docs/architecture/failure-observability-contract.md."
     },
--- a/src/khora/engines/chronicle/engine.py
+++ b/src/khora/engines/chronicle/engine.py
@@ -381,11 +381,13 @@ def _intersect_upper(window: datetime | None, filter_bound: datetime | None) -> 
     return min(window, fb)
 
 
-# System keys the recall-filter post-filter reads off a chunk record. The eight
-# denormalized document keys live on the chunk's ``source_document`` projection
-# when present; ``DocumentSource`` only carries a subset (title / source /
-# source_type), so the rest resolve to absent on the legacy ``chunks`` path —
-# a documented limitation (positive predicates on them return empty there).
+# System keys the recall-filter post-filter reads off a chunk record. All seven
+# denormalized document keys live on the per-document ``DocumentProjection`` the
+# recall path hydrates for filtered queries (``get_document_projections_batch``);
+# ``DocumentSource`` carries only a subset (title / source / source_type), kept as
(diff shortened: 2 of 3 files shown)


filippetrovic:
## Multi-Perspective Review — Document-Key Hydration on the Chronicle Recall Path

Reviewed by an 8-role review team (security, performance, correctness, test, database, observability, docs, devil's advocate) plus a dedicated security audit. Diff: 3 files, +334 / −96 (engine `+78`, tests `+271/−77`, telemetry contract `+1/−1` per side).

No CRITICAL or HIGH blocking findings. The one contract-accuracy gap surfaced during review was fixed in this PR (the degradation channel/reason enums were extended). Remaining items are LOW/defense-in-depth and either pre-existing repo-wide patterns or negligible micro-optimizations.

---

### CRITICAL
- None.

### HIGH
- None.

### MEDIUM
- **[Observability] — RESOLVED in this PR.** The new degradation path introduced a `doc_hydration` channel and a `projection_fetch_failed` reason on the existing `khora.chronicle.channel.degraded_total` counter. The contract-drift gate validates metric *names* (not label values), so CI was not blocked, but the telemetry contract's label enums are the documented source of truth ("extend when adding new fallback paths"). Extended both enums so operator dashboards stay accurate.

### LOW
- **[Security] `engine.py` (degradation `detail`)** — the fetch-failure handler records `detail=str(exc)`, which could carry backend error context into the recall result's `engine_info["degradations"]`. This is an **internal library surface** (no external client boundary in this package) and matches the existing convention at six sibling degradation sites in the same engine — it is not introduced by this change. A repo-wide move to `type(exc).__name__` (or redacted detail) is reasonable but belongs in a separate, consistent sweep rather than a one-line divergence here.
- **[Security] `engine.py` (broad `except Exception`)** — intentional best-effort degradation per the failure-observability convention; consistent with the surrounding channel handlers. Keeps recall crash-free on a projection-fetch failure.
- **[Performance] `engine.py` (hydration gate)** — `set(_DOC_PROJECTION_KEYS)` is rebuilt per recall (~tens of µs) and `filter_leaf_keys` is imported function-locally (cached in `sys.modules`). Both are negligible; hoisting to module scope is an optional micro-optimization, not worth a change absent profiling data.

### INFO / Confirmed correct (positive findings)
- **IDOR-safe:** `get_document_projections_batch` is namespace-scoped — cross-tenant document keys cannot hydrate into another namespace's recall (verified by security + database reviewers; query uses parameterized `.in_()`, no injection).
- **No N+1:** exactly one batched fetch per recall, deduped `document_id` set, and the fetch **short-circuits to zero queries** when the filter has no document-key leaf (no filter / recency-only / `occurred_at` / metadata-only). Confirmed by a call-count test.
- **Graceful degradation:** a fetch failure logs a warning with `exc_info`, records a degradation, and continues with keys absent — recall never crashes; a missing/orphaned single document simply omits that id.
- **Backward compatible:** `_chunk_to_record(chunk, doc=None)` defaults `doc` to `None`; resolution chain `chunk → projection → source_document → absent` preserves prior behaviour on projection-less paths.
- **Tests sound:** the strengthened integration test uses the real three-predicate motivating filter (real `source_name` + `occurred_at` range + `metadata.tag`); the constant-unit-vector retrieval is countered by positive/sentinel anti-masking pairs (sentinel → empty proves the key bit, not "absent → matches all"); the 5-in / 5-out split still has each out-of-scope row violating exactly one predicate.
- **Docs accurate:** the `_chunk_to_record` docstring, the key-set comment ("seven", matching the 7-tuple), and the test module docstring all correctly describe the new behaviour; no stale "no faithful fallback" claims remain.

### Security audit verdict: PASS
The change adds no new external entry points (internal library recall path). The only data flow is `document_id` UUIDs (from already-retrieved candidates) into a parameterized, namespace-scoped batch read. No injection / SSRF / deserialization / path / redirect / mass-assignment vectors in the added lines.

probablyrobot (approved):

probablyrobot (approved):

PR #1071, opened 2026-06-09
Add recall-filter pushdown-spy test suite across VectorCypher channels

filippetrovic:
## Summary

White-box "pushdown-spy" tests proving the recall filter reaches every VectorCypher channel and adaptive sub-search **unchanged**: each pushdown boundary is asserted to receive a filter whose `canonical_hash` matches the caller's (a dropped filter is absent; a mutated one mismatches), with a `len(calls) >= N` vacuity guard and a differential control recall, and **no result/ranking inspection** (that is a separate concern).

- **Embedded (sqlite+LanceDB), runs without Docker:** vector channel, BM25 wiring, EXPLICIT date-synthesis (both arms), plus a point-in-time fail-fast backstop.
- **Graph (PG+Neo4j), self-skips without the live stack:** Cypher channel, session fan-out, restrictive-fallback re-run guard, CHANGE decomposition, and graph over-fetch + metadata post-filter. Each is gated on `NEO4J_INTEGRATION_TEST` + Postgres reachability, consistent with the existing Neo4j-backed integration tests; the live-run command is documented in the module docstring.
- **Partial-failure (mock-level, runs in the unit job):** the filter is preserved when a compile error surfaces (no silent vector-only fallback), and the transient-fallback filter drop is captured by a known-gap test that flips to a positive assertion once the underlying issue is fixed.
- **`canonical_hash` stability unit** and a **`temporal_filter`/`filter_ast` audit gate** that fails if a new call site drops the filter (proven non-vacuous via a mutation harness).
- Shared spy helper (sync/async capture + deterministic seed corpus) and a new `filter_enforcement` pytest marker.
- Corrects the embedded temporal-filter xfail reason to its real cause (the point-in-time `NotImplementedError` gate), keeping it `strict`.

This is a **test-only** change (no engine source modified). The investigation surfaced two pre-existing latent filter gaps, both tracked separately and documented in-suite with known-gap/xfail tests: the embedded BM25 channel accepts but does not yet compile the filter, and the transient-graph-error vector-only fallback drops the filter.

## Test plan

- [x] `pytest -m filter_enforcement` passes without Docker — 34 passed, 29 skipped (graph spies self-skip cleanly), 1 xfailed; no failures/errors
- [x] New unit/recall files pass in isolation (29 passed); embedded spies pass
- [x] `make lint` clean; `make format` clean
- [ ] Graph-config spies pass against a live PG+Neo4j stack (`make dev`) — not runnable in CI without Neo4j; self-skip there
- [ ] CI green

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -256,7 +256,7 @@ markers = [
     "embedded: Tests that exercise the SQLite+LanceDB embedded stack (no Docker required)",
     "security: Structural/regression tests for the security surface (IGR-225 IDOR gate, etc.)",
     "filter_conformance: Recall-filter conformance corpus (own CI job; excluded from the main test job)",
-    "filter_enforcement: Recall-filter pushdown/enforcement tests against real backends (runs in the integration job)",
+    "filter_enforcement: Recall-filter pushdown/enforcement/threading tests against real backends (embedded spies run in the main test job; PG+Neo4j variants self-skip without their services)",
 ]
 asyncio_mode = "auto"
 
--- a/tests/integration/matrix/test_filter_enforcement_sqlite_lance.py
+++ b/tests/integration/matrix/test_filter_enforcement_sqlite_lance.py
@@ -1,0 +1,357 @@
+"""Recall-filter pushdown/threading spies on the embedded sqlite_lance stack.
+
+These are the NO-DOCKER critical-path spies for the filter-enforcement
+feature (#1051). They run the real ``Khora(engine="vectorcypher")`` against
+a fully-embedded SQLite+LanceDB coordinator (per-test ``tmp_path``), drive a
+genuine ``kb.recall(..., filter=...)``, and OBSERVE the ``filter_ast`` that
+flows past each channel boundary — asserting it is the SAME canonical AST the
+facade built, via ``canonical_hash`` equality plus a ``len(calls) >= N``
+vacuity guard.
+
+Scope discipline (mirrors the mock-level pushdown suite): we pin the WIRING
+contract — the validated filter reaches the channel unchanged — NOT the row
+set. The end-to-end "the rows are actually narrowed" proof is the
+filter-conformance suite's job. Spying here passes through to the real method,
+so the recall executes its genuine logic; we only read what was threaded.
+
+What enforces vs. only wires on THIS embedded backend (verified against
+source — the no-Docker stack is the point):
+
+* VECTOR channel — ``SQLiteLanceTemporalStore.search`` builds a
+  ``compile_python`` post-filter and a ``compile_lance`` JSON1 pushdown, so the
+  filter is genuinely ENFORCED. The spy asserts the wiring; enforcement is the
+  conformance suite's row-proof.
+* BM25 channel — the embedded ``SQLiteLanceTemporalStore.search_fulltext``
+  accepts ``filter_ast`` for protocol parity but does NOT compile it yet, so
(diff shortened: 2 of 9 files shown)


filippetrovic:

Six reviewers (security, performance, correctness, test-quality, docs, devil's advocate) + a dedicated security audit reviewed this test-only PR. **Overall: APPROVE.** No CRITICAL or HIGH findings survive verification.

### Reviewer verdicts
| Reviewer | Verdict |
| --- | --- |
| Correctness | **Approved** — every spy wraps the real production boundary; vacuity guards + differential controls + the `entry_entities>0` gate all verified; the `canonical_hash` contract genuinely proves "filter reaches unchanged" |
| Test-quality | **Pass** — vacuity guards bite (proven by the committed spy-helper self-tests); audit gate fails on a newly-dropped site (mutation-harness verified); known-gaps carry reasons and flip when fixed |
| Security | **Clean** — test-only; correct `monkeypatch` cleanup; `pleaseletmein` is the documented local `make dev` default, not a secret |
| Performance | **Approved** — disciplined test cost; the 2-recall tests are justified differential controls, no production impact |
| Docs | **Pass** — accurate docstrings; floor-vs-embedder wording correct; **zero internal identifiers** (hard gate) |
| Devil's Advocate | 2 "blocking" objections raised — **both investigated and refuted** (see below) |

### Findings

**No CRITICAL / HIGH.**

The Devil's Advocate raised two blockers; both were verified against the code and **do not hold**:
- *"Path-9 fallback could pass vacuously on embedded"* — refuted: there is no `_vector_only_fallback` spy in the embedded suite (path-9 is mock-level only), and the embedded stack sets `skip_neo4j=True` → `_neo4j_driver=None`, so the Neo4j-transient path cannot fire there.
- *"Restrictive-fallback (path-5) is not affirmatively tested"* — refuted: `test_restrictive_fallback_rerun_suppressed_under_filter` (graph module) asserts the unfiltered re-run is suppressed under a caller filter **and** carries a differential control proving the re-run is reachable without one.

**MEDIUM**
- **M1 (docs)** — the `_vector_only_fallback` known-gap allowlist reason should state explicitly that the leak affects the **live PG+Neo4j Neo4j-transient-error degradation path** (not "embedded immune"), and ideally link a tracking issue. Comment-only; suitable for the follow-up fix PR.
- **M2 (correctness, recurring)** — the `_vector_only_fallback` transient-error path drops the caller filter (returns unfiltered chunks). This is a **real pre-existing production leak**, correctly documented here by a known-gap test + audit-gate allowlist, and **tracked separately** for a focused fix PR (kept out of this test-only change by design). Several reviewers recommend prioritizing that fix.

**LOW**
- **L1 (perf)** — two tests issue 2 recalls each (control + positive); justified by the differential pattern, no change needed.
- **L2 (test hygiene)** — the embedded BM25 control runs before the positive case; cosmetic ordering only.

### Security Audit
Clean. No dangerous patterns (no `eval`/`exec`/`subprocess`/insecure deserialization/weak crypto), no dependency or cryptographic changes, no new endpoints. Test-only surface.

### CI note
`examples-smoke` shows a failure that is **unrelated to this change**: every adapter example (google_adk, hermes, llamaindex, openai_agents) fails identically on a transient `gpt-4o-mini` empty-response (`JSONDecodeError`) during entity extraction. The job is green on `main` and this PR touches no examples/docs/extraction code — a transient CI LLM blip; re-running.

probablyrobot (approved):

PR #1072, opened 2026-06-09
Distinguish exact-array and containment operands in filter canonical hash

filippetrovic:
## Summary
- A bare-list operand (`$eq` exact-array equality, lowered to a tuple) and an explicit `{"$eq": [list]}` operand (array containment, lowered to a list) canonicalized to the same JSON array, so they collided on `canonical_hash` even though they compile to different row sets. A recall-filter cache keyed on the hash could mis-serve one form's result for the other — a latent cache-correctness hazard.
- The fix records the operand's container kind as a **structural field on the clause's canonical representation** (a sibling of the operand value), so the two forms hash differently. Because the marker lives on the clause record and not inside the operand value, no user-supplied operand (arbitrary opaque JSON carried under the operand key) can forge it.
- Hash-only change: lowering, the operand types carried on the AST, and compiler row-set semantics are untouched. The canonicalization is `@internal` (not part of the public surface); the only effect is that in-memory cache keys for array-shaped operands change.
- Adds regression tests for the exact-array vs containment distinction and for the forge-proof property (a dict operand that mimics the internal marker does not collide).

## Test plan
- [x] Unit tests pass — full `tests/unit/` + `tests/recall/` suite green (8492 passed, 46 skipped) run serially
- [x] Filter AST suite (`tests/recall/test_filter_ast.py`) passes, including new regression tests
- [x] Conformance corpus unaffected (row-set driven, not hash driven) — passes
- [x] `ruff format` / `ruff check` clean
- [ ] Integration tests run in CI (require Docker; not available in the local dev environment)

### DIFF
--- a/src/khora/filter/ast.py
+++ b/src/khora/filter/ast.py
@@ -577,6 +577,15 @@ def canonical_hash(node: FilterNode | FilterClause) -> str:
       sibling predicates does not change the hash (``{a AND b}`` == ``{b AND a}``).
     * **Order is preserved** for a ``$not`` operand, for ``$in``/``$nin`` lists,
       and for ``$eq`` exact-array operands — those are order-significant.
+    * **An ordered-array operand (a tuple: ``$eq`` exact-array, ``$in``/``$nin``)
+      hashes distinctly from a containment-list operand (an explicit ``$eq`` list).**
+      The two are disambiguated by a **structural** ``operand_kind`` field on the
+      clause record — a sibling of ``operand``, not part of the operand value — so
+      an exact-array ``$eq`` and a containment ``$eq`` on the same path do **not**
+      share a hash. The sibling placement makes it **forge-proof**: a user-supplied
+      operand value (which is opaque JSON) cannot mint the distinguishing field
+      because the value is confined under the ``operand`` key. See
+      :func:`_canonicalize` and :func:`_canonicalize_operand`.
     * **Dict operands** (whole-subdocument / whole-blob equality) have their keys
       sorted so semantically equal objects hash equally.
     * Two ASTs that differ only in **semantics** (different op, path, or operand)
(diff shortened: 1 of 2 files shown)


filippetrovic:
## Consolidated multi-perspective review — PR #1072

Reviewed across security, performance, correctness, tests, docs, and a devil's-advocate pass. **Overall: APPROVE.** One CRITICAL collision was identified during review and resolved before this comment; the merged approach is verified collision-free and forge-proof.

### Overall assessment
The fix correctly breaks the `canonical_hash` collision between a bare-list `$eq` exact-array operand (lowered to a tuple) and an explicit `{"$eq": [list]}` containment operand (lowered to a list). The discriminator is a **structural `operand_kind` field on the clause record**, not an in-operand sentinel — so no user-supplied operand value can forge it. Change is surgical and hash-only; lowering, AST operand types, and compiler row-set semantics are untouched.

### CRITICAL (resolved before merge)
- **C1 — Forgeable discriminator (devil's advocate).** An earlier iteration tagged tuples in operand-value space as `{"$arr": [...]}`. That sentinel was forgeable: `{"metadata.tags": {"$eq": {"$arr": ["a","b"]}}}` is a valid filter whose operand canonicalized to the same `{"$arr": ["a","b"]}` as the bare-list tuple form — reintroducing a collision of the exact class being fixed. **Resolution:** the discriminator was moved out of operand-value space onto the clause record (`operand_kind`), where the operand (confined under the `"operand"` key) cannot reach it. Verified: the bare-list tuple no longer collides with `{"$eq": {"$arr": [...]}}` nor with a crafted `{"$eq": {"operand_kind": "tuple"}}`. A regression test pins this.

### Correctness — no open findings
- Original collision (bare-list tuple vs explicit-`$eq` list) broken ✓
- Forge-proof property holds against crafted dict operands ✓
- All documented stability invariants preserved: commutative `$and`/`$or` sibling sorting, order preservation for `$in`/`$nin`/exact-array `$eq`/`$not`, recursive dict-key sorting, `$date`/`$dt` tagging; `$in`≠`$nin`, `$eq`-array≠`$in` (op disambiguates), empty-tuple≠empty-list ✓
- Edge cases (`None`, scalar, `datetime`, `DateLiteral`, nested tuple/list/dict) handled; `_operand_kind` falls back to a neutral `"scalar"` for non-array kinds, which never collide with an array ✓
- No placeholders (`TODO`/`FIXME`/`NotImplementedError`/stubs) introduced ✓

### Tests — no blocking findings
- Headline AC covered (`test_bare_list_and_explicit_eq_list_hash_differently`) ✓
- Forge-proof regression covered (`test_bare_list_does_not_collide_with_dict_operand_mimicking_the_tag`) ✓
- Stability regression guards present ($in order, $in≠$nin, $eq-array≠$in, commutative AND/OR, dict-key order-insensitive, $date≠string, empty-tuple≠empty-list) ✓
- Behavior-based names, no hard-coded digests (relative ==/!= assertions only), conformance corpus (row-set driven) unaffected ✓
- **L1 (optional):** an explicit assertion that `$in`/`$nin` tuples carry `operand_kind="tuple"` yet stay distinct from `$eq` exact-array via the `op` field would document that path directly; current coverage already exercises it indirectly.

### Docs — no findings
- Docstrings of `canonical_hash`, `_canonicalize`, `_canonicalize_operand`, and the new `_operand_kind` accurately describe the structural discriminator and the forge-proof argument ✓
- Zero stale `$arr` references remain in `src/` ✓
- `@internal` contract (not exported from the top-level package; in-memory cache-key change only) correctly stated ✓

### Performance — no findings
- `_operand_kind` is O(1) (three `isinstance` checks on the top-level operand, no traversal); one small field added per clause ✓
- `canonical_hash` runs once per filter compile (not a hot loop); hash churn affects only in-memory cache keys ✓

### Security — no findings
- Pure-Python `@internal` canonicalization: no I/O, auth, SQL, secrets, or sensitive-data exposure; operand is already validator-bounded before canonicalization ✓

### CI
All required checks green (18/18) after re-running transient infra failures (empty-`.venv` cache misses on lint/test-unit/weaviate and one self-healing Neo4j keepalive canary — none related to this change).

filippetrovic:
Follow-up: the full reviewer panel (security, performance, correctness, tests, docs, devil's-advocate) completed. Verdicts: correctness **APPROVED**, devil's-advocate **MERGE**, performance **negligible**, security **no findings** (recursion bounded by the model's max-depth=100 guard), tests **strong coverage**. One actionable MEDIUM (docs): a stale operand-tag comment in a test referenced the earlier in-operand sentinel — fixed in ca7c888. Optional LOW nice-to-haves (a direct unit test of the kind discriminator; an empty-array case) were noted but not blocking.

probablyrobot (approved):

PR #1073, opened 2026-06-09
Wire all recall-filter conformance families into the live-DB backend matrix

filippetrovic:
## Summary

Makes the recall-filter conformance corpus actually run cross-backend. Previously only the operator-coverage family reached a live Postgres store and the other 13 families ran the Python oracle against itself, so the cross-backend determinism guard did not execute where the recurring bugs live.

- **All 14 families now execute against live Postgres** — the matrix runner and the Postgres case set iterate every family generator, no longer just the operator family.
- **Real per-backend executors for SurrealDB, Cypher (Neo4j), Weaviate, and SQLite/LanceDB.** Each seeds a store, runs the compiled filter, and asserts row-set equality vs the Python oracle. SurrealDB and Postgres are total-exact (the compiled `WHERE` decides the row set); Cypher, Weaviate, and SQLite/LanceDB run the production read path (server-side prefilter + Python post-filter), which is the correct conformance model for partial-pushdown backends.
- **Five real compiler divergences the live matrix surfaced are fixed** (all documented in CHANGELOG, each with a regression guard):
  - SurrealDB system-key range ops now exclude an absent value — SurrealQL's `NONE` comparison is asymmetric (`NONE < x` is TRUE), so a lower-bound range op wrongly included rows with an absent date.
  - Cypher and SurrealDB `$exists` on a system key now emit a constant, matching the always-present axiom already used by the Postgres/SQLite compilers and the oracle (previously a presence test that diverged on unset denormalized columns).
  - Postgres `$in`/`$nin` dict element is now array-aware exact-per-element object equality (an array field holding the exact subdocument now matches; a stored superset still does not).
  - SQLite/LanceDB defers a dict / JSON-null `$in`/`$nin` element to the post-filter instead of mis-pushing it.
  - Cypher `$not` / `$or` over an unpushable metadata leaf now defers the whole node to the post-filter under split mode (it previously compiled `NOT(<deferred-metadata>)` → `false`, under-returning the prefilter), mirroring the SQLite/LanceDB "AND distributes; OR/NOT are all-or-nothing" rule.
- **`$exists` primitive-routing restored as executed dual-mode coverage** (pushdown + post-filter), with a coverage meta-test, plus seeded array-of-dicts cases pinning dict `$eq`/`$in` semantics.
- **Chronicle leg documented as oracle-equivalent-by-construction** — it executes through the real plan/post-filter seam but is explicitly not counted as independent backend coverage.
- **CI: every conformance leg now executes seeded row-sets.** The four previously import-only "compile-guard" legs are flipped to real conformance legs; Neo4j and Weaviate containers are provisioned (embedded SurrealDB and SQLite/LanceDB need no service).

Per-backend executed case counts: python 207, chronicle 207, cypher 207, surrealdb 188, postgres 116, weaviate 116, sqlite_lance 122. Reduced counts on some backends are documented capability prunes (e.g. string-key value reads on chunk-only backends, SurrealDB string-stored-datetime quirks), not silent skips.

## Known limitations / follow-ups

- **The Neo4j (cypher) leg seeds chunk nodes via a raw `CREATE (c:Chunk) SET c = row`**, not the production graph write path. So the cypher leg asserts compiled-`WHERE` + post-filter row-set equality against the oracle, but does **not** exercise real per-backend persistence (metadata serialization / property mapping). The other live legs (Postgres, SurrealDB, SQLite/LanceDB, Weaviate) seed through their coordinator write API. Accepted as a documented limitation for this PR; wiring the Neo4j seeder to the production graph write API is tracked as a separate follow-up so it isn't lost.

## Test plan
- [x] Lint/format clean (ruff + ty)
- [x] Filter compiler + conformance harness unit tests pass
- [x] Embedded conformance legs pass locally: python 207, chronicle 207, surrealdb 188, sqlite_lance 122
- [x] Live-DB conformance legs (Postgres, Neo4j/Cypher, Weaviate) — all green on the Filter Conformance CI workflow on this PR
- [x] Integration + unit test suite green in CI


filippetrovic:
## Multi-Perspective Review — Recall-Filter Conformance Matrix

**Overall assessment: ✅ APPROVE** — no blocking findings. Seven independent reviewers (security, performance, correctness, test, database, devil's advocate) plus a dedicated security audit. All six domain reviewers approve; the devil's advocate raised non-blocking follow-ups; the security audit verdict is **PASS**.

CI (the conformance matrix + the main test suite) is the gating signal — the live Postgres / Neo4j / Weaviate legs are exercised there.

### CRITICAL
_None._

### HIGH
_None._

### MEDIUM
- **M1 — [DevOps / Risk]** `.github/workflows/filter-conformance.yml`: the two new live-service legs (Neo4j, Weaviate) add container provisioning + readiness-probe surface to a per-PR workflow — a potential steady-state flakiness source. _Adjudication:_ these legs are **required** by the change's acceptance criteria (every backend must execute seeded row-sets; a green check must not be import-only), and they mirror the service provisioning already used by the main CI workflow. A single-leg revert-to-`compile-guard` path exists if a leg proves flaky for provisioning reasons. **Recommendation:** watch the first few runs; revert only an offending leg if the flake rate is material.
- **M2 — [Coverage]** `src/khora/filter/conformance.py` (`_backends_for_filter`) + `tests/integration/matrix/_conformance_pg.py`: the live-Postgres leg runs ~116 of 207 cases — system-string-key *value* cases are pruned because the Postgres conformance seed (skeleton temporal store) doesn't carry the denormalized document string columns. _Adjudication:_ all 14 families still execute on Postgres (no family is zero-cased), and the divergences this change targets (metadata jsonb containment, array-of-dicts, `$exists` routing) **do** run on Postgres. The pruned string-key value cases run on the Cypher + SurrealDB legs and in the Postgres compiler unit tests, so they are not untested — just not on the live-PG leg. **Follow-up (separate change):** extend the Postgres conformance seed to carry the denormalized string keys for fuller live-PG coverage.

### LOW
- **L1 — [Design]** The per-backend executor abstraction (`LiveRunner` + four executor classes + four runner modules) could in principle collapse to a single branch function. _Adjudication:_ it mirrors the pre-existing `PostgresExecutor`/`PostgresRunner` split and isolates genuinely different per-backend SDK/connection/seeding logic; reasonable as-is.
- **L2 — [Scope]** The four production compiler fixes are bundled with the test-infra + CI change. _Adjudication:_ the fixes were surfaced by and are cross-validated by the matrix (intrinsically coupled); the five commits are individually cherry-pickable.

### Security Audit — Verdict: PASS
Scanned added/modified lines across all 16 OWASP-aligned categories. No CRITICAL / HIGH / MEDIUM findings.
- **[Injection]** Test seeder/runner f-string SQL (`_conformance_lance.py`, `_conformance_surreal.py`, `_conformance_pg.py`) builds `… WHERE id IN (?)` with positional binds and embeds the compiler-emitted predicate (annotated `# noqa: S608`); operand values bind via parameters and the compilers validate path segments (`_SAFE_SEGMENT_RE`) / bind operands. Safe; test-only.
- **[Secrets]** `NEO4J_AUTH=neo4j/password`, `KHORA_NEO4J_PASSWORD=password`, anonymous Weaviate access — ephemeral CI-container credentials, discarded after the run. No real secrets, no API keys.
- **[Supply chain]** Docker images version-pinned (`neo4j:2025.12.1`, `weaviate:1.28.4`, `pgvector/pgvector:pg17`); Actions pinned to major-version tags per repo convention.
- **[File / Deserialization]** Seed-map `open()` / `json.load` operate on an operator-controlled path (env var, not user input); `tempfile.mkdtemp` for the embedded store. No path traversal, no untrusted deserialization (no pickle/yaml.unsafe_load/eval/XML).

### Per-reviewer summary
- **Security:** Clean — all four compilers parameter-bind user values; the SurrealDB `_SAFE_SEGMENT_RE` identifier guard is load-bearing; CI credentials are ephemeral.
- **Performance:** Clean — the `jsonb_array_elements(...) EXISTS` change is a standard, properly-gated pattern (prod-safe); embedded stores are cached per-worker (seeded once, O(cases)); live legs seed out-of-band and read-only (no xdist write contention); CI ≈ 90s wall, well within the 25-min budget.
- **Correctness:** Pass on all scrutiny points — exact-per-element dict match (a stored superset does NOT match), `IS NOT NONE` range guard with correct `$not` polarity, system-`$exists` constant consistent across all five backends + oracle, lance all-or-nothing deferral of dict/None `$in` (no partial-pushdown bug), the empirical SurrealDB null-drop prune neither over- nor under-prunes, per-case isolation, no placeholders.
- **Test:** Approve — strong row-set assertions against hand-declared `expected_ids` (oracle remains falsifiable); F-EXISTS 8 shapes × 2 modes (pushdown + post-filter) explicitly verified; array-of-dicts asserted bidirectionally (`$in` exact match + `$nin` negation, superset rejected); every executor invokes the real backend compiler; no silent backend skips (explicit per-case `backends` frozenset); xdist-safe seeding.
- **Database:** Approve — all four compiler fixes SQL-correct; total-exact vs split + post-filter taxonomy sound; write-once/read-many seeding is race-free; the embedded daemon-loop event-loop pattern correctly avoids connection-binding deadlock; no schema migrations.
- **Docs:** Clean — the four CHANGELOG `### Fixed` entries are accurate and free of internal identifiers; executor / chronicle-oracle-equivalence / compiler docstrings match behavior; workflow comments accurate; no stale "only one family runs" comment left behind.
- **Devil's Advocate:** Raised M1 / M2 / L1 / L2 above (advisory; none blocking).

---
_Generated by a multi-perspective agent review team + dedicated security audit._

filippetrovic:
The harness change is correct — every leg now provisions a real store, seeds through the write API, and asserts row-set equality against the `compile_python` oracle, replacing the import-only compile-guard stubs. That is exactly why the `cypher` leg is now red: it caught four genuine `compile_cypher` divergences the stub was hiding. These must be resolved before merge — do not merge with a red conformance leg.

**The four failing cases** (`conformance (cypher)`; they pass on python / postgres / surrealdb):
- `F-POLARITY-num-not-gt` — `{"<num key>": {"$not": {"$gt": n}}}`
- `F-POLARITY-num-not-lt` — `{"$not": {"$lt": n}}`
- `F-POLARITY-date-not-gt` — `$not` over a date range op
- `F-LOGIC-not-exists` — `{"$not": {"$exists": …}}`

**Root cause.** `compile_cypher` emits `$not` as a bare `NOT (<compare>)`. In Cypher a comparison against an absent or wrong-type property evaluates to `null`, and `NOT null = null` → the row is excluded. The `compile_python` oracle treats `$not` as a *total* negation: it keeps rows where the key is absent or the value fails the type-gate. So `$not($gt)` / `$not($exists)` over the graph channel drop exactly the absent / wrong-type rows the oracle returns.

**Required change.** Make `compile_cypher` `$not` compile to total negation matching the oracle — include the absent and type-gate-failing rows, e.g. `(NOT <compare> OR c.<key> IS NULL)` plus the wrong-type inclusion the oracle applies. The four cases then go green on the live `cypher` leg.

Two acceptable ways to land this:
1. Fix `compile_cypher` negation in this PR — it already bundles the `$exists`-constant cypher change, so the negation fix is in scope — and merge green; or
2. Quarantine the four cases under a strict `xfail` whose reason names the cypher `$not` divergence, and fix the compiler in a separate change. The leg must be honestly red-or-xfail, never silently green.

Prefer (1): the harness and the compiler bug it surfaced are coupled, and a green cross-backend leg is the point of this PR.

**Non-blocking.** The Neo4j leg seeds via raw `CREATE (c:Chunk) SET c = row` rather than the production graph write API, so it doesn't exercise real per-backend persistence (metadata serialization / property mapping). Acceptable as a documented limitation if a follow-up tracks wiring the graph write path — note it in the PR description so it isn't lost.

filippetrovic:
Addressed via **option (1) — fixed `compile_cypher` in this PR** (commit `c78f1d5a`); the `conformance (cypher)` leg is now **green** along with all other 17 checks. No `xfail`.

**On the mechanism.** All four failing cases negate an *unpushable* `metadata.*` leaf:
- `F-POLARITY-num-not-gt` → `{metadata.score: {$not: {$gt: n}}}`
- `F-POLARITY-num-not-lt` → `{metadata.score: {$not: {$lt: n}}}`
- `F-POLARITY-date-not-gt` → `{metadata.due: {$not: {$gt: <date>}}}`
- `F-LOGIC-not-exists` → `{metadata.k: {$not: {$exists: true}}}`

Neo4j stores metadata as a serialized JSON string property (not nested map keys), so a metadata sub-path has **no `c.<key>` graph property** to compare or null-check — the `(NOT <compare> OR c.<key> IS NULL)` total-negation push only applies to *pushable* properties. Cypher already negates **pushable system keys** correctly: their leaves are built total (coalesce-to-false on absent/wrong-type), so a wrapping `NOT` flips absent rows back in — system-key `$not` was never the divergence.

For the unpushable metadata leaves, the fix defers the **whole** `$not` / `$or` node to the `compile_python` post-filter under split mode (it previously compiled `NOT(<deferred-leaf="true">)` → `false`, dropping every row). This is the same "AND distributes; OR/NOT are all-or-nothing" rule the SQLite/LanceDB compiler already uses. The post-filter **is** the oracle, so the result is exactly the oracle's total negation — including the absent and type-gate-failing rows — which is precisely the required behavior. AND still distributes (a deferred metadata conjunct becomes `true`, system-key conjuncts still push); `on_unsupported="raise"` is unchanged. Added 8 regression tests + a CHANGELOG `### Fixed` entry.

**Non-blocking note.** Documented the raw `CREATE (c:Chunk) SET c = row` Neo4j seeding limitation in the PR description, and opened a follow-up to wire the Neo4j conformance seeder to the production graph write path (so it exercises real metadata serialization / property mapping) — it won't be lost.

probablyrobot (approved):

PR #1075, opened 2026-06-10
Persist and read occurred_at on the SurrealDB chunk path

filippetrovic:
## Summary
- The SurrealDB `chunk` table is schemafull, and `occurred_at` / `last_accessed_at` were never declared via `DEFINE FIELD` — SurrealDB silently strips undeclared fields, so even a code-side write would no-op at the DB layer. Both fields are now declared as `option<datetime>`, alongside the existing `source_timestamp`.
- Wire both fields through the four chunk sites in the SurrealDB vector adapter (`create_chunk` SET clause, `create_chunks_batch` record dict, `_chunk_to_bindings`, `_row_to_chunk`), mirroring the existing `source_timestamp` handling. Previously a distinct `occurred_at` written through this backend read back NULL and the effective event time silently fell back to `source_timestamp` via `COALESCE(occurred_at, source_timestamp)` — benign for recall, but a distinct event time could not be honored. This closes the last backend gap; the sqlite_lance and pgvector siblings were fixed in #1054 and #1062.
- The temporal WHERE clauses in the adapter intentionally keep `(source_timestamp ?? created_at)`: recall filters on `occurred_at` are post-filter-only by design (only the `source_timestamp` axis pushes down), matching the other backends.
- Add an embedded SurrealDB (in-memory, no external service) integration test module mirroring the sqlite_lance/pgvector siblings: a distinct `occurred_at` (and `last_accessed_at`) round-trips through the real store, and an `occurred_at` recall filter is honored for a chunk whose `source_timestamp` is out of range — proving genuine persistence rather than COALESCE recovery.

## Test plan
- [x] `make test` green locally: 8577 unit + 378 integration passed, coverage 84% (gate 77%). Postgres/Neo4j-gated tests self-skip locally and run in CI.
- [x] New tests verified as genuine regression guards: with the adapter fix stashed, both fail (round-trip reads back NULL; the in-range-`occurred_at` chunk is wrongly dropped).
- [x] `make format` clean, `make lint` exits 0. Production diff is +10 lines.

### DIFF
--- a/src/khora/storage/backends/surrealdb/schema.py
+++ b/src/khora/storage/backends/surrealdb/schema.py
@@ -104,6 +104,8 @@ DEFINE FIELD IF NOT EXISTS metadata_ ON chunk FLEXIBLE TYPE option<object>;
 DEFINE FIELD IF NOT EXISTS chunker_info ON chunk FLEXIBLE TYPE object DEFAULT {};
 DEFINE FIELD IF NOT EXISTS created_at ON chunk TYPE datetime DEFAULT time::now();
 DEFINE FIELD IF NOT EXISTS source_timestamp ON chunk TYPE option<datetime>;
+DEFINE FIELD IF NOT EXISTS occurred_at ON chunk TYPE option<datetime>;
+DEFINE FIELD IF NOT EXISTS last_accessed_at ON chunk TYPE option<datetime>;
 DEFINE FIELD IF NOT EXISTS session_id ON chunk TYPE option<string>;
 DEFINE INDEX IF NOT EXISTS idx_chunk_namespace ON chunk FIELDS namespace;
 DEFINE INDEX IF NOT EXISTS idx_chunk_document ON chunk FIELDS document;
--- a/src/khora/storage/backends/surrealdb/vector.py
+++ b/src/khora/storage/backends/surrealdb/vector.py
@@ -131,6 +131,8 @@ class SurrealDBVectorAdapter:
             "chunker_info = $chunker_info, "
             "created_at = $created_at, "
             "source_timestamp = $source_timestamp, "
+            "occurred_at = $occurred_at, "
+            "last_accessed_at = $last_accessed_at, "
             "session_id = $session_id"
         )
         bindings = self._chunk_to_bindings(chunk)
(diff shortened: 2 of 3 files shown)


filippetrovic:
## Multi-Perspective Review (7 reviewers + security audit)

**Overall assessment: APPROVE** — no blocking findings. The production diff (+10 lines) closes the last backend gap for distinct `occurred_at` persistence, mirrors the sqlite_lance/pgvector siblings exactly, and the new test module was verified to fail without the fix.

### CRITICAL
None.

### HIGH
None.

### MEDIUM
- **M1 [Devil's Advocate / Correctness]** `src/khora/storage/backends/surrealdb/vector.py` — The SurrealDB adapter still lacks an `update_last_accessed()` method, which both siblings implement (`pgvector.py:530`, `sqlite_lance/vector.py:394`), so the recall-reinforcement path cannot stamp access times on SurrealDB. This is a **pre-existing gap, not introduced by this PR** — writing `last_accessed_at` at create time is exact parity with both siblings (`pgvector.py:390,431`; sqlite_lance INSERT column list), and in production the field is `None` at create, so behavior is unchanged. Recommended follow-up: implement `update_last_accessed()` on the SurrealDB adapter in a separate PR.

### LOW
- **L1 [Devil's Advocate]** Folding `last_accessed_at` into an `occurred_at`-titled PR is mild scope batching. Accepted as deliberate: the two fields share the same schemafull root cause (undeclared `DEFINE FIELD` → silent strip), the wiring is symmetric, and splitting would duplicate the test scaffold.

### Reviewer verdicts
| Reviewer | Findings |
|---|---|
| Security | None — new fields use `$placeholder` bindings (no interpolation), no secrets, no new trust boundaries |
| Performance | None — ~50-60 bytes/record overhead; no index needed (`occurred_at` recall filters are post-filter-only by design) |
| Correctness | None — `option<datetime>`/`datetime \| None` symmetric, tz-aware round-trip via `_parse_dt`, batch/single paths symmetric, no chunk update path to clobber the fields, no placeholders |
| Tests | None — regression guards verified (both tests fail with the adapter fix stashed); coverage exceeds siblings by asserting `last_accessed_at` round-trip |
| Database | None — `DEFINE FIELD IF NOT EXISTS` idempotent, double-checked-locking schema init, existing rows read back NULL, zero deployment risk |
| Docs | None — no doc drift; architecture doc doesn't enumerate chunk fields; CHANGELOG is per-release by convention |
| Devil's Advocate | M1 + L1 above; explicitly endorses the core fix (COALESCE fallback masks, not fixes, a dropped distinct event time) |

## Security Audit
Dedicated 16-category audit (injection, auth, secrets, access control, SSRF, deserialization, file handling, redirects, mass assignment, race conditions, ReDoS, supply chain, boundary validation, etc.) over the added lines: **no findings**. All SurrealQL writes are parameterized; the test module uses an in-memory database with deterministic fixtures and no credentials; no dependencies changed.

**Verdict: PASS**

msrdic (approved):

PR #1076, opened 2026-06-10
Seed Neo4j conformance leg via the production graph write path

filippetrovic:
## Summary
- The live-DB recall-filter conformance matrix's Neo4j (cypher) leg previously seeded `Chunk` nodes via a raw `UNWIND $rows AS row CREATE (c:Chunk) SET c = row` with a hand-rolled property map — so the leg never exercised the real per-backend persistence. This was the one documented seeding limitation left by #1073.
- `build_seed_map()` now adapts each corpus `SeedRecord` to a `TemporalChunk` (`_to_temporal_chunk`, mirroring the postgres leg's adapter) and writes through the production chunk-node writer (`DualNodeManager.create_chunk_nodes_batch`) — the same write path the VectorCypher engine uses. The compiled Cypher prefilter now reads system-key properties (ISO date strings, denormalized document string keys) produced by the real writer, so a future production-side property-mapping change breaks this leg instead of passing silently. The production `serialize_dict` metadata form also lands on the node (write path exercised), though cypher never pushes metadata into the WHERE — it stays post-filtered in-memory, as before.
- created_at semantics: the production writer stamps an absent `created_at` to `now()`, but the corpus keeps `created_at` cases on the cypher leg precisely because cypher leaves an absent `created_at` missing/NULL. A stamped value in the prefilter is a latent date-dependent divergence ($gt would wrongly keep the absent-created_at seed; $ne/$nin must keep it for Mongo-faithful negation). The seeder therefore strips the stamped property (`_strip_stamped_created_at`) from exactly the nodes whose `SeedRecord` left `created_at` None, restoring the documented missing-property semantics. The corpus canary seed separates correct from broken in both directions and is statically consistent with expected row-sets.
- The write-once seed-map / read-only test pattern, the read path (`run_live` / `executor_for`), the cypher compiler, and the production writer are all unchanged. Orphaned hand-rolled helpers (`_node_properties`, `_DATE_KEYS`/`_STRING_KEYS`) removed; module docstring updated to the new rationale.

## Test plan
- [x] Unit suites: `tests/unit/` + `tests/recall/` — 8,599 passed locally (filter suite 1258, conformance catalog+harness 692, vectorcypher pushdown 30, matrix seed-map/registry guards 8 included)
- [x] No raw `CREATE (c:Chunk) SET c = row` remains in `tests/`
- [x] Corpus parity: 207 cypher cases across all 14 families; seed-module and test-module case-ID sets identical (no write/read drift)
- [x] Seed entrypoint verified to build the corpus and the production writer correctly, failing only at the TCP connect in a DB-less environment
- [ ] Live cypher leg: `conformance (cypher)` CI check on this PR (no container runtime locally — same verification route as #1073)

### DIFF
--- a/tests/integration/matrix/_conformance_neo4j.py
+++ b/tests/integration/matrix/_conformance_neo4j.py
@@ -13,13 +13,26 @@ serialized JSON property on the chunk node, not pushable to Cypher) to the engin
 in-memory post-filter — exactly as the VectorCypher retriever does. So the Cypher
 candidate set is a superset the ``compile_python`` oracle narrows.
 
-Why a hand-rolled ``Chunk`` seeder (not ``seed_case`` through a coordinator): the
-graph backend is not coordinator-wired for chunk writes, and the cypher compiler
-reads only ``Chunk``-node properties — the system-key columns and a metadata
-property. This module writes EXACTLY that surface: every datetime system key as an
-``.isoformat()`` string (so the compiler's lexicographic ISO compares apply, matching
-``compile_cypher``'s date-binding) and ``metadata`` as a JSON-string property the
-post-filter decodes.
+Why the production write path: the seed goes through
+:meth:`DualNodeManager.create_chunk_nodes_batch` — the exact code VectorCypher runs
+to land ``Chunk`` nodes — so the conformance leg exercises the real property mapping,
+including the production ``serialize_dict`` form of ``metadata`` landing on the node,
+not a hand-rolled approximation of it. The compiled Cypher predicate then reads only
+the pushed-down system-key properties (per the split-pushdown note above); ``metadata``
+is never in the WHERE — it stays in the in-memory post-filter — so the production write
+form is what's under test on the write side. Each :class:`SeedRecord` is adapted to a
+:class:`TemporalChunk` (mirroring how ``_conformance_pg`` adapts to the skeleton
+store for the postgres leg) so both legs feed their production writer the same corpus.
+
+One corpus-fidelity adjustment: ``create_chunk_nodes_batch`` stamps an absent
+``created_at`` to ``now()`` (the production default), but the conformance corpus
+keeps ``created_at`` cases on the cypher leg *because* cypher is expected to leave an
+absent ``created_at`` NULL (an absent row that gets a ``now()`` value would satisfy a
+lower-bound pushdown and break the by-construction ``expected_ids``). So after the
+batch write, the seeder ``REMOVE``s ``created_at`` from exactly the nodes whose
+``SeedRecord`` left it ``None``, restoring the missing-property semantics the compiler
+relies on. ``occurred_at`` / ``source_timestamp`` are user-supplied and the production
+writer already leaves them absent when ``None``, so they need no fixup.
 
 Seed/read split (write-once, read-many), mirroring the live-Postgres leg. The graph
 is seeded ONCE by ``_conformance_seed`` (the workflow's one-time step), which also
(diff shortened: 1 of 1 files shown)


filippetrovic:

Seven independent reviewers (security, performance, correctness, tests, docs, database, devil's advocate) plus a dedicated security audit examined this PR. Findings below are deduplicated and severity-classified by the review lead.

## Review Summary

**Overall assessment: APPROVE** — no CRITICAL or HIGH findings. One design question recorded for discussion (M1, non-blocking); the rest are optional hardening/quality notes.

### MEDIUM

- **M1 [Devil's Advocate / design]** `_strip_stamped_created_at` — The seeder routes writes through the production writer, then strips the writer's `created_at` stamp from nodes whose seed left it `None`. The devil's advocate argues this tests a node state production itself never writes (production always stamps), and that pruning `created_at` cases for this backend (as the stamping backends do) would be more honest. **Lead's disposition:** the corpus deliberately classifies this backend as NULL-preserving for `created_at` and keeps those cases executed — that classification is a pre-existing corpus design decision from #1073, not this PR, and pruning would *reduce* executed coverage (≈11 cases). Absent-`created_at` nodes can also exist in real graphs (legacy data, other writers), so the NULL-semantics coverage has value. The module docstring documents the tradeoff explicitly. Non-blocking; if the corpus design changes later, the strip and the backend classification should change together.

### LOW

- **L1 [Security]** `tests/integration/matrix/_conformance_neo4j.py:219` — The strip query is an f-string interpolating `_NODE_ALIAS`. The alias is a hardcoded module constant (`"c"`) and all values bind via `$ids` parameters, so there is no injection path; the identical pattern pre-exists on the read path in this module. Optional hardening: hoist the query to a static string constant.
- **L2 [Database/Performance]** `build_seed_map` — `ensure_indexes()` is not called before seeding, so the strip query and reads may run as node scans on a fresh container. At corpus scale (~1k nodes) this is milliseconds; optional parity improvement to call it once after constructing the manager.
- **L3 [Database]** `build_seed_map` — Seeding is one-time by design; re-running against a dirty local container leaves stale nodes (CI uses a fresh container, and the seed map's fresh UUIDs keep tests correct regardless). Optional docstring note for local dev.
- **L4 [Security]** Seed-map path comes from an env var with no path validation — test-only tooling, trusted-environment assumption. Optional doc note.

### No findings

- **Correctness** — adapter maps every compiler-read field with no silent drops; strip targets exactly the `created_at is None` records and runs before the seed map is written; chunk-id round-trip (UUID → str → node property → UUID) verified end-to-end; empty-case and all-None edge cases handled.
- **Tests** — all 14 families / 207 cases preserved; seed and test modules share the single source-of-truth case selector (no drift possible); write-once/read-only xdist pattern intact; no assertion weakened.
- **Performance** — one manager/driver for the whole seed; ~90 extra strip round-trips ≈ 180–450 ms on a one-time CI step; negligible.
- **Docs** — every docstring claim verified against the code (stamp behavior, no-fixup for `occurred_at`/`source_timestamp`, metadata-not-in-WHERE); PR description matches the diff.

## Security Audit

Dedicated audit of the added/modified lines across 16 categories (injection, authn, crypto/secrets, access control, SSRF, deserialization, file handling, supply chain, …): **PASS**. No credentials, no new sinks, no untrusted input reaches the Cypher queries — all bound values are locally generated UUID strings via driver parameters; the single f-string interpolation is the module-constant alias (covered by L1). Default test credentials in the connection block are pre-existing unchanged lines (out of per-PR audit scope) and match the repo's CI/compose test stack.

## Verdict

**APPROVE.** The change delivers its goal — the conformance leg now seeds through the production chunk-node writer, so production-side property-mapping/serialization drift breaks this leg instead of passing silently — with a precise, well-documented corpus-fidelity adjustment for the writer's `created_at` stamp. CI is green across all conformance legs including the live cypher leg.

msrdic (approved):

PR #1077, opened 2026-06-10
Add deterministic end-to-end recall-filter test harness

filippetrovic:
## Summary
- Add a `tests/e2e/` suite that drives real `remember()` → `recall(filter=...)` per engine and asserts the filter narrows rows end to end, reconciled against the hand-authored conformance corpus.
- **Deterministic, network-free ingest:** a content-keyed stub extractor + hash embedder, HyDE and reranking disabled, a fresh namespace per test — so the candidate set and `expected_ids` stay stable across runs.
- **Doc-level `external_id` reconciliation:** each conformance record is seeded through `remember(external_id=...)`; recall applies the case filter; the surviving `external_id` set is asserted equal to the expected ids (set-based, never ranked) and cross-checked against the Python oracle. Seeding through the real ingest path (not a low-level store write) is what makes this an end-to-end proof.
- **Graph-channel proof** on the live engine lane: persistence (entity/relationship counts) plus a contribution gate and a negative tripwire (same corpus, no extracted entities → empty graph recall), so a vacuous pass is impossible.
- **F-EXISTS reachability** across the eight presence states (with a completeness guard) and a **multi-chunk denormalization-uniformity** check.
- **Engine coverage:** the embedded sqlite_lance lane runs with no Docker (the default lane); the live Postgres+graph and Postgres-only lanes self-skip cleanly without services and run under `make dev`.
- **No production code changes** — pure test harness.

## Test plan
- [x] Embedded sqlite_lance lane green with no Docker: `uv run pytest tests/e2e -m "e2e and not slow"` → 41 passed
- [x] Full-repo collection clean (no import/collection errors); `ruff format --check`, `ruff check`, `ty check` clean on the new suite
- [ ] Live graph + Postgres lanes: run under `make dev` with `NEO4J_INTEGRATION_TEST=1` (self-skip without services)
- [ ] CI: unit + integration suites pass with coverage

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -275,8 +275,15 @@ jobs:
         # compiler-dispatch, plus the skeleton-pgvector integration test which
         # self-skips when Postgres is unreachable); collect it here so the
         # filter tests gate in CI.
+        #
+        # tests/e2e/ holds the recall-filter end-to-end harness. The embedded
+        # sqlite_lance lane is hermetic (no services) and is the must-always-run
+        # lane, so it gates here; the live PG+Neo4j / chronicle modules carry the
+        # ``slow`` marker (excluded by the default ``-m "not slow ..."`` addopts)
+        # AND a reachability self-skip, so they collect and skip cleanly with no
+        # services. The dedicated slow/e2e job for the live lanes is a follow-up.
         run: |
-          uv run pytest tests/unit/ tests/recall/ --cov=src/khora --cov-branch --cov-report=xml:coverage-unit.xml --cov-fail-under=0 -n auto --timeout=120 --timeout-method=thread
+          uv run pytest tests/unit/ tests/recall/ tests/e2e/ -m "not slow and not filter_conformance" --cov=src/khora --cov-branch --cov-report=xml:coverage-unit.xml --cov-fail-under=0 -n auto --timeout=120 --timeout-method=thread
 
       - name: Generate JSON coverage for floor check
         run: uv run coverage json -o coverage.json
--- a/tests/e2e/DESIGN_NOTES.md
+++ b/tests/e2e/DESIGN_NOTES.md
@@ -44,10 +44,10 @@ tests/e2e/
   DESIGN_NOTES.md                  # this file (architect)
   conftest.py                      # BACKEND owns
   _harness.py                      # BACKEND owns
-  test_filter_rowset_embedded.py   # QA owns. sqlite_lance, no-Docker main lane -> AC1 + AC4 + AC5
-  test_filter_rowset_graph.py      # QA owns. live VectorCypher PG+Neo4j, self-skip -> AC3 (graph lane)
-  test_filter_rowset_chronicle.py  # QA owns. live Chronicle PG-only, self-skip -> AC3 (chronicle lane)
-  test_graph_contribution.py       # QA owns. live PG+Neo4j, self-skip -> AC2 + AC3 honesty (pre-flight + tripwire)
+  test_filter_rowset_embedded.py   # QA owns. sqlite_lance, no-Docker main lane -> AC1 + AC4 + AC5 (metadata)
+  test_filter_rowset_graph.py      # QA owns. live VectorCypher PG+Neo4j, self-skip -> AC3 + system-key AC4/AC5
+  test_filter_rowset_chronicle.py  # QA owns. live Chronicle PG-only, self-skip -> AC3 + system-key AC4/AC5
+  test_graph_contribution.py       # QA owns. live PG+Neo4j, self-skip -> AC2 + AC3 honesty (pre-flight + firing tripwire + filter-enforcement drop)
   test_multichunk_denorm.py        # QA owns. sqlite_lance, no-Docker -> AC6
 ```
 
@@ -59,10 +59,10 @@ embedded lane carries the corpus families (AC1 deterministic ingest, AC4 externa
 reconciliation, AC5 F-EXISTS reachability) since it runs in the no-Docker main job.
 
 - **`conftest.py` (Backend)** — per-engine `Khora` fixtures (sqlite_lance embedded
-  / vectorcypher PG+Neo4j / chronicle), the **autouse deterministic-LLM/extractor
-  fixture**, and HyDE-off (+ reranking-off) config. `_pg_reachable()` /
-  `_neo4j_reachable()` self-skip guards so a no-Docker run collects-and-skips the
-  live legs cleanly.
+  / vectorcypher PG+Neo4j / chronicle), each installing the deterministic
+  LLM/extractor stub (`stub_llm`) before building its `Khora`, with HyDE-off
+  (+ reranking-off) config. `_pg_reachable()` self-skip guard so a no-Docker run
+  collects-and-skips the live legs cleanly.
 - **`_harness.py` (Backend)** — the deterministic extract stub, the entity-bearing
   + multi-chunk seed builders, doc-level `reconcile()`, and a `neo4j_populated()`
   probe. The single surface QA imports.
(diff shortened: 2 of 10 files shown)


filippetrovic:
## Multi-perspective review — APPROVE

Six independent reviewers (security, correctness, test-quality, performance, docs, devil's advocate) plus a focused security audit reviewed this PR. **Overall: APPROVE.** Five reviewers approve; the devil's advocate raises scope/design concerns (recorded below) that are advisory and do not block merge. Pure test harness — zero production-code changes.

### MEDIUM
- **[Devil's Advocate]** `tests/e2e/DESIGN_NOTES.md` — the framing that the existing conformance suite "doesn't prove row narrowing" is imprecise. The conformance suite proves the **filter compiler** (directly-seeded rows executed by a test executor); this harness proves the **end-to-end ingest→recall path** (`remember()` → the production `recall(filter=…)` API), which is empirically distinct — directly-seeded rows are not returned by the engine recall path. Suggest the notes draw that distinction explicitly rather than implying the compiler proof is absent. The harness covers a genuinely different path, so this is a clarity note, not a redundancy.
- **[Devil's Advocate]** Shipping a ~281-line design document inside `tests/` is a judgment call; consider moving it to `docs/` or trimming it into docstrings. The other reviewers found it acceptable as an internal maintainer guide.
- **[Devil's Advocate / Test]** The live graph/chronicle lanes don't run in the default test lane (`slow` + self-skip without services) and rely on a documented entry-gate workaround (zeroed entity-similarity floor + a fail-loud `graph_chunk_count > 0` pre-flight tripwire). They need a services-up run to validate. This is by design — the no-Docker embedded lane gates CI; the live lanes run in the integration job — but flagged for visibility.

### LOW
- **[Security]** Default graph password hardcoded in the live fixture (environment-overridable, `SecretStr`-wrapped, matches existing integration patterns).
- **[Test]** Duplicated metadata-case selector between the embedded and graph lanes (`"…" in case.backends`); a shared helper would prevent silent divergence.
- **[Test]** The live fixture skips migrations (`run_migrations=False`) — correct (the live DB is assumed pre-migrated); a one-line comment would clarify operator responsibility.
- **[Correctness]** The oracle cross-check assertion is redundant with the expected-ids assertion (intentional defensive programming).
- **[Correctness]** Exact-instant timestamp cases are curated out of the engine lane without a completeness guard (the family stays covered by the conformance suite).
- **[Performance]** Parametrization cardinality grows linearly with the conformance corpus; seeding is single-threaded (fine while extraction is stubbed). Monitor if the corpus exceeds ~60 cases.

### Security Audit
Focused pass on the diff: no secrets beyond the known default test credential, no dangerous execution sinks (`eval`/`exec`/`pickle`/`subprocess`), **no new dependencies**, and no raw SQL/Cypher interpolation (filters route through the parameterized compiler). Clean.

### Strengths (consensus)
- Set-equality assertions (`frozenset`, generous limit) make results ranking-independent; oracle cross-checks catch authoring drift.
- The graph-channel proof is non-vacuous: a positive pre-flight gate plus a negative tripwire (same corpus, no entities → empty graph recall) prove causation rather than mere existence.
- Doc-level reconciliation by `external_id` over the returned chunks; fresh namespace per test; HyDE/reranking off; deterministic network-free stubs.
- A completeness guard pins the eight presence states; multi-chunk denormalization uniformity is verified end to end.
- Zero placeholders; proper fixture teardown; no resource leaks.

**Verdict: APPROVE.** The MEDIUM/LOW items are advisory; none blocks merge.

filippetrovic:
Review of the e2e recall-filter harness — three gaps to close, two cleanups.

**1. CI never collects `tests/e2e/`.** Every pytest invocation in the workflows uses explicit paths (`ci.yml` test-unit: `tests/unit/ tests/recall/`; test-integration: `tests/integration/ … -m "integration and not filter_conformance"`; `filter-conformance.yml`: two named files). The embedded sqlite_lance lane is described as the must-always-run no-Docker lane, but as merged no CI job will ever collect it — and the green checks on this PR executed none of the new tests. Fix in this PR: add `tests/e2e/` to the test-unit job's pytest paths (it is hermetic; the live modules carry `slow`/reachability skips and will skip cleanly, but prefer also deselecting them explicitly there). The dedicated slow/e2e job for the live lanes can stay a follow-up, but the embedded lane must gate now, and CI must demonstrably run it before merge.

**2. The graph-channel filter-drop proof is missing.** `test_rowset_reconciliation_graph` seeds without `plan_extraction`, so the stub extractor emits nothing, zero entities are written, and the graph channel holds no candidates during the filtered recalls — a regression where the graph channel ignores `filter_ast` and smuggles violating chunks into the fused result passes this suite. The contribution gate and negative tripwire only cover unfiltered recalls. Add a test on the live graph lane that (a) stages `plan_extraction` for a marker present in every seeded doc, (b) confirms contribution (`graph_chunk_count > 0`, as the existing gate does), then (c) runs `recall(filter=…)` where some graph-fed docs violate the filter and asserts the surviving `external_id` set excludes them — i.e. the "smuggled violating chunk fails the set-equality assertion" case. DESIGN_NOTES' own "Graph-channel proof" section prescribes exactly this filtered negative; the shipped tripwire (no-extraction ⇒ empty graph) proves causation of *firing*, not filter enforcement.

**3. F-EXISTS reachability covers 6 of 8 states.** `_exists_metadata_cases` keeps only `metadata.*` presence states and its docstring claims "the two `source_name` (system-key) states run on the live PG lane" — but neither `_graph_rowset_cases` nor `_chronicle_rowset_cases` includes `f_exists_cases()`, so `F-EXISTS-sys-true`/`-sys-false` are driven through real `remember()` nowhere. `source_name` is already threadable via `_REMEMBER_STRING_KEYS`; add the two system-key cases to the live-lane parametrization (or correct the docstring and record the gap if the embedded chunk row genuinely can't narrow on them anywhere).

**4. Dead harness machinery.** `_harness.ENGINE_PARAMS`, `engine_rowset_cases`, `_is_exact_timestamp_match`, and the conftest `kb`/`namespace_id` indirection fixtures are referenced by zero tests — every module pins its engine fixture directly and mints its own namespace. Either wire the three row-set modules through `ENGINE_PARAMS` (which also collapses the three near-identical `_*_rowset_cases` selectors) or delete the machinery. Note `engine_rowset_cases` is the only thing that curates the system-key F-OP families the embedded module's docstring defers to "the live PG/vectorcypher lane" — as shipped, no e2e case filters on a denormalized document column on any lane. Wiring it in on the live lane closes that too.

**5. DESIGN_NOTES drift.** It claims an autouse stub fixture plus a `KHORA_QUERY_ENABLE_HYDE=false` env override set "before any Khora is built"; conftest has neither (config field + `test_hyde_is_disabled` pin only — adequate, but the notes should match the code). Its AC3 row also credits the rowset graph/chronicle modules with "the graph channel empties under a violating filter," which is not implemented (see point 2). Also: the skeleton-pgvector engine lane from the planned engine matrix is absent — if deferred, say so in DESIGN_NOTES.

Solid as-is: the class-method stub seam (real pipeline above the LLM leaf, not `extract_entities=False`), distinct-content-per-record seeding, set-based `external_id` reconciliation keyed off returned chunks with the conformance `expected_ids` as the assertion target, per-test registry clear (no mock leakage, order-independent tripwire), the 8-state corpus-shrink guard, and the multi-chunk uniformity assertions.

filippetrovic:
Thanks — all five addressed in b3c8e0e0.

**1. CI never collects `tests/e2e/`.** Added `tests/e2e/` to the `test-unit` job's pytest paths with an explicit `-m "not slow and not filter_conformance"`: the hermetic sqlite_lance lane now runs there (it gated nowhere before), while the live modules' `slow` + reachability marks deselect/skip them cleanly. The dedicated slow/e2e job for the live lanes stays a follow-up (noted in DESIGN_NOTES). CI on this push exercises the embedded lane for the first time.

**2. Graph-channel filter-drop proof.** New `test_graph_channel_drops_violating_chunks` in `test_graph_contribution.py`: every seeded doc is entity-bearing (the stub emits the shared entity for all of them, so the graph channel feeds them all), half tagged `keep` and half `drop`; a no-filter GRAPH pre-flight confirms `graph_chunk_count > 0`, then the filtered GRAPH recall (`metadata.tag in {"keep"}`) must reconcile to exactly the keep set — a smuggled violating graph-fed chunk fails the set-equality assertion. This is the filter-**enforcement** case, distinct from the no-extraction firing tripwire.

**3. F-EXISTS 6/8 → 8/8.** The two `source_name` system-key states now run through real `remember()` on the live graph + chronicle lanes (they need a denormalized chunk column the embedded row lacks). All eight presence states are now driven end to end somewhere; the embedded docstring + the completeness guard are updated to match.

**4. Dead machinery.** Collapsed the three near-identical `_*_rowset_cases` selectors into one shared `_harness.rowset_cases(backend, *, include_system_keys)`. The live lanes pass `include_system_keys=True`, which wires the system-key `F-OP` families (via `engine_rowset_cases`, now live) + the `source_name` `F-EXISTS` states — so an e2e case now filters on a denormalized document column (it previously did nowhere). Deleted the unused `ENGINE_PARAMS` and the conftest `kb` / `namespace_id` indirection fixtures.

**5. DESIGN_NOTES drift.** Removed the "autouse fixture + `KHORA_QUERY_ENABLE_HYDE=false` env override" claim (as shipped: per-fixture `stub_llm` + `enable_hyde="never"` config + the `test_hyde_is_disabled` pin); the AC3 row now credits the filter-enforcement proof; the AC5 row reflects the 6-embedded/2-live split; added a CI-gating section and recorded the skeleton-pgvector lane as deferred; fixed the `seed_corpus` → `seed_records` reference.

Verified locally: embedded lane `-m "e2e and not slow"` → 41 passed; full `-m e2e` collects 207 (no errors) with the live lanes skipping cleanly without services; `ruff`/`ty` clean. The new live-lane cases (system-key `F-OP` + `source_name` `F-EXISTS` + the filter-drop test) collect and self-skip here — they execute under `make dev` + `NEO4J_INTEGRATION_TEST=1`; the existing live filter tests confirm the recall path narrows on `source_name`, so the system-key families are well-founded, but they want one services-up run to confirm green.

msrdic (approved):

PR #1079, opened 2026-06-10
Carry the recall filter through the transient vector-only fallback

filippetrovic:
## Summary
- VectorCypher's `retrieve()` catches transient Neo4j errors and degrades to `_vector_only_fallback` — but that fallback did not accept `filter_ast`, so a recall issued with a deterministic recall filter silently returned unfiltered (potentially filter-violating) chunks on the degraded path. This was previously pinned as a known gap by the pushdown-spy suite.
- `_vector_only_fallback` now accepts `filter_ast` and threads it into its `_simple_retrieve` call, so both the vector and BM25 channels on the degraded path honor the same `WHERE` as the primary path. If filter compilation fails there, the exception propagates intentionally — a filter the vector channel cannot honor fails loud rather than returning filter-violating chunks, matching the graph-path contract.
- The partial-failure suite's Arm-C known-gap test is flipped to a positive enforcement: it now asserts the fallback receives the caller's exact filter (`canonical_hash` equality), with the existing vacuity guard retained.
- The static audit gate's allowlist entry for the fallback omission is removed, and `_vector_only_fallback` is registered as a filter-aware method, so the gate now enforces that the fallback carries the filter.

## Test plan
- [x] Unit tests pass (`make test` unit phase: 8577 passed, 43 skipped)
- [x] Integration tests pass (`make test` integration phase: 376 passed, 173 skipped, 5 xfailed; coverage 84% ≥ 77% gate)
- [x] Flipped enforcement test exercises the real `retrieve()` transient-error path via a spy and verifies canonical-hash equality of the threaded filter

### DIFF
--- a/src/khora/engines/vectorcypher/retriever.py
+++ b/src/khora/engines/vectorcypher/retriever.py
@@ -808,6 +808,7 @@ class VectorCypherRetriever:
                         temporal_signal=temporal_signal,
                         min_similarity=min_similarity,
                         mode=mode,
+                        filter_ast=filter_ast,
                     )
 
             span.set_attribute("chunk_count", len(result.chunks))
@@ -2033,11 +2034,14 @@ class VectorCypherRetriever:
         temporal_signal: TemporalSignal | None = None,
         min_similarity: float = 0.0,
         mode: SearchMode = SearchMode.HYBRID,
+        filter_ast: FilterNode | None = None,
     ) -> VectorCypherResult:
         """Fallback to vector-only search when graph operations fail.
 
         This provides graceful degradation when Neo4j is unavailable or
         returns errors. Results are still useful, just without graph expansion.
+        Any recall ``filter_ast`` is threaded into the vector-only path so the
+        degraded result honors the same filter as the primary path.
         """
         logger.info("Using vector-only fallback due to graph search failure")
 
@@ -2047,6 +2051,9 @@ class VectorCypherRetriever:
         # *something* rather than an empty response - the metadata still
         # tracks the failure via ``graph_unavailable``.
         fallback_mode = SearchMode.HYBRID if mode == SearchMode.GRAPH else mode
+        # If filter compilation fails here, the exception propagates intentionally:
+        # a filter the vector channel cannot honor must fail loud rather than
+        # return filter-violating chunks, matching the graph-path contract.
         result = await self._simple_retrieve(
             query=query,
             query_embedding=query_embedding,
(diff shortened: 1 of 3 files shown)


filippetrovic:

Reviewers: security, performance, correctness, tests, docs, devil's advocate + dedicated security audit.

No CRITICAL or HIGH must-fix findings. The fix is surgical (filter threading + test flip + audit-gate enforcement), all threading paths were verified, and the flipped enforcement test is genuine (non-vacuous spy, canonical-hash equality).

### MEDIUM
- **[Devil's Advocate / design]** `src/khora/engines/vectorcypher/retriever.py:2054-2056` — Fail-loud-on-filter-compile-error now lives *inside* a degradation handler: a filter the vector channel cannot compile fails the whole query instead of returning unfiltered-but-usable chunks. **Lead assessment:** intentional and consistent with the engine-wide contract (an unsupported filter raises `RecallFilterUnsupportedError` rather than silently degrading — returning filter-violating chunks is the failure mode this PR exists to eliminate). The propagation is documented in-code. An alternative (catch + `compile_python` post-filter + degradation counter) is a viable follow-up if operational data ever shows this path firing.

### LOW
- **[Tests]** `tests/unit/engines/vectorcypher/test_filter_pushdown_partial_failure.py` — The flipped Arm-C test enforces the dispatch contract (AST reaches the fallback unchanged); end-to-end *result-set* filtering on the degraded path is covered by the integration-level spies, not this unit. A live-stack enforcement test for the fallback's results would be a nice follow-up.
- **[Tests]** No explicit test for `mode=GRAPH` → `HYBRID` downgrade interacting with a filter inside the fallback, nor for a filter-compile error raised from within the fallback. Both are orthogonal to the threading fix; follow-up material.
- **[Scope]** The embedded BM25 accept-but-don't-enforce boundary remains open (tracked in the audit gate's enforcement-pending list). Correctly out of scope here; noting for coherence.

### Security Audit
- Scope: 48 added lines. No new sources, sinks, entry points, or dependencies. Postgres filter compilation uses parameterized binds (`sa.literal()`, system-key whitelist); Cypher uses `$param` placeholders; the BM25 coordinator fallback explicitly refuses unfiltered fallback under a filter. The change *closes* a data-exposure gap (filter-violating chunks returned on transient graph errors).
- **Verdict: PASS** — no findings.

### Reviewer notes (no action needed)
- **Correctness:** all three `_simple_retrieve` call sites verified to pass `filter_ast`; `filter_ast=None` default preserves prior behavior for unfiltered callers.
- **Performance:** no recompilation added (the AST object is threaded, not re-parsed); pushdown reuses existing indexes; no new round trips.
- **Docs:** docstrings and audit-gate comments updated consistently; no architecture-doc drift; no internal identifiers in the diff.

msrdic (approved):

PR #1080, opened 2026-06-10
Enforce the recall filter on the embedded BM25 channel

filippetrovic:
## Summary
- The embedded SQLite+LanceDB full-text (BM25 / keyword) search path, `SQLiteLanceTemporalStore.search_fulltext`, accepted a recall-filter AST for protocol parity but did **not** forward it into the inner BM25 search. As a result, a `recall(filter=...)` over the embedded stack could let filter-excluded chunks leak through the BM25/keyword channel into RRF fusion whenever that channel was enabled. The vector channel and the PostgreSQL backend already enforced the filter; the embedded BM25 channel was the gap.
- `search_fulltext` now forwards `filter_ast` into `_bm25_search`, which pushes the compilable leaves into the SQL `WHERE` and re-checks the remainder against the decoded chunk via the Python post-filter oracle — bringing the embedded BM25 channel to parity with the vector channel and the Postgres sibling.
- Flips the test scaffolding that documented the gap: removes the strict-xfail enforcement probe (now passes), empties the enforcement-pending boundary entry in the filter audit gate, and inverts the source-pinned guard to assert the forward (so a regression that drops the filter again fails the suite).
- Adds a CHANGELOG entry under Fixed.

## Test plan
- [x] Unit tests pass — filter audit gate: 5 passed
- [x] Integration tests pass — embedded enforcement matrix: 6 passed (including the formerly-xfail enforcement probe); fresh-DB FTS5 regression: 5 passed
- [x] Non-vacuity verified — reverting the one-line forward makes the enforcement test fail (a filter-excluded chunk leaks through BM25); restoring it passes
- [ ] Full CI suite (runs on this PR)


filippetrovic:

**Overall assessment: ✅ APPROVE** — no CRITICAL or HIGH findings across 7 reviewer domains + a dedicated security audit. The one-line `filter_ast` forward is a surgical, architecturally-sound parity fix; the test scaffolding is flipped correctly and proven non-vacuous.

### Reviewer outcomes
| Domain | Verdict | Notes |
|---|---|---|
| Security | ✅ Clean | Closes a data-exposure leak (filter-excluded chunks previously reached RRF via the BM25 arm). No injection surface — AST compiles to positional binds; FTS5 query is phrase-escaped. |
| Correctness | ✅ Clean | Forward is sufficient (inner `_bm25_search` builds its own post-filter when none supplied); `filter_ast=None` path unchanged (match-all, no regression); edge cases handled. |
| Database/query | ✅ Clean | Aliased (`c.`) WHERE pushdown correct; param order sound (LIMIT bind last); superset-safe (unsupported leaves defer to the `compile_python` oracle, never wrongly exclude). |
| Performance | ✅ Clean | Over-fetch (`limit*4` when narrowing) is bounded with early-exit; ~O(4·limit) decodes, not O(corpus). Matches the vector channel's pattern. |
| Test quality | ✅ Clean | Enforcement probe non-vacuous (contradiction filter → empty recall; revert→leak proven); inverted source-pinned AST guard catches regression. |
| Docs | ✅ Clean | Docstrings + CHANGELOG accurate; no stale "wiring-only / does not compile" claims remain anywhere in the tree. |
| Devil's advocate | ✅ Approve | Forward (vs. compile-at-call-site) is correct — it reuses the single post-filter build and avoids duplication in RRF/thin-vector reuse. Scope is right. |

### Findings (all non-blocking)

**MEDIUM**
- **[Test] M1** `tests/integration/matrix/test_filter_enforcement_sqlite_lance.py` — Enforcement is proven with a *system-key* contradiction filter (`source_name`). A *metadata-leaf* contradiction case (e.g. `{"metadata.custom.team": "no-such-team"}`) isn't exercised on the embedded stack; metadata enforcement runs through the same shared `compile_python` oracle (covered on the Postgres sibling), so this is a coverage-completeness follow-up, not a correctness gap.
- **[Docs] M2** `CHANGELOG.md` — Entry is accurate but doesn't explain *why* forward-the-AST vs. compile-at-call-site (the forward lets the inner search build the post-filter once). Optional clarity nicety.

**LOW**
- **[Database] L1** When SQLite `JSON1` is unavailable, metadata-leaf predicates get zero SQL pushdown and fall entirely to the post-decode filter — correct (superset-safe) but worth operator awareness for embedded perf.

### Security audit (OWASP / deps / crypto)
Scanned all added/modified lines across the 16 categories. **PASS** — no injection/auth/SSRF/secret/deserialization/file-handling exposure; no dependency or manifest changes. The change is a strict-narrowing filter-enforcement fix that *closes* a prior filter-bypass data-exposure path.

---
*Reviewed by a 7-role agent team (security, performance, correctness, test, database, docs, devil's advocate) plus a dedicated security audit. No blocking changes requested.*

probablyrobot (approved):

PR #1083, opened 2026-06-10
Add slow/e2e recall-filter CI matrix across seven engine configs

filippetrovic:
## Summary
- Add a dedicated slow/e2e CI workflow (`.github/workflows/e2e.yml`) that runs the deterministic `remember()` → `recall(filter=)` row-set harness across **seven engine configurations**: VectorCypher (full PG+Neo4j and embedded sqlite+LanceDB), Skeleton on pgvector / surrealdb / weaviate / sqlite_lance, and Chronicle. It runs concurrently with the fast suite as its own check, provisions each container leg via an `if:`-guarded `docker run` (no idle containers on the container-free legs), fails loud when a required store is unreachable, and uses two-level parallelism (engine matrix × `-n auto`) with no network.
- **Enable embedded occurred-bounds temporal recall** on the VectorCypher engine: the embedded backend now wires a `start_time`/`end_time` (occurred_at) filter into the vector query instead of raising `NotImplementedError`. The over-broad point-in-time fail-fast is removed; a structured degradation is recorded only at the genuine entity-version no-op site — point-in-time entity-version queries (which the embedded schema cannot answer) degrade with a warning + an internal counter rather than failing the whole recall. This closes the embedded filter-pushdown gap that was previously a strict xfail.
- Add four Skeleton e2e fixtures and a backend resolver with raise-on-empty + shipped-precedent equality guards, so a misconfigured lane fails loud instead of silently under-covering; plus five per-lane row-set test modules. `include_system_keys` is selected per backend based on whether that backend's chunk row carries the denormalized document system keys.

## Test plan
- [x] Unit + embedded suites pass: 381 passed, 84.0% coverage (gate 77%)
- [x] Container-free e2e lanes pass locally (surrealdb / sqlite_lance / vc_embedded): 102 passed
- [ ] Container lanes (PG / Neo4j / Weaviate) verified by the new e2e workflow in CI
- [x] `ruff format` + `ruff check` clean; `ty` typecheck unchanged


filippetrovic:
## Consolidated Review — multi-perspective agent team

8-reviewer team (security, performance, correctness, test-quality, database, observability, docs, devil's-advocate) + a dedicated security audit. Overall: **APPROVE** — no CRITICAL/HIGH blocking findings remain. The two doc findings below were addressed in `f2bb2b70`; the one MEDIUM is a tracked post-merge follow-up. All 7 e2e matrix legs are green in CI.

### HIGH
- **[Docs]** `tests/e2e/DESIGN_NOTES.md` — the "CI gating + deferred lanes" section described the dedicated slow/e2e job and the Skeleton lanes as future follow-ups, which this PR ships. **Fixed in `f2bb2b70`** (rewritten to describe the 7-leg matrix, the `KHORA_E2E_BACKEND` resolver, and the serial/parallel worker split).

### MEDIUM
- **[Docs]** Embedded temporal behavior change (raise → degrade) lacked a user-facing note. **Fixed in `f2bb2b70`** (CLAUDE.md gotcha added next to the bi-temporal note).
- **[Database]** `src/khora/engines/skeleton/backends/pgvector.py::connect()` re-runs schema DDL (`metadata.create_all` + `CREATE OR REPLACE FUNCTION` + trigger) without an advisory lock, which is why concurrent `-n auto` workers deadlocked. The shipped fix (run the live shared-store legs `-n 0`, matching `ci.yml`'s `test-integration`) is valid and idiomatic. **Recommended follow-up (separate ticket):** wrap the schema-init block in a global `pg_advisory_xact_lock`, mirroring the pattern already in `storage/backends/pgvector.py`, to harden concurrent multi-process schema init and allow `-n auto`. Not PR-blocking.

### LOW
- **[Performance]** The 81-case serial legs (`vc_full`, `chronicle`) were flagged as a possible 25-min-timeout risk. The deterministic harness stubs the LLM/embedder (no network), so the live legs complete in ~2–3 min in CI (`vc_full` 2m54s, `chronicle` 2m17s, `skeleton_pgvector` 2m16s) — comfortably within budget. Non-issue; monitor on slower runners.
- **[Selection]** `lane_skip` is applied to all 8 slow modules AND the workflow uses per-leg `-k`/path selection — redundant defense-in-depth. Intentional (the token keys always agree, so no leg can go empty); kept deliberately.

### CLEAN (no findings)
- **[Security]** No real secrets (PG/Neo4j/Weaviate creds are standard throwaway test creds matching `ci.yml`); no injection/SSRF/deserialization sinks; `neo4j_password` correctly `SecretStr`; Actions pinned `@v6/@v7` consistent with the repo's other workflows.
- **[Correctness]** The temporal-gate change is sound — the degradation fires only on the genuine point-in-time entity-version no-op site (`_neo4j_driver is None`); a plain occurred-bounds `start_time` recall routes SIMPLE and stays degradation-free. `_E2E_BACKEND_MAP` `include_system_keys` flags match each backend's chunk-row columns; the `-n` postgres/auto split covers exactly the 4 live-store legs; anti-vacuity guards (raise-on-empty, Layer-3 shipped-precedent equality) are correct.
- **[Observability]** `khora.vectorcypher.version_filter.degraded_total` matches the contract exactly, tagged `internal`, `reason` label only (no `namespace_id`); a structured `Degradation` at WARNING with no `exc_info` (condition-driven). Drift gate satisfied.
- **[Test-quality]** New lane modules use HYBRID mode, fresh namespace-per-test, anti-vacuity + Layer-3 guards; the temporal test uses two-sided assertions (old ABSENT + recent PRESENT) plus a control test proving the bound (not vector top-k) is the narrowing force; the flipped `NotImplementedError` tests assert the new degrade/no-raise behavior faithfully.
- **[Devil's-advocate]** Closing the embedded temporal xfail is in the ticket's stated scope (the embedded leg's reason for inclusion). The 7-leg matrix is complementary to the conformance matrix: conformance direct-seeds per backend; this proves the full `remember()`→`recall(filter=)` ingest path end to end. `-n 0` on shared-store legs is the right call over a backend src change in a CI-focused PR.

## Security Audit
**Verdict: PASS.** Scanned all added lines across the 16 categories. No CRITICAL/HIGH/MEDIUM. INFO only: throwaway test credentials in the workflow `docker run` blocks (match `ci.yml`/`filter-conformance.yml`); Actions pinned to major-version tags consistent with repo convention (SHA-pinning would be a repo-wide hardening, not specific to this PR); the single `urlparse` operates on the server-controlled `WEAVIATE_URL` env var (not user input → no SSRF).

---
**Assessment: APPROVE.** Doc findings fixed; the advisory-lock hardening is a tracked follow-up; CI green across all 7 e2e legs + the conformance matrix.

filippetrovic:
### Review addendum — final two perspectives

The test-quality and devil's-advocate reviewers completed; their findings confirm the consolidated review above, with these specifics:

- **[Test-quality] CLEAN.** The temporal test fix is genuine — `test_vc_temporal_filter` drops the xfail, stamps `occurred_at` via the real `source_timestamp` ingest path (no raw-SQL backdating), asserts two-sided (recent PRESENT + old ABSENT), and the control test seeds only the back-dated doc → asserts EMPTY plus a without-floor sanity check, so it cannot pass vacuously. The flipped `NotImplementedError` unit/integration tests assert the new degrade/no-raise behavior (no assertions merely deleted), and `test_no_driver_records_degradation` covers the new path. *LOW nit:* a comment in `test_filter_enforcement_sqlite_lance.py` describing the **old** "fails fast" behavior reads slightly backwards; the test code itself is correct.
- **[Devil's-advocate] Four design/scope challenges, none blocking:**
  1. *`-n 0` vs the parallelism goal* (MEDIUM design call) — dispositioned above: serial-on-shared-store is the idiomatic fix (matches `ci.yml`); the advisory-lock alternative is a tracked follow-up. The PR **title** does not claim two-level parallelism, so no rename is needed; the engine-matrix level stays parallel and the 3 isolated legs keep `-n auto`.
  2. *`-k` × `lane_skip` redundancy* (LOW) — intentional, documented defense-in-depth; consolidation is a low-priority cleanup.
  3. *Embedded temporal change "scope"* (MEDIUM) — this is **in-scope, not creep**: closing the embedded filter-pushdown xfail is an explicit, defined deliverable of this change (the embedded VectorCypher leg exists precisely to exercise it). It is intentionally shipped together with the leg that guards it.
  4. *e2e vs filter-conformance overlap* (LOW) — complementary, and the value-add is documented in `tests/e2e/DESIGN_NOTES.md`: conformance direct-seeds per backend (wiring proof); this harness drives the **real `remember()`→`recall(filter=)` ingest pipeline** end to end (row-set proof), and adds four Skeleton backend combinations not previously covered end-to-end.

No new blocking findings. **Assessment stands: APPROVE** — the one MEDIUM (advisory-lock hardening) is a tracked post-merge follow-up.

probablyrobot (approved):

PR #1084, opened 2026-06-10
Keep deprecated recall time bounds on the recency-window axis

filippetrovic:
## Summary
- The deprecated `start_time`/`end_time` recall bounds previously folded into an `occurred_at` event-time filter **in addition** to the recency-window `temporal_filter`. On the Chronicle engine, chunks ingested without an explicit timestamp carry no `occurred_at` or `source_timestamp` anchor, so the AND-ed event-time post-filter dropped them even when they were recent by ingest time — a false-empty of in-scope results.
- These bounds now forward **only** the recency-window `temporal_filter` (`COALESCE(source_timestamp, created_at)`). The change is purely subtractive in the facade shim.
- The public `filter={"occurred_at": ...}` API is unchanged and keeps its documented event-time semantics (no `created_at` fallback) — the two paths now correctly diverge: the window axis includes anchor-less recent chunks, the event-time axis excludes them.
- Skeleton/pgvector is unaffected: it applies `temporal_filter` independently of `filter_ast`, and its lake always populates `occurred_at`, so it never had the false-empty.

## Test plan
- [x] Unit: facade shim (deprecated bounds → window only, `filter_ast` None), window-axis edge cases (start-only / end-only / inverted / naive→UTC), and the rewritten temporal-bounds suite
- [x] Engine-level guard: an anchor-less chunk (no `occurred_at`/`source_timestamp`, recent `created_at`) survives the deprecated-bounds window path on Chronicle, alongside its event-time-filter exclusion contrast
- [x] Public-filter boundary semantics: `$gte` inclusive / `$lt` exclusive, proven non-vacuous via operator-flip controls
- [x] Embedded sqlite_lance integration leg green (occurred_at-derived-from-source_timestamp path; no conflation)
- [x] Regression guards proven non-vacuous (fail on pre-fix code, pass after)
- [ ] Full integration matrix (live PostgreSQL / Neo4j) — runs in CI

### DIFF
--- a/src/khora/khora.py
+++ b/src/khora/khora.py
@@ -2022,19 +2022,23 @@ class Khora:
         from khora.filter import RecallFilter, canonical_hash, parse_to_ast
 
         try:
-            # Resolve the recall filter once, at the facade. Two inputs feed the
-            # single canonical AST: the public ``filter=`` kwarg and the
-            # deprecated ``start_time``/``end_time`` bounds. They are mutually
-            # exclusive. ``temporal_filter`` is still built from the deprecated
-            # bounds so the engines' existing temporal behavior (recency
-            # weighting, version filter) is unchanged. The skeleton engine now
-            # threads ``filter_ast`` through to its temporal store, where the
-            # pgvector backend compiles it to a WHERE predicate (the other
-            # backends accept-and-ignore it). When the deprecated bounds are
-            # used, both ``temporal_filter`` and the folded AST are AND-ed on
-            # the pgvector path, so the folded bounds below mirror
-            # ``temporal_filter``'s boundary semantics exactly to stay
-            # consistent and avoid dropping boundary rows.
+            # Resolve the recall filter once, at the facade. The public
+            # ``filter=`` kwarg and the deprecated ``start_time``/``end_time``
+            # bounds are mutually exclusive but feed DIFFERENT axes on purpose:
+            #
+            #   * ``filter=`` produces a canonical ``filter_ast`` the engines
+            #     compile and post-filter, so an ``occurred_at`` predicate is
+            #     enforced against the EVENT-time axis (no created_at fallback) —
+            #     the documented event-time semantics of the new API.
+            #   * ``start_time``/``end_time`` produce ONLY a ``temporal_filter``
+            #     recency window. They are a window-axis API: every engine narrows
+            #     its channel reads on ``COALESCE(source_timestamp, created_at)``,
+            #     so a chunk recent by ingest time survives even when it carries no
+            #     event-time anchor. Folding these bounds into an ``occurred_at``
+            #     ``filter_ast`` (as an earlier revision did, speculatively) would
+            #     AND an event-time post-filter on top of the window and false-empty
+            #     every anchor-less chunk a plain ``remember()`` produces. They must
+            #     NOT set ``filter_ast``.
             temporal_filter: Any = None
             filter_ast: Any = None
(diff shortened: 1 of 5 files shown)


filippetrovic:

Six independent reviewers (Security, Performance, Correctness, Test, Docs, Devil's Advocate) plus a dedicated security audit. **No CRITICAL or HIGH defects.** The core fix is correct, minimal, and well-guarded.

Removing the speculative `occurred_at` `filter_ast` fold from the deprecated `start_time`/`end_time` path is the right, minimal fix. Those bounds are a **recency-window** API (`COALESCE(source_timestamp, created_at)`); folding them onto the **event-time** axis (no `created_at` fallback) AND-ed a post-filter that false-emptied anchor-less chunks on the Chronicle engine. The public `filter={"occurred_at": …}` API is untouched and keeps its event-time semantics. The two axes now correctly diverge, and the guards prove it.

### Per-domain results
| Domain | Result |
| --- | --- |
| Security | ✅ No findings — change is security-positive (eliminates a filter-composition footgun); no new input/sink/secret/dep |
| Performance | ✅ No regression — removes a redundant post-filter pass; SQL window narrowing unchanged |
| Correctness | ✅ PASS (high confidence) — axis distinction sound; inverted-range stays empty-matching; naive→UTC coercion correct; no placeholders |
| Test quality | ✅ Guards genuinely **bite** (fail if the fold returns) and the engine-level survival assertion is a true positive, not vacuous |
| Docs | ✅ Comments/docstrings accurate; deprecation guidance correct; repo-hygiene clean |
| Security audit | ✅ PASS — 16-category scan of added lines clean (pattern hits were substring false positives) |

### MEDIUM (advisory — non-blocking)
- **M1 · [Devil's Advocate/Scope]** `src/khora/khora.py` — the fix is at the shared facade, so it changes the deprecated-bounds path for **all** engines, but explicit new tests cover Chronicle + the facade. *Mitigation:* the change is purely subtractive (restores pre-fold behavior); skeleton/pgvector was verified to apply `temporal_filter` independently of `filter_ast`; and the CI conformance matrix exercises the other engines (vectorcypher/cypher/postgres/sqlite_lance/surrealdb/weaviate). Optional follow-up: add an explicit VectorCypher deprecated-bounds regression case.
- **M2 · [Devil's Advocate/Process]** Deprecated `start_time`/`end_time` no longer apply event-time exclusion — a (bugfix) behavior change for a deprecated path. It is documented in the commit message and this PR description (the changelog of record for this repo); no separate release-notes edit is warranted for a deprecated-path bugfix.

### LOW (optional)
- **L1 · [Test]** The facade capture tests build `Khora` via `__new__`, creating a maintenance coupling to the class shape. Common, documented pattern — acceptable; revisit if `Khora.__init__` changes.
- **L2 · [Devil's Advocate/Test]** The facade shim assertions could be consolidated into parametrized cases. Current tests are load-bearing and clear; consolidation is cosmetic.

### Considered and not pursued
- **Ingest-layer alternative** (default-populate `occurred_at` at write): rejected — `source_timestamp` is also `None` for anchor-less chunks, and falling back to `created_at` would violate the deliberate "ingest time ≠ event time / no created_at fallback" contract that the public `occurred_at` filter depends on.
- **Per-engine control of the fold**: strictly more complex than the subtractive facade fix and not required; the facade is the correct single decision point for the window-vs-event-time axis.

_Reviewed by an automated multi-perspective review team. MEDIUM/LOW items are advisory; none block merge._

probablyrobot (approved):

Issue #778, opened 2026-05-20
Track upstream fixes for security CVE allowlist (post-v0.16.0)


probablyrobot:
## Background

PR for `ib/security-cve-allowlist-v0.16.1` adds an `--ignore-vuln` allowlist to the `security` CI job covering 20 upstream CVEs in pinned ML dependencies (joblib 1.5.3, transformers 5.8.0, torch 2.11.0). None of these have an upstream fix version yet, so the only way to keep the `security` job functioning as an early-warning gate for **new** CVEs was to allowlist the unfixable existing ones with per-CVE justifications.

This issue tracks the allowlist itself — each entry should be removed once upstream ships a fix and we bump the dep. **Revisit weekly.**

## The 20 CVEs

| CVE | Package | Vulnerable codepath | Why khora is not exposed |
| --- | --- | --- | --- |
| PYSEC-2024-277 | joblib 1.5.3 | `joblib.numpy_pickle::NumpyArrayWrapper().read_array()` deserialization (disputed; only triggers when caching of trusted content is reloaded) | khora never imports joblib; transitive dep of scikit-learn / sentence-transformers, no `.joblib` files loaded |
| PYSEC-2025-211 | transformers 5.8.0 | Perceiver model checkpoint deserialization RCE | khora never instantiates Perceiver; only `sentence_transformers.CrossEncoder(<vetted_model_id>)` |
| PYSEC-2025-212 | transformers 5.8.0 | Transformer-XL checkpoint deserialization RCE | khora never loads Transformer-XL |
| PYSEC-2025-213 | transformers 5.8.0 | megatron_gpt2 checkpoint deserialization RCE | khora never loads megatron_gpt2 |
| PYSEC-2025-214 | transformers 5.8.0 | SEW `convert_config` code injection RCE | khora never runs checkpoint conversion; only `from_pretrained` on vetted IDs |
| PYSEC-2025-215 | transformers 5.8.0 | SEW-D `convert_config` code injection RCE | same — no conversion |
| PYSEC-2025-216 | transformers 5.8.0 | HuBERT `convert_config` code injection RCE | same — no conversion |
| PYSEC-2025-217 | transformers 5.8.0 | X-CLIP checkpoint conversion deserialization RCE | khora never loads X-CLIP or runs conversion |
| PYSEC-2025-218 | transformers 5.8.0 | GLM4 checkpoint deserialization RCE | khora never loads GLM4 |
[...]

PR #1090, opened 2026-06-10
Add update_last_accessed to the SurrealDB vector adapter

filippetrovic:
## Summary
- Implement `update_last_accessed()` on the SurrealDB vector adapter so recall-reinforcement can stamp `last_accessed_at` on recalled chunks. Previously this was a silent no-op on the SurrealDB backend — the pgvector and sqlite_lance adapters already implement it, so reinforcement-on-recall had no effect there.
- Mirror the sibling adapters' contract exactly: same name and signature, empty `chunk_ids` short-circuits to `0`, the write is namespace-scoped (a forged cross-namespace id touches zero rows), and the method returns the affected row count.
- Scope the UPDATE deterministically: resolve the namespace to its record id(s) up front and match the chunk's `namespace` link by direct RID equality, rather than a live record-link dereference. This keeps the cross-tenant guard deterministic on the embedded engine.
- Mark surrealdb as a supported backend in the reinforcement-on-recall capability gate (the chunk schema already declares `last_accessed_at`); the gate previously rejected it.
- Tests: embedded round-trip (tz-aware + microsecond precision, no sibling over-stamping), cross-namespace rejection (rowcount 0 and stamp unchanged), empty-list no-op, partial-match rowcount, and a dispatch-capability unit test with a negative control. The test fixture warms the embedded engine to avoid a rare cold-start non-determinism.

## Test plan
- [x] New dispatch-capability unit test passes (positive + negative control)
- [x] New embedded SurrealDB integration tests pass (round-trip, cross-namespace, empty-list, partial-match) — full module green across repeated serial runs
- [x] `ruff format`, `ruff check`, and `ty` type-check clean
- [ ] CI: unit + integration suites pass

### DIFF
--- a/src/khora/engines/chronicle/engine.py
+++ b/src/khora/engines/chronicle/engine.py
@@ -871,8 +871,7 @@ class ChronicleEngine:
         # silently no-op the UPDATE - users on those backends would see a
         # config flag with no effect. Fail loudly at connect time instead so
         # the operator knows reinforcement is unavailable on this stack.
-        # Currently supported: pgvector, sqlite_lance. NOT supported: surrealdb
-        # (chunk schema lacks a last_accessed_at field).
+        # Currently supported: pgvector, sqlite_lance, surrealdb.
         qs = self._config.query
         reinforcement_on = getattr(qs, "chronicle_enable_recall_reinforcement", False) if qs else False
         if reinforcement_on:
@@ -883,7 +882,7 @@ class ChronicleEngine:
                     "Chronicle reinforcement-on-recall "
                     "(KHORA_QUERY_CHRONICLE_ENABLE_RECALL_REINFORCEMENT=true) "
                     f"requires the vector backend to implement update_last_accessed; "
-                    f"{backend_name} does not. Supported backends: pgvector, sqlite_lance."
+                    f"{backend_name} does not. Supported backends: pgvector, sqlite_lance, surrealdb."
                 )
 
         # Create embedder
--- a/src/khora/storage/backends/surrealdb/vector.py
+++ b/src/khora/storage/backends/surrealdb/vector.py
@@ -261,32 +261,37 @@ class SurrealDBVectorAdapter:
     ) -> int:
         """Stamp ``last_accessed_at = ts`` on the given chunks.
 
-        Single UPDATE statement, scoped to ``namespace_id`` to prevent
-        cross-tenant writes through forged ids. Returns the row count.
-        Used by the Chronicle reinforcement-on-recall path.
+        Scoped to ``namespace_id`` to prevent cross-tenant writes through
+        forged ids. Returns the row count. Used by the Chronicle
+        reinforcement-on-recall path.
 
-        The namespace is resolved to its record id(s) up front and the
-        chunk's ``namespace`` link is matched by *direct* RID equality
-        (``namespace IN $ns_rids``) rather than the sibling read methods'
-        ``OR namespace.namespace_id = $ns_str`` deref. The live record-link
-        dereference is non-deterministic on the embedded engine, so it is
-        avoided here to keep the cross-tenant guard deterministic — a flaky
-        security predicate is worse than a missed reinforcement.
+        Namespace matching mirrors the deterministic ``namespace = $ns_rid``
+        arm of the sibling read methods, by direct RID equality. Chunks store
+        their ``namespace`` link as ``rid(namespace_id)`` where ``namespace_id``
+        is whatever the caller passed at ingest — the public API resolves to
+        the stable ``namespace_id`` (so the link is ``rid(stable_id)``), while
+        the row itself lives at ``rid(row_id)``. The caller's own RID is matched
+        directly (the production case: chunk linked by stable id, caller passes
+        stable id), and the namespace is additionally resolved to its existing
+        row id(s) to cover the mixed case (chunk linked by row id, caller
+        passing the stable id). No ``namespace.namespace_id`` link dereference.
         """
         if not chunk_ids:
             return 0
         chunk_rids = [_rid("chunk", cid) for cid in chunk_ids]
-        # Resolve the namespace to its record id(s) up front and match the chunk's
-        # ``namespace`` link by direct RID equality. This avoids the non-deterministic
-        # chunk->namespace ``.namespace_id`` dereference (flaky on the embedded engine)
-        # while still honoring both the row-id and stable-id namespace forms.
+        # Match by direct RID equality, with no live record-link dereference.
+        # Start from the caller's own RID (covers chunks linked by the same id
+        # form the caller passes — the production path, where both are the
+        # stable namespace_id), then add any resolved namespace row id(s) to
+        # cover the mixed case (chunk linked by row id, caller passing the
+        # stable id). The caller's RID is always present, so an absent namespace
+        # row does not short-circuit a write to chunks carrying that link.
+        ns_rids = [_rid("memory_namespace", namespace_id)]
         ns_rows = await self._conn.query(
             "SELECT id FROM memory_namespace WHERE namespace_id = $ns_str OR id = $ns_rid",
             {"ns_str": str(namespace_id), "ns_rid": _rid("memory_namespace", namespace_id)},
         )
-        ns_rids = [row["id"] for row in ns_rows]
-        if not ns_rids:
-            return 0
+        ns_rids.extend(row["id"] for row in ns_rows)
         rows = await self._conn.query(
             "UPDATE chunk SET last_accessed_at = $ts WHERE id IN $ids AND namespace IN $ns_rids",
             {"ts": ts, "ids": chunk_rids, "ns_rids": ns_rids},
--- a/src/khora/storage/backends/surrealdb/vector.py
+++ b/src/khora/storage/backends/surrealdb/vector.py
@@ -253,6 +253,46 @@ class SurrealDBVectorAdapter:
 
         return count
 
+    async def update_last_accessed(
+        self,
+        namespace_id: UUID,
+        chunk_ids: list[UUID],
+        ts: datetime,
+    ) -> int:
+        """Stamp ``last_accessed_at = ts`` on the given chunks.
+
+        Single UPDATE statement, scoped to ``namespace_id`` to prevent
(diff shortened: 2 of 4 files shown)


filippetrovic:

Reviewed by a 7-perspective team (security, performance, correctness, database, tests, docs, devil's advocate) plus a focused security audit. Overall: **APPROVE** — no CRITICAL or HIGH findings. The implementation is sound; remaining items are advisory.

### Severity summary
- **CRITICAL:** none
- **HIGH:** none
- **MEDIUM:** M1–M3 (advisory — approach/documentation, not bugs)
- **LOW:** L1

### Findings

**[Security] — CLEAN.** The UPDATE is namespace-scoped via up-front resolution (`SELECT id FROM memory_namespace WHERE namespace_id = $ns_str OR id = $ns_rid`) then `WHERE id IN $ids AND namespace IN $ns_rids`. A forged cross-namespace chunk id matches zero rows (the chunk's `namespace` RID points elsewhere). All values are parameterized (`$ts/$ids/$ns_rid/$ns_str/$ns_rids`) — no SurrealQL injection. Cross-namespace test asserts both rowcount 0 and stamp unchanged. No secrets/internal identifiers in the diff.

**[Correctness] — APPROVED.** `len(rows)` is a faithful matched-row count (SurrealDB UPDATE returns the rows it set); empty `chunk_ids` short-circuits to 0; namespace resolution handles both row-id and stable-id forms (no false-negative reinforcement). The engine.py change is text-only (gate logic unchanged). No placeholders.

**[Database] — PASSED.** Matching `namespace IN $ns_rids` (direct RID equality) is correct against the stored `namespace` record link; the resolution SELECT is index-covered (`namespace_id`). No migration needed (declarative schema; `last_accessed_at` already defined).

**[Performance] — ACCEPTABLE.** Two queries (resolve + update) vs the siblings' one, but the method runs once per recall as a fire-and-forget task with batched `chunk_ids` (no N+1), and the resolution SELECT is indexed. The extra round-trip is an acceptable cost for a deterministic isolation predicate on a best-effort write.

**[Tests] — STRONG.** Round-trip (tz-aware + microsecond precision, sibling not over-stamped), cross-namespace IDOR (rowcount 0 + stamp unchanged), empty-list no-op, partial-match rowcount (real + nonexistent id → 1, discriminates `len(rows)` from `len(chunk_ids)`), and a dispatch unit test with a negative control proving the gate is live. Seed/new timestamps are distinct so positives can't pass by coincidence. The fixture warm-up uses a **throwaway namespace** (namespace-scoped, so it cannot affect any test's data) — it improves per-test self-containment rather than coupling tests.

**[Docs] — CLEAN.** Docstring accurately explains the deterministic RID-match; the `ChronicleEngine` reinforcement-gate comment and `ConfigurationError` message now correctly list surrealdb. No stale "SurrealDB unsupported for reinforcement" text remains in `docs/`. No internal identifiers leaked.

**[Security Audit] — no additional findings.** No new dependencies, no crypto, no auth flows; the only inputs are typed (`UUID`/`datetime`) and parameterized. Attack surface is limited to the tenant-isolation predicate, which is covered above.

### Devil's Advocate (advisory, non-blocking)
- **M1 — Divergence from the sibling read idiom.** `update_last_accessed` resolves the namespace and matches by direct RID equality instead of the read methods' live `namespace.namespace_id` dereference. This is a **deliberate hardening** for a tenant-isolation *write*: the embedded engine's live link-deref was observed to be non-deterministic under cold-start/concurrent load during development, and a flaky isolation predicate is worse than a missed best-effort reinforcement. Endorsed by the security, correctness, database, and performance reviewers. Kept by design; the docstring documents the rationale.
- **M2 — Extra resolution SELECT.** Same rationale as M1; the single-statement form would reintroduce the live deref on the write path. Disposition: keep.
- **M3 — Fixture warm-up treats a symptom.** Fair point that the embedded cold-start non-determinism is ultimately an engine-level artifact. Disposition: keep the warm-up (it makes the test reliable and is self-contained); a short note in the repo's SurrealDB gotchas documenting the embedded cold-start/contention non-determinism is a reasonable **follow-up**.

### LOW / follow-ups
- **L1 — Vestigial `namespace_id` column on `chunk`** (pre-existing): declared in the schema but left NULL by the create path, while `delete_chunks_by_document` filters on it. Out of scope here; worth a separate cleanup ticket to populate or drop it.

**Verification:** full integration module passes repeatedly under serial execution (how CI runs integration); the unit dispatch test passes; `ruff format`/`ruff check`/`ty` clean. CI integration + conformance (incl. the SurrealDB leg) are green on this PR.

probablyrobot (approved):

filippetrovic:
`update_last_accessed` returns 0 rows for every namespace created through the public API. `MemoryNamespace.id` and `MemoryNamespace.namespace_id` are independent `uuid4` defaults (`core/models/tenancy.py`), and `SurrealDBRelationalAdapter.resolve_namespace` resolves callers to the **stable** `namespace_id` — which is what `recall`/`remember` use, and what chunks store as `namespace = rid(stable_id)`. That RID is dangling: the namespace row's id is `rid(row_id)`. The new guard resolves the namespace to *existing row ids* and requires `namespace IN $ns_rids`, so the chunk's dangling `rid(stable_id)` never matches.

Repro at this head (memory mode, warm-up included):

```python
ns = await coord.create_namespace(MemoryNamespace())
resolved = await coord.resolve_namespace(ns.namespace_id)   # stable id — what recall passes
# ingest chunk with namespace_id=resolved ...
await coord.get_chunk(chunk.id, namespace_id=resolved)       # found
await coord.update_last_accessed(resolved, [chunk.id], ts)   # rowcount 0, stamp unchanged
```

So reinforcement still silently no-ops on the production path — and now the connect-time gate passes and `_reinforce_last_accessed` increments the success counter, making the no-op invisible. The tests don't catch it because they all use `ns.id` (row-id form), where the link and the row id coincide.

Fix: add the caller's RID as a direct arm, the same deterministic predicate every read method already uses (`namespace = $ns_rid`, no link deref):

```python
ns_rids = [row["id"] for row in ns_rows]
ns_rids.append(_rid("memory_namespace", namespace_id))
# UPDATE chunk SET last_accessed_at = $ts WHERE id IN $ids AND namespace IN $ns_rids
```

Verified at this head: stamps land (rowcount 1, tz + microseconds intact) and the cross-namespace probe still returns 0. Keep the row-resolution query for the mixed form (chunks linked by row id, caller passing the stable id). Drop the early `if not ns_rids: return 0` short-circuit or it will still no-op when no namespace row exists in the vector store while chunks carry dangling links.

Please also add an integration test that ingests and reinforces via `coord.resolve_namespace(ns.namespace_id)` output — the only form the public API produces; it fails on the current head.

Separate from the fix: the `coord` fixture comment asserts the embedded engine can cold-start into "a namespace-scoped UPDATE transiently matching the wrong row". If that is reproducible it's an isolation bug that exists in production (no warm-up there) and needs an upstream issue referenced here, not a test-only workaround. If it isn't, remove the claim. Same for the docstring's "live record-link dereference is non-deterministic" — the read methods (`get_chunk`, `search_similar`) still rely on that deref as half their namespace predicate, so either substantiate it or reconcile.

filippetrovic:
Fixed in b4fd1185 — thanks, this was a real production-path no-op and the diagnosis was exactly right.

**Reproduced and fixed.** A chunk ingested via the public path (caller passes `resolve_namespace(namespace_id)` → the stable id, so the chunk's link is `rid(stable_id)`) was never matched: the guard resolved the namespace to its *row* id and required `namespace IN [rid(row_id)]`, which the dangling `rid(stable_id)` link never hits. `get_chunk` found the chunk (its predicate includes the direct `namespace = $ns_rid` arm) while `update_last_accessed` returned rowcount 0 — silent no-op, masked by the connect-time gate and the success counter.

**The fix** is your prescription: seed `ns_rids` with the caller's own RID (`_rid("memory_namespace", namespace_id)`) — the same deterministic `namespace = $ns_rid` arm every read method uses — and *additionally* keep the row-resolution query for the mixed form (chunk linked by row id, caller passing the stable id). The `if not ns_rids: return 0` short-circuit is gone, so a chunk carrying a dangling link is still stamped when no namespace row matches. Verified at this head: production-path stamp lands (rowcount 1, tz + microseconds intact); cross-namespace probe still returns 0 with the stamp unchanged.

**New test.** Added `test_update_last_accessed_stamps_via_resolved_namespace_id`, which ingests and reinforces through `coord.resolve_namespace(ns.namespace_id)` — the only form the public API produces. Mutation-checked: it FAILS against the old resolution-only guard (rowcount 0) while the existing `ns.id` (row-id form) tests pass on that same broken code — confirming it catches exactly this regression and that the prior tests structurally could not.

**On the non-determinism claims — you're right, removed both.** I reverted the fixture warm-up and dropped the "cold-start / live-deref is non-deterministic" assertions from the comment and docstring. The cross-namespace test is deterministic in true isolation (50/50 with no warm-up); the earlier flakiness I saw was contention from concurrent test processes I was running, not a cold-start artifact. And since `get_chunk`/`search_similar` rely on the `namespace.namespace_id` deref and work in production, asserting it's non-deterministic was unsubstantiated. The new docstring explains the real reason for the direct-RID match — the row-id-vs-stable-id linkage — with no engine-behavior claim.

The vestigial `chunk.namespace_id` column (L1) remains a separate pre-existing cleanup, out of scope here.

filippetrovic:
CI note: every check is green **except `security`**, and that failure is unrelated to this PR. The `security` job's `pip-audit` step now flags **CVE-2025-3000** in `torch 2.11.0` — a transitive dependency (via the sentence-transformers / cross-encoder reranker), not touched by this change. It's a newly-surfaced advisory: this job passed on the parent commit earlier today, and the base branch would fail it now too if re-run.

The CVE is MEDIUM (CVSS 5.3), local-only, in `torch.jit.script` list-attribute handling, with no fix version published — khora does not call `torch.jit.script` on user-defined classes, so it doesn't reach the vulnerable path. This fits the existing "upstream CVE with no fix" allowlist (the `--ignore-vuln` block in `ci.yml`, revisited weekly per #778). Recommend handling it there as a separate maintenance change rather than folding a shared-CI/security edit into this feature PR.

All substantive checks for this change pass: `test-unit`, `test-integration`, all conformance legs incl. `conformance (surrealdb)`, `lint`, `examples-smoke`, `codecov/patch`.

Issue #1088, opened 2026-06-10
Skeleton embedded: namespace_id stable-vs-row-level resolution gap breaks metadata-filtered recall


filippetrovic:
## Gap

On the embedded Skeleton stack (SQLite + LanceDB), a metadata/tag-filtered recall misses rows because of a namespace-id resolution mismatch at the engine boundary.

`kb.remember` resolves the public, stable `namespace_id` to the row-level `id` before persisting, so `khora_chunks.namespace_id` stores the row-level id. But `engine.recall` is called with the stable `namespace_id` straight through, so the recall filter compares the stable id against the stored row-level id and matches nothing.

## Affected test

`tests/integration/matrix/test_skeleton_sqlite_lance.py::test_skeleton_recall_with_metadata_filter` is the acceptance test (a tag filter should restrict recall to chunks carrying the requested tag). It is currently marked `@pytest.mark.xfail(strict=True)` and tracks this gap. The same root cause is noted behind the Postgres-side xfail of the sibling skeleton test.

## Proposed resolution

Either resolve the stable `namespace_id` to the row-level `id` at the engine recall boundary (matching the remember path), or update the engine/test contract so both sides agree on which id the chunk rows key on.

Issue #1085, opened 2026-06-10
VectorCypher embedded: point-in-time temporal recall (occurred-bounds temporal_filter) unsupported


filippetrovic:
## Gap

On the embedded VectorCypher stack (SQLite + LanceDB), a point-in-time temporal recall does not work. An occurred-bounds `temporal_filter` (e.g. derived from a `start_time` on `recall`) trips a fail-fast `NotImplementedError` in the retriever before any LanceDB-level temporal narrowing happens, so the recall raises rather than returning a time-filtered result.

## Affected test

`tests/integration/matrix/test_vectorcypher_sqlite_lance.py::test_vc_temporal_filter` is the end-to-end acceptance test for this behavior. It is currently marked `@pytest.mark.xfail(strict=True)` and tracks this gap. When the embedded backend honors point-in-time temporal queries, that xfail should be removed and the test should pass as the acceptance test for the work.

## Expected behavior

A `recall(..., start_time=...)` on the embedded VectorCypher stack should narrow results to chunks within the requested occurred-bounds window, consistent with the production (Postgres) stack.

filippetrovic:
Fixed via #1083: the blanket NotImplementedError fail-fast was removed; occurred-bounds temporal_filter now narrows on the embedded stack via the source_timestamp→occurred_at ingest path. test_vc_temporal_filter was rewritten with presence-and-absence assertions plus a floor-attribution control (test_vc_temporal_filter_excludes_only_old_doc) and passes un-xfailed on main. The residual genuine no-op (point-in-time target_date entity-version queries, no version columns on embedded) is tracked separately and now records a structured degradation instead of raising.

Issue #1081, opened 2026-06-10
Wire a dedicated slow/e2e live-DB CI lane for the filter rowset tests (PG+Neo4j)


filippetrovic:
## Gap

The end-to-end filter row-set acceptance tests that need a live database stack are currently not executed by any CI job:

- `tests/e2e/test_filter_rowset_graph.py` — requires a live **Postgres + Neo4j** stack (gated on `NEO4J_INTEGRATION_TEST` + Postgres reachability).
- `tests/e2e/test_filter_rowset_chronicle.py` — requires a live **Postgres** stack.

Both carry `@pytest.mark.slow`. The `test-unit` CI job includes `tests/e2e/` in its path set but selects `-m "not slow and not filter_conformance"`, which **excludes** these `slow` lanes. No other job runs `tests/e2e/` with `-m slow` against a provisioned PG+Neo4j stack, so these acceptance tests never execute in CI — they pass locally only when a developer brings up `make dev` and sets the env flags.

## Proposed work

Add a dedicated CI job (a "slow/e2e live-DB" lane) that:
- provisions Postgres + Neo4j (mirroring `test-integration`'s services block),
- sets `NEO4J_INTEGRATION_TEST=1` and `KHORA_PG_REQUIRED=1`,
- runs `uv run pytest tests/e2e/ -m slow`.

## Tracking

Until this job lands, both lanes are recorded in the verification-coverage gate's `_KNOWN_UNRUN` allowlist (`tests/unit/filter/test_verification_coverage_gate.py`) so the gap is visible and tracked rather than silently green. The allowlist entries (and this issue reference) should be removed once a CI job selects `-m slow` with PG+Neo4j.

filippetrovic:
Landed via #1083: .github/workflows/e2e.yml runs the slow e2e suite across seven engine lanes (incl. PG+Neo4j and Chronicle-on-PG) with skip→fail tripwires, gating PRs. The verification-coverage gate's _KNOWN_UNRUN entries for these two lanes were removed in #1091's rebase — allowlist is now empty.

PR #1091, opened 2026-06-10
Add verification-coverage meta-gate for the filter/recall test suite

filippetrovic:
## Summary
- Add a hermetic meta-test (`tests/unit/filter/test_verification_coverage_gate.py`) that parses the CI workflow YAMLs and the pytest config, statically enumerates the filter/recall test modules, and asserts each test is *effectively selected* (path intersected with the marker expression) by at least one CI job that provisions its required backend(s) — or carries a resolvable tracking reference. This makes a "silently disabled" filter test (green in CI but never actually executed, e.g. a `skipif(not _pg_reachable())` test running only in a job with no Postgres) impossible to merge unnoticed.
- Mirrors the two existing sibling meta-gates: a falsifiability self-test, an explicit allowlist with documented reasons, a marker-authority cross-check (any marked filter test must live inside the enumerated universe), and an empty-glob tripwire so the gate cannot pass vacuously.
- Wire the conformance seed-map round-trip test into the filter-conformance job — it carried the conformance marker but was selected by no job, so it never ran. A genuine silent-disablement the gate caught on its first run.
- Replace stale/missing `xfail` tracking references in the embedded VectorCypher and Skeleton matrix tests with resolvable issue references.

## Test plan
- [x] New gate passes and is hermetic — parses YAML + AST only, no DB/network/engine imports (proven with sockets blocked + zero project modules imported); 10/10 sub-tests green
- [x] Gate has real teeth — falsifiability self-test, marker-replacement semantics, empty-glob tripwire, and marker-authority cross-check all verified to fail on injected violations, then restored
- [x] Unit + recall suite green (`pytest tests/unit/filter/ tests/recall/ -m "not slow"`: 1730 passed)
- [x] `make format` clean; `make lint` exit 0
- [ ] Filter-conformance CI matrix passes (runs on this PR)

### DIFF
--- a/.github/workflows/filter-conformance.yml
+++ b/.github/workflows/filter-conformance.yml
@@ -311,6 +311,7 @@ jobs:
           uv run pytest \
             tests/integration/matrix/test_filter_conformance.py \
             tests/integration/matrix/test_compiler_registry_keys.py \
+            tests/integration/matrix/test_seed_map_roundtrip.py \
             -m "filter_conformance" -n auto --no-cov -v
 
       - name: Stop Postgres
--- a/tests/integration/matrix/test_skeleton_sqlite_lance.py
+++ b/tests/integration/matrix/test_skeleton_sqlite_lance.py
@@ -290,9 +290,8 @@ async def test_skeleton_recall_top_k_ordering(kb: Khora, namespace_id: UUID) -> 
         "``engine.recall`` while ``kb.remember`` resolves it to the "
         "row-level ``id`` before persisting. ``khora_chunks.namespace_id`` "
         "therefore stores the row-level id and the recall filter misses. "
-        "Same issue lurks behind PG's xfail; needs a separate "
-        "ticket to either resolve at the engine boundary or update the "
-        "test contract."
+        "Same issue lurks behind PG's xfail. Tracked in #1088 — resolve at the "
+        "engine boundary or update the test contract."
     ),
 )
 async def test_skeleton_recall_with_metadata_filter(kb: Khora, namespace_id: UUID) -> None:
--- a/tests/integration/matrix/test_vectorcypher_sqlite_lance.py
+++ b/tests/integration/matrix/test_vectorcypher_sqlite_lance.py
@@ -359,11 +359,12 @@ async def test_vc_second_ingest_sharing_entity_does_not_crash(kb: Khora, namespa
 @pytest.mark.xfail(
     strict=False,
     reason=(
-        "Khora.recall() removed the ``graph_depth`` kwarg at some point and the test "
-        "still passes it - separate issue from #806. Keeping the test under xfail "
-        "while we figure out the new traversal-depth surface; the FK-on-second-ingest "
-        "shape the previous xfail tracked is fixed in #806 and covered by "
-        "test_vc_second_ingest_sharing_entity_does_not_crash above."
+        "VectorCypher embedded multi-hop CTE traversal does not reliably cross the "
+        "second hop, so a 2-hop-reachable entity is not surfaced (and Khora.recall() "
+        "no longer accepts the ``graph_depth`` kwarg this test passes). Tracked in "
+        "#1086. NOTE: the FK-on-second-ingest crash the previous citation referenced "
+        "is a separate, fixed problem (closed #806; covered by "
+        "test_vc_second_ingest_sharing_entity_does_not_crash above)."
     ),
     raises=TypeError,
 )
@@ -636,10 +637,10 @@ async def test_vc_cooccurrence_carries_source_document_ids(kb: Khora, namespace_
 @pytest.mark.xfail(
     strict=False,
     reason=(
-        "Same root cause as test_vc_two_hop_traversal: VC's sync remember "
-        "does not rebind extraction-time entity IDs after upsert resolution, "
-        "so a second doc re-mentioning a prior entity hits FOREIGN KEY on "
-        "create_relationships_batch."
+        "VectorCypher embedded prefer_current does not honor expired edges via the "
+        "SQLite CTE traversal: an edge whose valid_to is in the past is still walked, "
+        "so content reachable only through it leaks into the recall result. Tracked in "
+        "#1087; this is the acceptance test for that bi-temporal-invalidation work."
     ),
     raises=Exception,
 )
--- a/tests/unit/filter/test_verification_coverage_gate.py
+++ b/tests/unit/filter/test_verification_coverage_gate.py
@@ -172,25 +172,17 @@ _TRACKING_REF = re.compile(r"#\d{3,}\b|ADR-\d+\b")
 # Consciously-unrun tests: known gaps awaiting a job that provisions them.
 #
 # These are KNOWN-TRACKED gaps, not silent ones — that is the whole point of the
-# gate. The live ``slow`` e2e rowset lanes need a dedicated slow/e2e live-DB CI
-# job (PG+Neo4j), which has not landed yet; until it does, no CI leg both selects
-# ``-m slow`` AND provisions those services. Each entry MUST carry a tracking
-# ref. REMOVE an entry — and clear its tracking issue — the moment a real job
-# selects ``-m slow`` with PG+Neo4j; the gate flags a stale entry that a job now
-# covers (drift in both directions).
+# gate. An entry belongs here only while a test is *consciously* not yet run by
+# any provisioning job, and always with a tracking ref. REMOVE an entry — and
+# clear its tracking issue — the moment a real job covers it; the gate flags a
+# stale entry that a job now covers (drift in both directions), so this dict
+# cannot quietly accumulate resolved gaps.
+#
+# Currently empty: the slow/e2e live-DB rowset lanes (graph / chronicle) that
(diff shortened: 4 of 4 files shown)


filippetrovic:
## Multi-perspective review — verification-coverage meta-gate

Reviewed by an agent team across 6 perspectives (security, performance, correctness, test-quality, docs, devil's advocate) plus a dedicated security audit.

**Overall assessment: ✅ APPROVE (merge-ready).** Correctness verified no silent-disablement escape hatches; security, test-quality, and docs are clean; performance is fine with optional follow-ups. The Devil's Advocate raised valuable *maintainability/longevity* concerns (captured below as non-blocking follow-ups) — none are defects in the current implementation, which is provably correct and has teeth on the live suite.

### CRITICAL / HIGH
None.

### MEDIUM (non-blocking; recommended follow-ups)
- **M1 — [Performance]** The 9 sub-tests each call `_all_modules()` / `_all_legs()`, re-parsing the YAMLs and AST-walking modules per test. Measured ~1.4s today, but module-scoped fixtures memoizing these would future-proof it as the suite grows. *Follow-up, not blocking.*
- **M2 — [Maintainability / Devil's Advocate]** The gate re-derives pytest's `-m` selection and CI job/provisioning semantics by parsing the workflow YAMLs + AST. This is correct **today** (correctness review verified all paths against the live workflows; the gate even caught a real silently-disabled test), but the parsers can drift if CI structure changes (e.g. a job rewires its pytest invocation shape). Mitigated by the zero-leg assertion and the `>=20` enumeration tripwire, which prevent a *vacuous* pass. Suggested hardening: a small smoke-assert that each known job yields ≥1 parsed pytest invocation, and revisit the parser if the CI layout changes materially.
- **M3 — [Maintainability / Devil's Advocate]** The `_KNOWN_UNRUN` allowlist depends partly on review discipline — a deferred entry with a weak issue ref is accepted as long as the ref matches `#NNNN`. Mitigated by curated-dir enumeration + the marker-authority cross-check (a marked filter test cannot escape the universe) + the staleness check (entries are flagged once a job covers them). This matches the repo's existing allowlist-gate pattern. Future hardening: require every filter test to carry a marker so nothing can be deferred marker-less.
- **M4 — [Maintainability / Devil's Advocate]** AST inference does not model `pytest.skip(..., allow_module_level=True)`, markers assigned inside `if` blocks, or dynamically-constructed skip reasons. None of these patterns exist in the current filter/recall suite, so there is no present gap — worth documenting as known limitations to expand coverage if such patterns appear.

### LOW
- **L1 — [Performance]** `_analyze_module()` walks each module's AST twice (markers/deferrals, then backend signals); combining into one walk is a ~50% AST-cost win. Optional.
- **L2 — [Devil's Advocate]** `_MIN_EXPECTED_FILTER_MODULES = 20` is a heuristic floor; it catches a full glob collapse but not a partial one that stays ≥20. Mitigated by curated dirs + marker-authority. Acceptable as a tripwire.
- **L3 — [Correctness]** `_const_str()` on f-string reasons and the `ast.unparse()` fallback return empty on exotic inputs; double-covered by `_code_signal_text()`, so no functional gap.
- **L4 — [Devil's Advocate]** Ensure the freshly-filed tracking issues (#1081, #1085–#1088) are triaged as real work items rather than reference-holders.

### Security Audit
No findings. Pure test-infra: `yaml.safe_load` (no unsafe deserialization), `ast.parse` only (no `eval`/`exec`/`compile` on file contents), bounded tracking-ref regex (no ReDoS), symlink-safe `Path.resolve()` bounded to the repo, **no dependency/manifest changes** (no supply-chain surface), and no secrets/crypto in the diff.

### Worth noting (in the PR's favor)
- The gate's purpose is to verify backend **provisioning**, not merely collection/execution — so the simpler "run-and-assert-collected" alternatives the DA proposed are complementary, not a drop-in replacement for the stated requirement.
- The falsifiability self-test reuses the real orphan-detection core (proven teeth, not vacuous), and on its first run the gate caught a genuinely silently-disabled conformance test plus several untracked xfails — exactly its intended value.

_Test plan verified during implementation: gate 10/10 and hermetic (sockets blocked, zero project imports), broad unit+recall suite 1730 passed, `make format`/`make lint` clean._

filippetrovic:
The gate's workflow universe is hardcoded to `ci.yml` + `filter-conformance.yml` (`test_verification_coverage_gate.py:81-82`). #1083 delivers the slow/e2e lanes as a new workflow file (`.github/workflows/e2e.yml`), which this gate will never parse. Consequences:

1. `test_known_unrun_entries_are_not_stale` can never fire for the two `_KNOWN_UNRUN` entries — their stated removal condition (#1081, "remove once that job lands") is a job in a file the gate doesn't read. The bidirectional-drift guarantee is dead for the one resolution path the allowlist names.
2. Once #1083 merges (either order), its five new `tests/e2e/test_filter_rowset_*.py` modules enter the universe (filenames contain "filter", `slow`-marked) with no visible covering leg, and `test_no_orphaned_filter_test` goes red. The only in-design escape is adding tests that *do* run in CI to `_KNOWN_UNRUN`, which inverts the allowlist's meaning.
3. Adding e2e.yml to the file list is not sufficient: its pytest step is `${{ matrix.select }} -m "e2e and slow"`, so `_steps_pytest_invocations` finds no `tests/`-prefixed token and drops the invocation (`:335`).

Requested changes:
- Discover workflows by glob (`.github/workflows/*.yml`) instead of a hardcoded pair.
- Substitute per-`include` `${{ matrix.<key> }}` values into `run` text before tokenizing, so `matrix.select` resolves to concrete paths/`-k` args.
- Model `-k <token>` as a substring match against the module path (the e2e lane tokens are embedded in filenames). Do not ignore `-k` — that would over-claim coverage for every module under the path.
- While extending: add `KHORA_E2E_PG_REQUIRED` / `KHORA_E2E_NEO4J_REQUIRED` / `KHORA_E2E_WEAVIATE_REQUIRED` to `_backends_from_signal`, and widen `_TRACKING_REF` from `#\d{3,4}` to `#\d{3,}`.

Non-blocking: the gate is selected only by `test-unit`; a workflow edit dropping `tests/unit/` from that job deselects the gate itself with nothing left to notice. Consider asserting that some parsed leg selects this module's own path, and/or running the gate in a second workflow. Also note `_analyze_module` unions markers module-wide — a module mixing marked and unmarked functions can be counted covered by a leg that runs only the marked subset; fine today (module-level `pytestmark` everywhere), worth a comment.

Everything else verified: falsifiability self-test is a real negative control reusing `_orphaned_modules`; marker-replacement semantics pinned; seed-map roundtrip correctly wired into the conformance invocation; #1081/#1085–#1088 all exist and match the xfail/allowlist subjects.

filippetrovic:
Thanks — addressed all the requested changes in `0d95b8d4`. The core point is right: pinning the workflow universe to two files would have silently excluded a new slow/e2e workflow and broken the drift guarantee. Changes:

1. **Glob discovery** — `_workflow_files()` now globs `.github/workflows/*.yml` (+`.yaml`); the hard-coded pair is gone. `ci.yml` remains a sanity anchor (must exist). New `test_workflow_discovery_is_glob_based` pins it.
2. **Matrix substitution** — `_substitute_matrix()` resolves `${{ matrix.<key> }}` against each `include` entry before tokenizing, and `_expand_jobs` now computes invocations *per include entry*. Job detection switched to "has a pytest run step" so a job whose path is matrix-only (e.g. `${{ matrix.select }}`) is no longer dropped before substitution.
3. **`-k` modeling** — invocations now carry a `k_expr`; `_k_selects()` evaluates it as a substring match against the module path (boolean `-k` grammar supported, with a plain-substring fallback), so a `-k` lane narrows rather than over-claims. `_eval_marker_expr` was generalized to `_eval_bool_expr(expr, leaf)` and reused for both `-m` and `-k`.
4. **E2E env signals** — `_backends_from_signal` now maps `KHORA_E2E_PG_REQUIRED` / `KHORA_E2E_NEO4J_REQUIRED` / `KHORA_E2E_WEAVIATE_REQUIRED` → postgres/neo4j/weaviate.
5. **Tracking-ref regex** — widened `#\d{3,4}` → `#\d{3,}`.

Combined effect (2)+(3)+(1): once the slow/e2e workflow lands, its `matrix.select`/`-k` legs resolve to concrete paths, provision PG+Neo4j, and *cover* the two `tests/e2e/test_filter_rowset_*.py` modules — so they leave `_KNOWN_UNRUN` (the staleness check will flag the entries) instead of going orphan. New `test_matrix_select_and_dash_k_modeling` exercises that exact shape (matrix selector + `-k` filename token + `-m "e2e and slow"`) as a forward-looking negative control.

Non-blocking items also done:
- **Gate self-coverage** — `test_gate_module_is_itself_selected_by_a_leg` asserts some leg selects this module's own path, so a workflow edit dropping `tests/unit/` from `test-unit` (deselecting the gate) fails loudly.
- **Module-wide marker union** — added the comment near `_analyze_module` noting markers are unioned module-wide (sound today because every filter/recall module uses module-level `pytestmark`).

Gate: 15/15 green, still hermetic (~1.5s). `make format`/`make lint` clean; no internal identifiers in the diff. CI re-running on `0d95b8d4`.

probablyrobot (approved):

PR #1092, opened 2026-06-11
Resolve namespace id in embedded Skeleton metadata-filter recall test

### LINKED ISSUE #1088: Skeleton embedded: namespace_id stable-vs-row-level resolution gap breaks metadata-filtered recall

filippetrovic:
## Summary
- The embedded Skeleton (SQLite + LanceDB) metadata/tag-filter recall test (`test_skeleton_recall_with_metadata_filter`) called `engine.recall` directly with the **stable** `namespace_id`, but the engine layer expects the **row-level** id — the FK target on `khora_chunks.namespace_id` — which `Khora.recall` resolves at the facade before delegating. The stable-vs-row-level mismatch made the tag filter match nothing, so the test was marked `xfail(strict=True)`.
- Resolve the stable namespace id to the row-level id in the test (via `kb._resolve_namespace(...)`) before the direct `engine.recall` call, mirroring the Postgres sibling test, and remove the `xfail`. The engine contract (caller passes the row-level id) is unchanged.
- Test-only change — no `src/` / engine behavior changes.

## Test plan
- [x] `make format` clean; `make lint` exits 0 (no new diagnostics)
- [x] Embedded Skeleton matrix `tests/integration/matrix/test_skeleton_sqlite_lance.py` — 13 passed, 0 failed (target test now passes instead of xfailing)
- [ ] CI matrix (integration + e2e + conformance legs) green

Fixes #1088

### DIFF
--- a/tests/integration/matrix/test_skeleton_sqlite_lance.py
+++ b/tests/integration/matrix/test_skeleton_sqlite_lance.py
@@ -282,26 +282,14 @@ async def test_skeleton_recall_top_k_ordering(kb: Khora, namespace_id: UUID) -> 
         assert prev >= curr, f"similarity ordering violated: {prev} < {curr} in {scores}"
 
 
-@pytest.mark.xfail(
-    strict=True,
-    reason=(
-        "Engine-level namespace resolution gap: "
-        "the test fixture passes the stable ``ns.namespace_id`` straight to "
-        "``engine.recall`` while ``kb.remember`` resolves it to the "
-        "row-level ``id`` before persisting. ``khora_chunks.namespace_id`` "
-        "therefore stores the row-level id and the recall filter misses. "
-        "Same issue lurks behind PG's xfail. Tracked in #1088 — resolve at the "
-        "engine boundary or update the test contract."
-    ),
-)
 async def test_skeleton_recall_with_metadata_filter(kb: Khora, namespace_id: UUID) -> None:
     """Tag filter restricts recall to chunks carrying the requested tag.
 
     Unlike the PG sibling, the embedded path doesn't hit the ARRAY
     incompatibility (SQLite serializes ``tags`` as JSON-text — no
-    ``ARRAY(String).contains`` incompatibility). The current xfail tracks a separate engine-layer
-    namespace-resolution gap surfaced by this test bypassing
-    :meth:`Khora.recall` and calling ``engine.recall`` directly.
+    ``ARRAY(String).contains`` incompatibility). Because this test bypasses
+    :meth:`Khora.recall` and calls ``engine.recall`` directly, it resolves the
+    stable namespace id to the row-level id itself (see below).
     """
     await _remember(
         kb,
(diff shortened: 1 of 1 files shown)


filippetrovic:
## 🔍 Multi-perspective review — consolidated

Reviewed by a 6-role team (Security, Performance, Correctness, Test Quality, Docs, Devil's Advocate) plus a security-focused audit pass.

**Overall assessment: ✅ APPROVE** — surgical, test-only change. No CRITICAL / HIGH / MEDIUM / LOW findings.

### What this changes
`tests/integration/matrix/test_skeleton_sqlite_lance.py::test_skeleton_recall_with_metadata_filter` now resolves the stable `namespace_id` to the row-level `id` (via `kb._resolve_namespace(...)`) before calling `engine.recall(...)` directly, and the `@pytest.mark.xfail(strict=True)` is removed. This mirrors the Postgres sibling (`test_skeleton_pg.py`). No `src/` / engine behavior changes.

### Findings by severity
- **CRITICAL:** none
- **HIGH:** none
- **MEDIUM:** none
- **LOW:** none

### Per-dimension notes
- **Correctness:** The engine layer expects the row-level namespace id (the FK on `khora_chunks.namespace_id`); the production `Khora.recall` facade resolves it before delegating. The test bypassed the facade and passed the stable id, so the tag filter matched nothing. Resolving in the test fixes this and matches the engine contract. `resolve_namespace` is idempotent (matches both `namespace_id` and `id`, filters to the active version, raises if absent) — verified in `storage/backends/sqlite_lance/relational.py`.
- **Test quality:** The previously-xfailed test now runs and asserts real behavior (group-A returned, group-B excluded). No assertion was weakened or skipped to pass; removing the `strict` xfail is correct (it would otherwise XPASS-error). No stale `xfail`/`#1088` references remain.
- **Performance:** No production hot-path impact — the resolution cost already exists in `Khora.recall`; the test now pays it too, purely for correctness.
- **Security:** Test-only change; no production code, secrets, isolation boundaries, or injection surfaces touched. Namespace isolation (the sole isolation key) is preserved. Security-focused audit (OWASP / dependencies / crypto): N/A for this diff — no findings.
- **Docs:** Docstring + inline comment accurately describe the namespace-id distinction; no user-facing docs or CHANGELOG entry warranted (test-contract fix, contract unchanged). Open-source hygiene clean — no internal identifiers in the diff or PR.
- **Devil's advocate:** Considered resolving at the engine boundary instead; rejected as redundant on the hot path (facade already resolves) and divergent from the established PG-sibling precedent. The "engine expects row-level id" edge exists only for callers bypassing the facade (the engine protocol is not a public API) — not a product bug.

msrdic (approved):

Issue #1087, opened 2026-06-10
VectorCypher embedded: prefer_current does not honor expired edges via CTE traversal


filippetrovic:
## Gap

On the embedded VectorCypher stack (SQLite + LanceDB, SQL-emulated graph via recursive CTEs), `prefer_current=True` does not honor bi-temporally invalidated/expired relationships during traversal. An edge whose `valid_to` is in the past should be skipped when `prefer_current=True`, making content reachable only through that expired edge unreachable. Today the CTE traversal walks expired edges anyway, so expired-edge content leaks into the recall result.

## Affected test

`tests/integration/matrix/test_vectorcypher_sqlite_lance.py::test_vc_prefer_current_via_cte` is the end-to-end acceptance test (a B->C edge is expired; with `prefer_current=True`, C must be unreachable from A). It is currently marked `@pytest.mark.xfail` and tracks this gap.

## Expected behavior

With `prefer_current=True`, the embedded CTE traversal should exclude relationships whose validity window has closed (expired `valid_to`), consistent with Graphiti-style bi-temporal invalidation on the production stack.

Issue #905, opened 2026-05-30
get_chunks_by_document() returns [] on Skeleton and VectorCypher (sqlite_lance) despite chunks_created>0 and recall finding them


damirkrstanovic:
## Summary

After `kb.remember(text)` succeeds and reports `RememberResult.chunks_created=1`,
calling `Khora.storage.get_chunks_by_document(rem.document_id, namespace_id=...)`
returns an **empty list** on engine `skeleton` and engine `vectorcypher` with
`backend=sqlite_lance`. The chunks **are** stored — a free-text
`Khora.recall()` against the same namespace returns them — but they are not
findable via the document id that `RememberResult` reports. `engine=chronicle`
on the same backend works correctly: chunks are both recallable and findable
by document id.

User-visible effect: any program that uses `rem.document_id` to walk the
chunks of a just-stored document (audit, display, post-processing) sees an
empty list on the two non-chronicle engines. The data is there; the
document-id wiring between `RememberResult` and the storage layer is
inconsistent across engines.

## Affected

- Engines: **skeleton**, **vectorcypher** (chronicle unaffected)
- Backend: **sqlite_lance**

## Environment

- Khora 0.17.4
- Python 3.13
- Storage: `sqlite_lance` (embedded)

## Reproduction

Runs the same flow on all three engines against a fresh `sqlite_lance` store:
`remember()` → `get_chunks_by_document(document_id, namespace_id=row_id)` →
free-text `recall()`. (`get_chunks_by_document` is a coordinator method and
does not auto-resolve the stable namespace id, so the **row** id from
`resolve_namespace` is passed.)

```bash
python repro.py
```

```python
"""Repro: Khora.storage.get_chunks_by_document(rem.document_id) returns [] on
engines skeleton and vectorcypher after a successful remember() (backend=sqlite_lance).

The chunks exist (a free-text recall returns them), but they are not findable
by the document id that RememberResult reports. chronicle is unaffected.

Run:  python repro.py
"""
import asyncio
import logging
import os
import tempfile

os.environ.setdefault("LITELLM_LOG", "ERROR")
os.environ.setdefault("HF_HUB_DISABLE_PROGRESS_BARS", "1")
os.environ.setdefault("TRANSFORMERS_VERBOSITY", "error")
os.environ.setdefault("TOKENIZERS_PARALLELISM", "false")
logging.getLogger("LiteLLM").setLevel(logging.ERROR)

from loguru import logger
logger.remove()

from khora import Khora
[...]

PR #1093, opened 2026-06-11
Fix embedded VectorCypher two-hop graph traversal

### LINKED ISSUE #1086: VectorCypher embedded: multi-hop CTE traversal correctness
## Gap

On the embedded VectorCypher stack (SQLite + LanceDB, SQL-emulated graph via recursive CTEs), multi-hop graph traversal during recall does not reliably surface entities reachable across more than one hop. A query anchored on entity A is expected to surface content connected via A -> B -> C (2-hop), but the CTE traversal does not cross the second hop, so the 2-hop-reachable content is missing from the recall context.

## Affected test

`tests/integration/matrix/test_vectorcypher_sqlite_lance.py::test_vc_two_hop_traversal` is the end-to-end acceptance test (3 connected docs A->B->C; querying A must surface C). It is currently marked `@pytest.mark.xfail` and tracks this gap.

Note: this test previously carried a citation to a now-closed issue about a foreign-key crash on a second ingest sharing an entity. That FK crash is a *separate, fixed* problem (covered by `test_vc_second_ingest_sharing_entity_does_not_crash`). This issue tracks the distinct, still-open multi-hop CTE traversal correctness gap.

## Expected behavior

Recall with a multi-hop graph depth on the embedded VectorCypher stack should surface entities/content reachable across the requested number of hops, consistent with the production (Neo4j) stack.

filippetrovic:
## Summary

On the embedded VectorCypher stack (SQLite + LanceDB, SQL-emulated graph via recursive CTEs), a `recall(..., graph_depth=2)` did not surface a 2-hop-reachable entity (anchor `A` → `B` → `C`, where `C` should be reachable from `A` at depth 2).

Root cause was wiring, not the recursive CTE itself:

- `Khora.recall()` did not accept a `graph_depth` keyword argument, so passing it raised `TypeError` at the public boundary.
- The VectorCypher retriever called the storage coordinator's `get_neighborhoods_batch(...)` without the required keyword-only `namespace_id`, which raised `TypeError` inside the embedded graph-expansion step and crashed the recall before the recursive CTE ever ran. This blocked *all* embedded graph expansion, not just the second hop.

Changes:

- `Khora.recall()` now accepts `graph_depth` and forwards it to the engine only when set (VectorCypher honors it; other engines ignore it, so their behavior is unchanged).
- The retriever passes `namespace_id` through to the coordinator's `get_neighborhoods_batch`.
- The recursive CTE that walks edges was already correct (proper depth increment, next-hop joins, edge-visited cycle guard, depth clamp) — it is unchanged.
- Removes the now-passing acceptance test's `xfail` and refreshes the module docstring.

Closes the multi-hop traversal gap so the embedded path matches the production graph stack's variable-length-path semantics.

## Test plan

- [x] `test_vc_two_hop_traversal` passes non-vacuously (the 2-hop concept surfaces) with the `xfail` removed
- [x] Full `tests/integration/matrix/test_vectorcypher_sqlite_lance.py`: `13 passed, 1 xfailed` (the unrelated `prefer_current` gap is untouched and still xfails)
- [x] Broader unit sweep over the VectorCypher retriever, embedded graph backend, coordinator, and recall facade: green
- [ ] CI integration matrix (incl. embedded VectorCypher lane) green

Fixes #1086


filippetrovic:
## Consolidated Review — PR #1093

**Overall assessment: ✅ APPROVE — no blocking findings.** The fix is correct, minimal, and scoped to the second-hop traversal gap (#1086). Reviewed by a 7-perspective team (Security, Performance, Correctness, Test, Database/Graph, Docs, Devil's Advocate) plus a dedicated security audit. **No CRITICAL or HIGH findings.** Every MEDIUM below is either pre-existing code or an architectural follow-up — none is a defect introduced by this PR.

### Security Audit — PASS
- 20 added lines; no injection / secrets / auth / file-path / SSRF / deserialization patterns; no dependency or manifest changes.
- **INFO:** the added `namespace_id=namespace_id` *strengthens* multi-tenant isolation in the embedded graph-expansion path (it was the missing keyword-only arg that raised `TypeError` and crashed recall before the CTE ran). The recursive CTE binds `namespace_id` as a SQL parameter on every UNION arm — no injection vector.

### MEDIUM — non-blocking (pre-existing or follow-up)
- **M1 · [Correctness / Devil's Advocate]** `retriever.py` (`base_depth = graph_depth or routing.graph_depth`) — falsy-zero gotcha: a public `graph_depth=0` silently falls through to the routed depth instead of meaning "no expansion". The line is **pre-existing**, only now reachable via the new public kwarg. Not a defect introduced here. Follow-up: `graph_depth if graph_depth is not None else routing.graph_depth`.
- **M2 · [Database/Graph]** embedded SQLite clamps traversal depth to 6 hops while the Neo4j path clamps to 4 — a behavioral asymmetry for depth 5–6. **Pre-existing**; not exercised by #1086 (routers request depth ≤ 2). Follow-up: unify the ceiling or document the divergence.
- **M3 · [Devil's Advocate]** `graph_depth` is forwarded through a conditional kwargs dict in `Khora.recall()` rather than declared on `MemoryEngineProtocol.recall()` (the way `filter_ast` is). It works correctly and is documented inline, but a protocol-level declaration would make the contract explicit. Follow-up: add `graph_depth` to the protocol and forward unconditionally.
- **M4 · [Test]** only `depth=2` is exercised; `depth>2` and cycle handling are untested. #1086's acceptance criterion is `depth=2`; a follow-up parametrized `depth ∈ {1,2,3}` test would harden the N-hop path.

### LOW / INFO — confirmations
- **[Performance]** Traversal is well-bounded: dual depth clamp (≤ 4 on VectorCypher), `limit_per_entity=20`, edge-visited cycle guard, and 3 batched queries per call (no N+1). `visited` string grows to ~150 B at depth 4 (`instr()` O(n)) — negligible. No dense-graph stress test, acceptable for a feature gate.
- **[Test]** `test_vc_two_hop_traversal` is **non-vacuous**: the test's embeddings are hash-based (not semantic), so the 2-hop concept can surface *only* via true A→B→C graph expansion, never via vector similarity. Safe to remove the xfail. The sibling `test_vc_prefer_current_via_cte` (#1087) is correctly left untouched and still xfails.
- **[Docs]** New `graph_depth` docstring, CHANGELOG entry, and test-module docstring are accurate and consistent. No leaked internal identifiers — only public GitHub issue refs.
- **[Correctness]** Conditional forwarding verified correct for non-VectorCypher engines (skeleton/chronicle don't accept the kwarg and never receive it). `namespace_id` is in scope and correctly typed. No placeholders.
- **[Database/Graph]** Recursive-CTE depth semantics are correct (`[*1..2]` parity with the Neo4j variable-length path), namespace isolation is enforced on every CTE arm, and the edge-uniqueness cycle guard matches Neo4j `[*1..N]` semantics.

---
*Reviewers: security · performance · correctness · test · database/graph · docs · devil's-advocate, plus a dedicated security audit (verdict: PASS). All MEDIUM items are tracked as follow-ups; none blocks this PR.*

filippetrovic:
The two wiring fixes are correct and needed: `Khora.recall()` lacked the `graph_depth` kwarg, and the `get_neighborhoods_batch` call was missing the required keyword-only `namespace_id` (only Neo4j-transient errors are caught around `_cypher_expand`, so the TypeError crashed any recall reaching graph expansion). The recursive CTE itself checks out — called directly with this test's A→B→C data it returns depth=1 → {bob}, depth=2 → {bob, carol}, with namespace predicates on every arm and edge-level cycle protection.

But `test_vc_two_hop_traversal` does not exercise the fix, and the end-to-end gap in #1086 is still reproducible:

1. **The test is vacuous.** The query `"Alice"` routes as `QueryComplexity.SIMPLE`, so retrieval takes `_simple_retrieve`; `graph_depth=2` is ignored and `_cypher_expand` never runs. The assertion passes because the vector channel returns all 3 docs (corpus of 3, `limit=10`). With the `retriever.py` hunk of this PR reverted, the test still passes — the only change it actually gates is the facade kwarg removing the TypeError the old xfail was pinned to.

2. **With the graph channel forced** (`mode=SearchMode.GRAPH`, query `"alice"` so the hash-embedder's entity search returns the seed), traversal returns bob/carol but recall yields zero chunks:
   - the normalization in `_cypher_expand` keeps only `isinstance(entity_data, dict)` entries, while the sqlite_lance backend returns `Entity` objects — every expanded entity is silently dropped, with no degradation recorded (see `docs/architecture/failure-observability-contract.md`);
   - `SQLiteLanceVectorAdapter.get_chunks_batch` queries only `chunks`, but this engine writes to `khora_chunks` (#905) — `get_chunks_by_document` has the fallback, `get_chunks_batch` does not — so `_fetch_chunks_from_entities` gets 0 rows.

Requested changes:

1. Normalize `Entity` objects in `_cypher_expand` (map to `{"id": ..., "name": ..., "entity_type": ...}` before the dict check; record a degradation for anything still unrecognized).
2. Add the `khora_chunks` fallback to `get_chunks_batch`, mirroring `get_chunks_by_document`.
3. Make the test non-vacuous: `mode=SearchMode.GRAPH` + query `"alice"` + `graph_depth=2` so chunks can only arrive via graph expansion; keep the `"graph databases"` assertion and additionally assert the 2-hop chunk is absent at `graph_depth=1`, so hop-semantics regressions fail loudly.
4. Alternatively, re-scope: land only the wiring fixes, restore the xfail with a corrected reason (remaining gap = entity-shape normalization + chunk-table lookup, not the CTE), and drop `Fixes #1086` so the issue stays open.
5. `Khora.recall()` docstring: chronicle/skeleton `recall()` don't accept `graph_depth` — passing it there raises TypeError, so "other engines ignore it" is inaccurate; also note the value is a base for adaptive depth and has no effect on SIMPLE-routed queries.

Note: this PR and #1096 both rewrite the module docstring / xfail markers in `test_vectorcypher_sqlite_lance.py`; whichever merges second needs reconciliation.

filippetrovic:
Closing without merge. Decision: `graph_depth` will not be added to the public `Khora.recall()` surface — traversal depth stays engine-internal (adaptive routing already selects depth ≥ 2), and the acceptance test as written cannot validate the traversal path (see review above). The real remaining gaps behind #1086 — `Entity`-object normalization in `_cypher_expand`, the `khora_chunks` fallback in `get_chunks_batch`, and the `namespace_id` wiring fix from this PR — will land together in a follow-up PR with engine-level (non-facade) test coverage. #1086 stays open until then.

PR #1094, opened 2026-06-11
Add fail-loud tripwires for container-free e2e CI lanes

filippetrovic:
## Summary
- The three container-free e2e matrix lanes (embedded sqlite_lance and in-process SurrealDB) skip every test and stay **green** if their optional dependencies ever stop installing, masking real failures. The container-backed lanes are already protected by skip→fail tripwires (`*_REQUIRED` env flags checked in `pytest_configure`); the container-free lanes had no equivalent.
- Add `KHORA_E2E_EMBEDDED_REQUIRED` / `KHORA_E2E_SURREAL_REQUIRED` flags, wired per-leg in the e2e workflow matrix, that make the e2e `pytest_configure` abort the session (exit 1) when the flag is set but the embedded stack (`aiosqlite` + `lancedb`) or the SurrealDB SDK is not importable — mirroring the existing Postgres/Neo4j/Weaviate tripwires.
- Teach the verification-coverage meta-gate to recognize both new flags (both map to the embedded sentinel, which every leg provides).
- Add a hermetic regression test proving the tripwires fire on dependency drift and stay quiet when the deps are present.

Behavior is unchanged when the embedded extras are installed (CI reality): the flags only convert an all-skip into a loud failure.

## Test plan
- [x] Unit + recall suite passes locally (`make test-unit`: 8631 passed, 35 skipped)
- [x] New negative-verification test asserts `pytest_configure` aborts when the import probe is `False` and returns quietly when `True`
- [x] Verification-coverage gate still green (the new test is a pure-unit module; no false orphan)
- [ ] Integration + 7-lane e2e matrix pass in CI (require Docker; exercised by the CI workflows)

### DIFF
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -71,25 +71,33 @@ jobs:
             select: tests/e2e/test_filter_rowset_graph.py tests/e2e/test_graph_contribution.py
             postgres: true
             neo4j: true
-          # Embedded vectorcypher engine — sqlite_lance, no service.
+          # Embedded vectorcypher engine — sqlite_lance, no service. `embedded`
+          # flips the conftest tripwire on so a missing aiosqlite/lancedb fails
+          # RED instead of skipping green.
           - backend: vc_embedded
             select: tests/e2e/ -k vc_embedded
+            embedded: true
           # Pgvector skeleton — live Postgres only.
           - backend: skeleton_pgvector
             select: tests/e2e/ -k skeleton_pgvector
             postgres: true
-          # SurrealDB skeleton — embedded (memory://), no service.
+          # SurrealDB skeleton — embedded (memory://), no service. `surreal`
+          # flips the conftest tripwire on so a missing surrealdb SDK fails RED
+          # instead of skipping green.
           - backend: skeleton_surrealdb
             select: tests/e2e/ -k skeleton_surrealdb
+            surreal: true
           # Weaviate skeleton — vectors in live Weaviate, documents in live
           # Postgres (so this leg provisions BOTH stores).
           - backend: skeleton_weaviate
             select: tests/e2e/ -k skeleton_weaviate
             postgres: true
             weaviate: true
-          # SQLite + LanceDB skeleton — embedded, no service.
+          # SQLite + LanceDB skeleton — embedded, no service. `embedded` flips
+          # the conftest tripwire on (same as vc_embedded).
           - backend: skeleton_sqlite_lance
             select: tests/e2e/ -k skeleton_sqlite_lance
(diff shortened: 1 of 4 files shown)


filippetrovic:

Six independent reviewers (security, performance, correctness, tests, docs, devil's advocate) plus a focused security audit. **No blocking findings.** The change is correct, surgical, and well-tested.

### What was verified
- **Correctness** — the two new `pytest_configure` tripwires mirror the existing Postgres/Neo4j/Weaviate checks exactly (env name, `== "1"`, `not <probe>()`, `returncode=1`, message style). Matrix flags sit on the correct three legs (`embedded` → vc_embedded + skeleton_sqlite_lance; `surreal` → skeleton_surrealdb). `_backends_from_signal` maps both flags to the embedded sentinel and is test-pinned. No placeholders.
- **Tests** — the new negative test has real teeth: it asserts `pytest_configure` aborts (`_pytest.outcomes.Exit`) when the import probe is `False` and returns quietly when `True`, for both embedded and SurrealDB. Probes are monkeypatched on the shared module object; sibling `*_REQUIRED` flags are cleared for isolation. Verification-coverage gate stays green (no false orphan).
- **Security** — CI/test-harness only; no production code, no dependency/lockfile changes, no secrets, no network/eval. GitHub Actions `${{ matrix.x && '1' || '' }}` expressions are injection-safe; `pytest.exit` messages are static. OWASP/dependency/crypto surface: N/A.
- **Performance** — checks run once at session start via cheap `importlib.util.find_spec` probes; zero per-test overhead.
- **Docs** — workflow comments, conftest tripwire doc-comment, and test docstrings consistently and accurately name the new flags (aiosqlite + lancedb for embedded; SurrealDB SDK for surreal). No drift.

### Non-blocking notes (LOW)
- **L1 [maintainability]** `tests/e2e/conftest.py` — the local `_embedded_available()` helper is now unused, but it was **already dead before this change** (the prior `pytest_configure` never called it; the test modules use the harness probe). Per surgical-change discipline it's left as-is and flagged as a good follow-up cleanup rather than expanded scope here.
- **L2 [CI style]** The `matrix.<flag>` + `${{ … && '1' || '' }}` indirection was considered vs. inline per-leg env. Kept intentionally: it matches the existing PG/Neo4j/Weaviate idiom, and GitHub Actions `matrix.include` entries can't carry step-level `env:` anyway, so a matrix var is the idiomatic route.
- **L3 [docs nit]** A conftest comment phrases the container-free legs as having "no store to socket-probe" — slightly loose since the embedded probes are importability checks; harmless.

**Verdict:** ready to merge once CI is green.

msrdic (approved):

PR #1095, opened 2026-06-11
Make sqlite_lance vector-channel temporal upper bound inclusive

filippetrovic:
## Summary
- The SQLite-side recency-window refinement in the `sqlite_lance` **vector** channel used a half-open `<` on the `COALESCE(source_timestamp, created_at)` upper bound, while pgvector, SurrealDB (vector + BM25), and this backend's own BM25 channel all use an inclusive `<=`.
- The half-open form **false-excluded** a chunk whose effective timestamp equals the bound instant exactly: the row was dropped pre-filter, and the Python post-filter can only remove rows, never restore them — a cross-backend determinism break and an intra-backend split with the BM25 channel.
- Standardized on inclusive `<=` so the pushed candidate set stays a safe **superset** of the post-filtered set (the bound instant stays in the window; `$lt` vs `$lte` strictness is re-checked by the Python post-filter, not carried into the pushed bound).
- Replaced the misleading comment that claimed a non-existent "half-open contract" with the accurate inclusive / superset-safe pushdown rationale.
- Added a vector-channel regression test that seeds a chunk on the exact upper-bound instant and asserts an inclusive `$lte` filter returns it (stash-verified: fails with `<`, passes with `<=`).

All four chunk-search temporal pushdown sites now use `>=` lower / `<=` upper; lower bounds were left untouched.

## Test plan
- [x] New regression test `test_source_timestamp_upper_bound_is_inclusive_at_boundary` passes via the vector channel (semantic-only routing) and fails when the one-character fix is reverted
- [x] Embedded / hermetic suites green locally (`chronicle_filter_embedded`: 6 passed; `sqlite_lance` vector unit: 157 passed; recall-filter suite: 453 passed)
- [x] `ruff format` / `ruff check` / type check clean
- [ ] Full integration suite + combined coverage gate (Postgres/Neo4j) — runs in CI

### DIFF
--- a/src/khora/storage/backends/sqlite_lance/vector.py
+++ b/src/khora/storage/backends/sqlite_lance/vector.py
@@ -481,9 +481,14 @@ class SQLiteLanceVectorAdapter:
         # truth for chunk metadata and tracks ``source_timestamp`` (LanceDB
         # only stores ``created_at``), so we re-apply the bounds here using
         # ``COALESCE(source_timestamp, created_at)`` — matches the pgvector
-        # backend's column-precedence rule (PR #470). Half-open
-        # interval ``>= start AND < end`` to match the Chronicle pushdown
-        # contract.
+        # backend's column-precedence rule (PR #470). The upper bound is
+        # INCLUSIVE (``<=``) to honor the superset-safe pushdown contract: the
+        # pushed candidate set must be a superset of the post-filtered set and
+        # never false-exclude, so the bound instant itself stays inside the
+        # window. Strictness (``$lt`` vs ``$lte``) is deliberately NOT carried
+        # into the pushed bound — ``compile_python`` re-checks exact strictness
+        # post-filter. This also matches pgvector, SurrealDB, and this
+        # backend's own BM25 channel.
         # Defense-in-depth (IDOR family): also enforce namespace_id on the SQLite
         # side rather than trusting LanceDB's filter alone. If LanceDB's
         # where-clause regressed, the SQLite filter would still keep cross-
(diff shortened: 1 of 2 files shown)


filippetrovic:

Reviewed across 7 domains (security, performance, correctness, test quality, database, docs, devil's advocate) plus a security audit pass.

**Overall assessment: ✅ APPROVE** — no blocking findings. The `<` → `<=` change correctly makes the sqlite_lance vector-channel temporal upper bound inclusive, in line with the superset-safe pushdown contract (`ChronicleDateBound`: the pushed window is an over-approximation; exact `$lt`/`$lte` strictness is re-checked by the `compile_python` post-filter) and matching pgvector, SurrealDB, and this backend's own BM25 channel. Comment rewrite is accurate; the regression test exercises the vector channel at the exact boundary instant and fails without the fix (stash-verified).

### CRITICAL
- None.

### HIGH
- None.

### MEDIUM
- **[Test] M1** `tests/integration/test_chronicle_filter_embedded.py` — Boundary coverage is upper-bound-only. The lower bound (`>=`, unchanged) has no boundary-focused test. A paired lower-bound boundary test would strengthen parity coverage. Non-functional (consistency/coverage), not blocking this fix.

### LOW
- **[Database/Correctness] L1** `src/khora/storage/backends/sqlite_lance/vector.py:466` — The LanceDB coarse pre-filter still uses strict `<` on `created_at`-only (not `COALESCE(source_timestamp, created_at)`). Latent edge: a row with `source_timestamp IS NULL` and `created_at` exactly equal to the bound could be false-excluded at the LanceDB layer, violating the superset-safe contract. **Pre-existing, not introduced by this PR**, and out of scope for this one-character fix — recommend a separate follow-up to either align the coarse pre-filter to inclusive/COALESCE semantics or add a NULL-edge regression test.
- **[Test] L2** The new test does not directly assert that the LanceDB coarse pre-filter returned both seeded rows (it isolates the SQLite refinement by design). Minor observability gap, not functional.
- **[Test] L3** No exact-boundary case with `source_timestamp IS NULL` (COALESCE-fallback path). Low priority since chunks typically carry `source_timestamp`; ties into L1.

### Security Audit
- No findings. The change is a hardcoded SQL operator on a fully parameterized predicate (bound `?` placeholder) — no injection surface change, no data exposure, no auth/authz impact; the namespace_id defense-in-depth check is preserved. No new dependencies. No secrets. (The repo's only outstanding dependency advisory is unrelated to this diff and already allowlisted upstream.)

### Notes
- Scope is correct and minimal (one-char operator + comment correction + one regression test); no over-engineering.
- All temporal pushdown sites now use `>=` lower / `<=` upper (verified by grep across pgvector, SurrealDB, and both sqlite_lance channels).

msrdic (approved):

PR #1097, opened 2026-06-11
Fail loud on recall filters in the turbopuffer backend

filippetrovic:
## Summary
- The turbopuffer skeleton backend accepted a `filter_ast` argument for protocol parity but **silently ignored it** on both the vector ANN path and the hybrid BM25 multi-query channel. Only the temporal filter reached the server, so a skeleton-on-turbopuffer deployment dropped every deterministic recall filter with no signal — `recall(query, filter={...})` returned unfiltered results.
- `TurbopufferTemporalStore.search` now raises `RecallFilterUnsupportedError` at the top of the method when `filter_ast` is non-None, before any I/O, so the guard covers every search channel. The `filter_ast=None` path is byte-for-byte unchanged.
- This backend does not implement deterministic recall filters; failing loud upholds the never-return-violating-rows contract rather than leaking filter-excluded rows. Native server-side filter pushdown is intentionally left as a possible future enhancement.
- Updated the adapter docstring and the operator docs (Turbopuffer section) to state that recall filters are unsupported and raise.

## Test plan
- [x] Adapter unit tests: `filter_ast=None` keeps existing behavior; non-None raises `RecallFilterUnsupportedError`
- [x] Engine-level test proves the error surfaces out of `recall()` un-swallowed (no fallback catches it)
- [x] AST audit-gate test pins the fail-loud contract (fails if the guard is deleted or weakened)
- [x] `ruff format` + `ruff check` clean; targeted suite green (62 passed)
- [ ] CI full suite passes

### DIFF
--- a/docs/engines/skeleton-engine.md
+++ b/docs/engines/skeleton-engine.md
@@ -378,6 +378,14 @@ to turbopuffer's filter DSL (`Eq`, `Gte`, `Lte`, `Lt`, `In`, `Contains`,
 `ContainsAny`). The one workaround: ALL-tags semantics fold into an
 `And` of N `Contains` clauses (turbopuffer has no native `ContainsAll`).
 
+**Recall filters are unsupported.** The deterministic recall-filter API
+(the `filter_ast` argument to `search`) is **not** implemented on
+turbopuffer. Passing a non-None filter RAISES
+`RecallFilterUnsupportedError` rather than silently returning unfiltered
+results - the backend fails loud instead of dropping the filter on the
+floor. Native server-side filter pushdown is a possible future
+enhancement.
+
 **Features:**
 
 - Multi-region serverless (no infrastructure to run)
@@ -402,7 +410,7 @@ turbopuffer wins. Numbers as of 2026-05-24 - re-check
 
 **Install:** `pip install khora[turbopuffer]`.
 
-**Tests.** 42 unit tests exercise the SDK via a fake `turbopuffer`
+**Tests.** A suite of unit tests exercises the SDK via a fake `turbopuffer`
 module injected into `sys.modules` (no real network or API key needed).
 Real-cluster integration tests are gated behind
 `TURBOPUFFER_INTEGRATION_TEST=1` + a `TURBOPUFFER_API_KEY` env var; not
--- a/src/khora/engines/skeleton/backends/turbopuffer.py
+++ b/src/khora/engines/skeleton/backends/turbopuffer.py
@@ -340,11 +340,16 @@ class TurbopufferTemporalStore(TemporalVectorStore):
             the operator docs so users picking turbopuffer know
             ``hybrid_alpha`` is a no-op on this backend.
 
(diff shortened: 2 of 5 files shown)


filippetrovic:
## Multi-perspective review

Six reviewers (correctness, security, performance, tests, docs, devil's advocate) plus a dedicated security audit. The change is correct, surgical, and well-tested for the constraint-bearing path — but one edge case raises spuriously and should be guarded.

**Overall: REQUEST CHANGES** (one HIGH finding; fix is one line + a test and is being applied in a follow-up commit on this branch).

### HIGH

**H1 — An explicitly-empty filter raises spuriously (correctness / devil's-advocate).**
An empty, constraint-free filter (`filter={}` or `RecallFilter()`) normalizes to a **non-None** empty-`AND` node, not `None`. Verified empirically: `parse_to_ast(RecallFilter())` returns `FilterNode(op=AND, children=())`. The new guard `if filter_ast is not None:` therefore raises `RecallFilterUnsupportedError` on a no-op "match-everything" filter that imposes no constraints and should simply pass through. For contrast, the pgvector path compiles an empty filter to a no-op, and the engine's own pushdown reporting already uses `filter_ast is not None and bool(filter_ast.children)` as the "has real constraints" signal.

**Fix:** guard on `if filter_ast is not None and filter_ast.children:` so only filters carrying actual predicates raise; add a test asserting an empty filter passes through unchanged.

### No findings (approved dimensions)

- **Correctness** — guard sits before all I/O and before the hybrid branch (covers both the vector-only and hybrid/RRF channels); exact exception type with a runtime import; `filter_ast=None` byte-for-byte unchanged; no-swallow propagation verified through `recall()` (no wrapping `try/except`) up to the facade's bare re-raise; only `search` accepts `filter_ast`. ✓ (modulo H1)
- **Security** — no new input/auth/crypto/secret surface; failing loud replaces a *silent* return of filter-excluded rows (a data-exposure improvement); no secret-typed model changes.
- **Performance** — O(1) early check before any network I/O; common `None` path unchanged.
- **Tests** — both branches covered; the engine-level test proves the raise surfaces un-swallowed using a real store; the AST audit-gate pin fails if the guard is deleted or weakened. (Extend with an empty-filter pass-through case per H1.)
- **Docs** — docstring and operator note are accurate; the brittle unit-test count was reworded rather than replaced with another number.
- **Scope** — 5 files, no scope creep.

### Security Audit

No findings. The diff adds no endpoints, input parsing, deserialization, auth, crypto, secrets, or dependencies — it is a parameter check that raises a domain exception, so the OWASP Top 10 categories do not apply. The fail-loud behavior strengthens the data-exposure posture relative to the prior silent-ignore.

filippetrovic:
Description nit: bullet 2 and the test plan still say the guard raises on any non-None `filter_ast`; after the second commit it raises only on constraint-bearing filters (`filter_ast.children` non-empty) and the empty match-everything filter passes through (matching how the facade normalizes `filter={}` and how sqlite_lance treats empty-children as no-constraint). Update the body so the public record matches the merged behavior.

msrdic (approved):

PR #1098, opened 2026-06-11
Enforce recall filter in SurrealDB BM25 full-text search

filippetrovic:
## Summary
- The SurrealDB temporal store's `search_fulltext` (the BM25 keyword channel) accepted a deterministic recall-filter AST but ignored it — it built its `WHERE` from namespace + temporal bounds only. As a result, the BM25 arm could return filter-excluded chunks into RRF fusion, so a filtered recall leaked rows the caller asked to exclude.
- Compile the filter AST to a SurrealQL `WHERE` predicate and AND it into the BM25 query, mirroring the vector channel's existing pushdown exactly (same compiler, `metadata` → `metadata_` mapping, raise-on-unsupported). When no filter is supplied, behavior is byte-identical to before.
- Update the method docstring, which previously stated the filter was ignored.

## Tests
- New embedded-SurrealDB (in-process `memory://`, no Docker) enforcement test that drives the public BM25 lookup directly with a three-way proof: an unfiltered control returns the full corpus, a *matching* metadata filter retains every row, and a *contradiction* metadata filter excludes every row. Stash-verified to fail without the source change (rows leak) and pass with it.
- New source-pinned audit-gate guard asserting `search_fulltext` declares the filter parameter, compiles it, and threads the compiled predicate/bindings into the BM25 `WHERE` — so a future regression that drops the thread-through fails the unit suite.

## Test plan
- [x] Unit tests pass (filter, recall, surrealdb backend, audit gate)
- [x] New embedded-SurrealDB BM25 enforcement test passes (stash-verified teeth)
- [ ] CI: full matrix + SurrealDB conformance leg

### DIFF
--- a/src/khora/engines/skeleton/backends/surrealdb.py
+++ b/src/khora/engines/skeleton/backends/surrealdb.py
@@ -503,8 +503,9 @@ class SurrealDBTemporalStore(TemporalVectorStore):
         ``Chunk`` adaptation. Falls back to ``[]`` on backends without
         a BM25 index configured (``optimize_storage()`` not yet run).
 
-        ``filter_ast`` is accepted for protocol parity; this backend does not
-        compile the recall-filter AST yet, so it is ignored.
+        ``filter_ast`` is the deterministic recall-filter AST. When provided it
+        is compiled to a SurrealQL ``WHERE`` predicate and AND-ed into the BM25
+        query alongside the namespace + temporal-bound clauses.
         """
         if not query_text or not query_text.strip():
             return []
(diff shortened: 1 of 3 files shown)


filippetrovic:
## 🤖 Multi-Perspective Review — Consolidated

**Overall assessment: ✅ APPROVE** — no blocking findings. Seven independent reviewers (security, performance, correctness, database, docs, test, devil's advocate) plus a dedicated security audit. The change is minimal, mirrors the existing vector-channel pushdown exactly, and is well-tested (three-way value-discrimination proof + source-pinned audit guard, stash-verified to fail without the fix).

### CRITICAL
_None._

### HIGH
_None._

### MEDIUM
- **[Maintainability]** `src/khora/engines/skeleton/backends/surrealdb.py` — the ~12-line `compile_surrealdb` + `filter_clauses.append` / `filter_bindings.update` block in `search_fulltext` duplicates the identical block in `_search_inner` (the vector path). A shared `_apply_filter_ast(...)` helper would remove the drift risk. **Deferred, non-blocking:** the duplication is the intentional, mandated "mirror the vector channel" shape, and the new audit-gate guard pins the *inline* presence of those calls inside `search_fulltext` — so extracting a helper is a deliberate larger refactor (it must also rewrite the guard) better suited to a follow-up than this surgical fix.

### LOW
- **[Test robustness]** `tests/integration/matrix/test_filter_enforcement_surrealdb.py` — the BM25 index is defined via inline raw DDL (because `ensure_search_indexes()` bundles a 1536-dim HNSW the test's small embedding dim can't satisfy). Could reference the canonical index constant to prevent silent drift. The test reviewer independently rated this **not fragile** (`IF NOT EXISTS` idempotent, analyzer is part of schema init, embedded-scoped). Optional.
- **[Coverage]** No end-to-end `recall(mode=KEYWORD)` test. Note: on this backend `recall(KEYWORD)` routes through `_search_inner` (which already enforced the filter), so an end-to-end test would be teeth-less for *this* fix — the direct `search_fulltext` test is the correct surface. An end-to-end test would only guard the retriever's routing; optional.
- **[Test style]** Direct backend access (`kb._engine._get_temporal_store()`) to reach the store under test — necessary to isolate the enforcement surface; acceptable, fails loudly if internals change.

## Security Audit
**Verdict: PASS.** The compiled predicate binds all user values as named parameters (no string interpolation into the query); the metadata path-segment guard rejects unsafe identifiers; BM25 query text and the contradiction values are parameterized. No new entry points, secrets, dependencies, deserialization, file, or SSRF surface. The only `execute()` in the diff is a static DDL string in test setup.

---
_Reviewers: Security ✅ · Performance ✅ · Correctness ✅ · Database ✅ · Docs ✅ (docstring accurate, no drift) · Test ✅ (teeth verified) · Devil's Advocate (maintainability suggestions above)._

msrdic (approved):

PR #1103, opened 2026-06-11
Make examples-smoke assert results and cut per-example runtime

### LINKED ISSUE #1099: examples-smoke: retry storms from non-JSON mock LLM, cross-encoder HF download eats the 60s budget, examples lack assertions
Review of the examples-smoke job surfaced three issues (observed on run 27349539957 and the PR #1083 timeout flake where google_adk hit 60.08s against the 60s budget):

1. **Mock LLM triggers extraction retry storms.** `examples/_helpers/mock_llm.py` defaults to `"stub response"`, which is not JSON. Every `remember()` runs the extraction pipeline through 3 tenacity attempts with exponential backoff (~6s of pure sleep per document) and logs `JSON parse error` warnings plus a final `Multi-extraction failed after 3 attempts` ERROR. Two example docstrings (openai_agents, hermes) explicitly shrank their scenarios to work around this instead of fixing the mock.

2. **Cross-encoder reranker download inside `recall()`.** `QuerySettings.enable_reranking` defaults to true, so the first recall in each example downloads `BAAI/bge-reranker-v2-m3` from HF Hub unauthenticated (rate-limit-prone) and loads it — ~26s of the google_adk example's ~51s wall time. This is the main reason the example sits within single-digit seconds of the 60s timeout.

3. **Examples assert nothing.** Apart from one `assert item is not None` in langgraph, the examples only print results — `Recovered 0 memory entries` would pass the smoke job. The mock-refresh loop in `install_mock_llm` also never patches non-litellm modules (the `khora.` prefix branch is dead code), so CrewAI's held reference to `litellm.completion` goes unpatched and produces `OpenAI API call failed: OPENAI_API_KEY is required` ERROR noise.

Fix: default the mock to a valid empty extraction payload, disable reranking in the `embedded_khora` fixture, patch held references by identity, and add real assertions to all six examples (with byte-identical doc sync).

filippetrovic:
## Summary
- Fixes #1099. Review of the examples-smoke job found the examples assert nothing, every ingest burns ~6s in extraction retry sleeps against the mock LLM's non-JSON default, and the first `recall()` downloads the cross-encoder reranker from HF Hub (~26s, unauthenticated → rate-limit prone). Together these put google_adk at ~51s of the 60s budget and caused the #1083 timeout flake at 60.08s.
- `examples/_helpers/mock_llm.py`: default completion is now a valid empty extraction payload (`{"entities": [], "relationships": []}`) so extraction succeeds first-attempt — no tenacity backoff sleeps, no `JSON parse error` / `Multi-extraction failed` log noise. The module-refresh loop now re-patches held references by identity across all loaded modules (the previous `khora.` prefix branch was dead code and non-litellm modules were never covered).
- `examples/_helpers/khora_fixtures.py`: `embedded_khora` disables cross-encoder reranking — the examples demonstrate adapter wiring, not ranking quality, and the HF Hub download was both the bulk of the wall time and a network flake source.
- Every example now asserts on its results instead of just printing. This immediately exposed that the crewai example's recall has returned zero matches in every CI run: hash-derived mock embeddings score near-zero (often negative) for paraphrased queries, and the adapter's 0.0 similarity floor drops them. crewai and hermes now recall verbatim text (cosine 1.0), the pattern the llamaindex example already documented; hermes additionally has a 0.1 floor on `memory_search`.
- `docs/integrations/*.md` snippets re-synced byte-identical (drift gate green), stale 30s-budget comments in `ci.yml` and `CLAUDE.md` refreshed.

Observed locally: google_adk ~51s → ~20s, crewai ~24s → ~5s, all six examples pass with assertions enabled.

## Test plan
- `uv run python tools/check_examples_drift.py` — 6 snippets byte-identical.
- All six examples run green locally via the CI smoke invocation (`PYTHONPATH=$REPO uv run --directory examples/integrations/<name> python example.py`), each well under the 60s budget.
- `uv run pytest tests/integration/integrations/ tests/unit/tools/test_examples_helpers.py` — 10 passed, 2 skipped (env-gated combos).
- `ruff check` / `ruff format --check` clean on changed files.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -560,10 +560,12 @@ jobs:
             # path — without this, the helper package is unreachable
             # because there's no ``examples/__init__.py`` at the root.
             #
-            # 60s budget (was 30s): the crewai example runs through the
-            # full ingest pipeline + LLM stub retries and observed times
-            # around 25-34s on GitHub-hosted runners. 30s was too tight
-            # and produced intermittent flakes (#741 / v0.15.1 release).
+            # 60s budget (was 30s): raised after intermittent flakes
+            # (#741 / v0.15.1 release). Since #1099 the mock LLM returns
+            # valid extraction JSON (no tenacity retry sleeps) and the
+            # fixture disables cross-encoder reranking (no HF Hub model
+            # download inside recall), so examples should sit well under
+            # the budget; 60s is headroom for runner variance.
             if ! PYTHONPATH="${GITHUB_WORKSPACE}" timeout 60s uv run --directory "$ex" python example.py; then
               echo "[examples-smoke] $name: FAILED"
               fail=1
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -96,7 +96,7 @@ Always run `make format && make test` before committing. CI will reject PRs that
 
 ### Integration examples
 
-Every adapter ships `examples/integrations/<name>/example.py` that runs without external services (sqlite_lance fixture + mock LLM helpers under `examples/_helpers/`). The `python title="example.py"` block in `docs/integrations/<name>.md` must be byte-identical to that file. The `examples-smoke` CI job gates drift via `tools/check_examples_drift.py` and smoke-runs each example under a 30s timeout.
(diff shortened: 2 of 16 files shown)


filippetrovic:
Closing: this approach syncs test scaffolding (assertions, verbatim-recall queries, mock-workaround comments) into the doc-mirrored example snippets, which degrades the documentation. The examples need a design where the documented file stays pure real usage and verification is injected by the CI harness instead — to be addressed separately. The helper fixes (valid-JSON mock default, identity-based reference patching, reranking disabled in the fixture) are independent of that question and can be resubmitted on their own.

Issue #1102, opened 2026-06-11
LLM extractor retries missing-credentials errors — deterministic failures burn ~6s of backoff per call


filippetrovic:
## Problem

The extraction retry policy in `extraction/extractors/llm.py` (tenacity retry around line ~1698) retries on `litellm.InternalServerError`. LiteLLM maps **missing credentials** (`OpenAIException - Missing credentials. Please pass an api_key ... or set the OPENAI_API_KEY ...`) to `InternalServerError`, so a deterministic, unrecoverable configuration error is retried 3 times with ~2s/~4s backoff — ~6s wasted per extraction call before degrading to an empty result.

Observed in CI: `tests/recall/test_filter_chronicle.py` burned ~9.5 minutes of the `test-unit` job this way (see #1101). Beyond tests, a misconfigured production deploy (missing/rotated key) would pay the same silent 6s penalty on every `remember()` while logging only WARNINGs.

## Fix

Detect credential/auth errors inside the retry predicate and fail fast (no retries) — e.g. match `AuthenticationError` and the known missing-credentials message shape inside `InternalServerError`. Genuine transient 5xx/server errors should keep the current retry behavior.

Tracked in Linear as IGR-289 (follow-up).

filippetrovic:
Fixed by #1105 (merged): the retry predicate now fails fast on `litellm.AuthenticationError` / `PermissionDeniedError` and on credential-shaped `InternalServerError` messages, while genuine transient server errors keep the full retry budget. Degradation semantics unchanged (empty ExtractionResult with error metadata). 16 regression tests pin attempt counts and the false-positive guard (tests/unit/test_extraction_retry_fail_fast.py).

Issue #1101, opened 2026-06-11
test_filter_chronicle.py burns ~6s of LiteLLM retries per remember() — 9.5 min of CI test-unit time


filippetrovic:
## Problem

The `test-unit` CI job spends ~9.5 of its ~13 pytest minutes on 12 tests in `tests/recall/test_filter_chronicle.py` (9 tests at ~61s, 3 at ~12s — see the 574s tail in [this run](https://github.com/DeytaHQ/khora/actions/runs/27329779751/job/80739459661)).

Every `remember()` in that file takes ~6s. The tests pass `entity_types=["PERSON"]` / `relationship_types=["KNOWS"]`, which fires live `gpt-4o-mini` entity extraction through LiteLLM. With no `OPENAI_API_KEY` in the environment the call fails, is retried 3 times with ~2s/~4s backoff (`extraction/extractors/llm.py`), then degrades gracefully to 0 entities. The tests only assert filter narrowing, so they pass anyway — silently burning 6s per remember. The 61s tests seed 10 rows each.

The file's autouse fixture already stubs the embedder and disables event/fact extraction via `ExpertiseConfig`; entity extraction is the one LLM path left live.

## Fix

Stub entity extraction the same way the embedder is stubbed (autouse monkeypatch of the LLM extractor returning empty results), or stop passing `entity_types`/`relationship_types` — no test in the file asserts on entities.

Expected effect: file goes from ~585s to ~10s of test time; `test-unit` wall time drops from ~15 min to ~5 min.

Optional while in there: add `--durations=25` to the CI pytest invocation so slow-test regressions are visible in the job log.

Tracked in Linear as IGR-289.

PR #1105, opened 2026-06-11
Add fail-fast on credential errors for LLM extractor

filippetrovic:
## Summary

Implement fail-fast credential-error detection to avoid wasted retries on permanent auth failures. Adds `_is_nonretryable_auth_error()` helper that detects:

1. Genuine `litellm.AuthenticationError` exceptions
2. Missing-credentials error messages (conservative substring matching)

Wired into both AsyncRetrying blocks in `extract()` and `_extract_multi_batch()` via `retry_if_exception` predicate.

## Impact

- Reduces wasted retry attempts on permanent failures (3 retries × ~60s backoff = ~180s saved)
- Maintains retry for transient errors (rate limits, timeouts, gateway errors)
- Defensive type resolution handles litellm import failures

## Test plan

- [ ] Run `make test` — full test suite passes
- [ ] Manually verify non-retryable errors return `ExtractionResult` with error metadata (no raise)
- [ ] Verify transient errors still retry normally

### DIFF
--- a/src/khora/extraction/extractors/llm.py
+++ b/src/khora/extraction/extractors/llm.py
@@ -79,45 +79,79 @@ def _repair_json(content: str) -> str:
     return content
 
 
+# Deterministic auth/authz exception types. LiteLLM normalizes provider auth
+# failures onto these regardless of underlying SDK: a bad/expired/invalid key or
+# HTTP 401 → AuthenticationError; an authorization denial (e.g. Bedrock
+# AccessDeniedException, Vertex/GCP IAM "not authorized") → PermissionDeniedError.
+# Both are siblings of one another and of InternalServerError/RateLimitError, so
+# matching them by name does NOT catch transient 5xx or 429s.
+_NONRETRYABLE_AUTH_EXC_NAMES = frozenset({"AuthenticationError", "PermissionDeniedError"})
+
+# Message-shape match for credentials-absent errors that litellm surfaces as
+# InternalServerError (client-construction failures, before any HTTP request).
+# Both a credential token AND a missing/absent signal must be present so we never
+# fail-fast a RETRYABLE error that merely echoes a key token (e.g. a 429 reading
+# "Rate limit reached ... on api_key sk-...").
+_CREDENTIAL_TOKENS = ("credential", "api_key", "api key", "_api_key", "api-key")
+_MISSING_CREDENTIAL_SIGNALS = (
+    "missing",
+    "no api",
+    "not set",
+    "pass an api",
+    "set the",
+    "provide an api",
+    "could not resolve auth",
+    "authentication",
+)
+
+
 def _is_nonretryable_auth_error(exc: BaseException) -> bool:
     """Return True for deterministic credential/auth failures that must NOT be retried.
 
-    Retrying a missing-credentials error just burns three attempts and ~180s of
-    backoff for a result that will never change. We treat two shapes as
+    Retrying a missing-credentials or auth error just burns three attempts and
+    ~180s of backoff for a result that will never change. We treat two shapes as
     non-retryable:
 
-    1. A genuine ``litellm.AuthenticationError``. This is *always* non-retryable.
-       litellm is imported lazily inside the extractor methods, so resolve the
-       type defensively: try the real class, and also walk the MRO for a class
-       literally named ``AuthenticationError`` so the check still works if the
-       litellm import fails here.
-    2. A missing-credentials message. LiteLLM maps absent credentials onto
-       ``InternalServerError`` (a sibling of AuthenticationError, not a
-       subclass), e.g. "OpenAIException - Missing credentials. Please pass an
-       api_key, or set the OPENAI_API_KEY environment variable." We match
-       conservatively to avoid swallowing transient 5xx errors: require
-       "credentials" together with "missing", OR an explicit api-key hint.
+    1. A genuine ``litellm.AuthenticationError`` or ``litellm.PermissionDeniedError``.
+       This is *always* non-retryable. litellm normalizes provider auth/authz
+       failures (invalid key, 401, IAM denial) onto these types. litellm is
+       imported lazily inside the extractor methods, so resolve the type
+       defensively: try the real classes, and also walk the MRO for a class
+       literally named ``AuthenticationError`` / ``PermissionDeniedError`` so the
+       check still works if the litellm import fails here.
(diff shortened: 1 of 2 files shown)


filippetrovic:
Implementation reviewed — predicate logic and wiring are correct. Verified against pinned versions (tenacity 9.1.4, litellm 1.84.0): when `retry_if_exception` returns False, tenacity re-raises the original exception, which lands in the same `except Exception` degrade handlers as the exhaustion path — degradation semantics (empty `ExtractionResult` with error metadata, never a raise) are preserved and only retries are skipped. The token+signal message match is conservative; OpenAI 429/5xx shapes don't false-positive, and the real missing-key shape ("Missing credentials. Please pass an `api_key`, … or set the `OPENAI_API_KEY` environment variable", wrapped by litellm as `InternalServerError`) matches via `api_key` + `missing`/`set the`.

**Blocking: no tests.** This changes the retry policy for every extraction call (#1101/#1102 context), and the fail-fast behavior depends on tenacity internals (predicate-False → original-exception re-raise). Add unit tests that mock `litellm.acompletion` and assert **attempt counts**, not just the final result:

1. `InternalServerError` with the missing-credentials message → exactly 1 attempt, no retry sleep, returns `ExtractionResult` with error metadata.
2. `InternalServerError` with a generic message ("the server had an error") → retried up to `_max_retries`.
3. `litellm.AuthenticationError` (any message) → exactly 1 attempt.
4. Same trio for `_extract_multi_batch` (or parametrize predicate tests + one test per retry block).
5. A direct `_is_nonretryable_auth_error` test pinning the false-positive guard: a 429-shaped message echoing "api_key" without a missing/absent signal stays retryable.

Non-blocking:
- The ERROR log "Extraction failed after {self._max_retries} attempts" is wrong on the fail-fast path (1 attempt) — the exact debugging scenario this PR targets. Suggest "Extraction failed (non-retryable): …" when the predicate fired.
- Add `Fixes #1102` to the body; the "3 retries × ~60s = ~180s" figure is the stop-cap, not the typical waste (~6s/call per #1101).

filippetrovic:
## Multi-perspective review — fail-fast on credential/auth errors

7 independent reviewers (security, performance, correctness, test, docs, AI/LLM-pipeline, devil's-advocate) plus a diff-scoped security audit. **No blocking findings.**

**Overall: ✅ APPROVE / merge.** The change is correct, surgical (2 files), and well-tested. Both `AsyncRetrying` blocks (`extract()` and `_extract_multi_batch()`) are patched identically with `retry=retry_if_exception(lambda e: not _is_nonretryable_auth_error(e))` and keep `reraise=True`, so deterministic auth/credential failures fail fast (one attempt, no backoff) and still return an error `ExtractionResult` — they don't raise out. Genuine transient 5xx / rate-limit / timeout errors retry exactly as before.

### CRITICAL
None.

### HIGH
None.

### MEDIUM
- **M1 [Security] Pre-existing, not introduced** — `src/khora/extraction/extractors/llm.py` (`before_sleep` log ~860-872 / ~1784-1796 and the `except` handler ~972/~1947) logs/stores the raw exception string, which for some providers can contain an API key. This predates the PR and is *reduced* by it (fewer retries → fewer log lines). Per per-PR audit scope it's not a finding for this change; worth a separate log-scrubbing follow-up.
- **M2 [Scope] Sibling retry site not covered** — `src/khora/extraction/embedders/litellm.py:399` uses the same `AsyncRetrying(stop=…, wait=…)` with **no** `retry=` predicate, so embedding calls still burn full backoff on missing-credential errors (and embeddings run per-chunk, so the waste compounds on bulk ingest). Out of scope for this change (which targets the extractor); recommend a follow-up to apply the same predicate there.
- **M3 [Maintainability] Message heuristic may drift** — the secondary message branch (`_CREDENTIAL_TOKENS` × `_MISSING_CREDENTIAL_SIGNALS`) is provider-wording-sensitive. Mitigated by the primary branch being type-based (`AuthenticationError`/`PermissionDeniedError`), which litellm normalizes any 401/403 onto by contract; the message branch is only the net for the client-construction "missing credentials" case litellm surfaces as `InternalServerError`. Acceptable as-is; revisit if a new provider's wording slips through.

### LOW
- **L1 [Correctness]** The defensive `except Exception: pass` around the lazy `import litellm` is broader than needed (`except ImportError` would be tighter). Intentional and harmless — optional tightening.
- **L2 [AI]** The second-pass relationship extractor (`_extract_additional_relationships`) isn't wrapped in retry/fail-fast. Largely moot: it only runs when the first pass returned entities, and a fail-fast first pass returns none — so missing-creds is caught before the second pass executes. Optional follow-up.
- **L3 [Test]** No explicit test asserting transient `Timeout`/connection errors are retried (they are, by default predicate). Nice-to-have.

### Security Audit
Diff-scoped scan of added/modified lines: no dangerous sinks (no eval/exec/subprocess/deserialization/template/redirect/file or query sinks), no secrets committed, no new third-party dependencies (only `tenacity.retry_if_exception`, already a dependency). The only security-adjacent item is the pre-existing M1 logging, out of scope per per-PR audit rules. **Clean.**

### Tests
Comprehensive (16 tests): predicate unit tests (auth types, message shapes, MRO-name fallback, subclass-via-MRO), both code paths (`extract()` + `_extract_multi_batch()`) asserting exactly one attempt and no backoff sleep on fail-fast, and regression guards proving generic 5xx **and** a `RateLimitError` mentioning `api_key` still retry the full budget (the key false-positive guard). Mocks construct **real** litellm exception instances (so `isinstance` is exercised), not patched classes.

filippetrovic:
Re-reviewed at e6de03d — blocking item resolved. The new test file (tests/unit/test_extraction_retry_fail_fast.py, 16 tests) covers the full matrix with attempt-count + sleep assertions: credential-shaped InternalServerError → 1 attempt/no sleep; transient 5xx → full retry budget; AuthenticationError and PermissionDeniedError → 1 attempt; _extract_multi_batch block covered directly; false-positive guard pinned at both predicate and integration level (429 echoing api_key stays retryable). Verified the asyncio.sleep patch genuinely intercepts tenacity backoff (call-time resolution in 9.1.4) and CI collection (+16 vs prior head, all passed). Remaining nits are non-blocking: the fail-fast path still logs "failed after {max_retries} attempts" (1 attempt was made), and the body could carry Fixes #1102. Good to merge.

probablyrobot (approved):

PR #1106, opened 2026-06-11
Stub entity extraction in chronicle recall-filter tests

### LINKED ISSUE #1101: test_filter_chronicle.py burns ~6s of LiteLLM retries per remember() — 9.5 min of CI test-unit time

filippetrovic:
## Summary
- The Chronicle recall-filter integration tests (`tests/recall/test_filter_chronicle.py`) passed `entity_types=["PERSON"]` / `relationship_types=["KNOWS"]` to every `remember()` call, which fired **live** LiteLLM entity extraction. With no API key configured, each call retried with exponential backoff before degrading to zero entities — roughly 6s per `remember()`, ~9.5 minutes across the file.
- No test in the file asserts on extracted entities (the corpus narrows only on `source_name`, `occurred_at`, and `metadata.tag`). This adds an autouse fixture that stubs `LLMEntityExtractor.extract_multi` to return one empty `ExtractionResult` per chunk, mirroring the existing embedder stub (`_patch_embedder`).
- This removes the only remaining live-LLM path in the file — event/fact extraction was already disabled via the per-test expertise config — so the suite is now fully hermetic. The 7 `remember()` call sites are unchanged, preserving the realistic call shape; only the extractor is intercepted.
- Net effect: file runtime drops from ~585s to ~8s with no change to any filter-narrowing assertion.

## Test plan
- [x] `env -u OPENAI_API_KEY uv run pytest tests/recall/test_filter_chronicle.py -p no:cacheprovider --no-cov -q` → **12 passed in ~7s** (was ~585s), `OPENAI_API_KEY` unset to prove no live LLM call occurs
- [x] `make format` clean, `make lint` exit 0 (no new diagnostics in the changed file)
- [ ] CI `test-unit` job green

Fixes #1101

### DIFF
--- a/tests/recall/test_filter_chronicle.py
+++ b/tests/recall/test_filter_chronicle.py
@@ -78,6 +78,7 @@ except ImportError:
 
 from khora.config import KhoraConfig
 from khora.config.schema import SQLiteLanceConfig
+from khora.extraction.extractors.base import ExtractionResult
 from khora.extraction.skills import ExpertiseConfig
 from khora.extraction.skills.base import EventExtractionConfig, FactExtractionConfig
 from khora.filter import telemetry as filter_telemetry
@@ -130,6 +131,38 @@ def _patch_embedder(monkeypatch: pytest.MonkeyPatch) -> None:
     monkeypatch.setattr(
         "khora.extraction.embedders.litellm.LiteLLMEmbedder.embed",
         _stub_embed,
+    )
+
+
+# ---------------------------------------------------------------------------
+# Entity-extraction stub — no OPENAI_API_KEY required.
+# ---------------------------------------------------------------------------
+#
+# Every ``remember`` here passes ``entity_types=["PERSON"]`` /
+# ``relationship_types=["KNOWS"]``, so the ingest pipeline's ``extract_entities``
+# task calls ``LLMEntityExtractor.extract_multi`` — a LIVE LiteLLM call. Without
+# an API key each call retries with exponential backoff before degrading to zero
+# entities, which is slow and network-dependent. No test in this file asserts on
+# extracted entities (the corpus filters only on source_name / occurred_at /
+# metadata.tag), so stubbing the extractor to return empty results keeps these
+# tests hermetic and fast while leaving the filter-narrowing contract untouched.
+#
+# ``extract_multi`` is the method the pipeline calls; the consumer zips its result
+# one-to-one with the input chunks, so the stub returns exactly one empty
+# ``ExtractionResult`` per text. (Events/facts extraction is already disabled via
+# ``_no_event_fact_extraction()``, so the entity extractor is the only live LLM
+# extractor left to stub.)
+
+
+async def _stub_extract_multi(self: Any, texts: list[str], *args: Any, **kwargs: Any) -> list[ExtractionResult]:
+    return [ExtractionResult() for _ in texts]
+
+
+@pytest.fixture(autouse=True)
+def _patch_entity_extractor(monkeypatch: pytest.MonkeyPatch) -> None:
+    monkeypatch.setattr(
+        "khora.extraction.extractors.llm.LLMEntityExtractor.extract_multi",
+        _stub_extract_multi,
     )
 
 


filippetrovic:

Reviewed by a 6-perspective team (Security, Performance, Correctness, Test-quality, Docs, Devil's Advocate) plus a dedicated security-audit pass. The change is a single test file (+33 / −0): an autouse fixture that monkeypatches `LLMEntityExtractor.extract_multi` to return one empty `ExtractionResult` per chunk, removing the only remaining live-LLM path in the recall-filter integration tests.

**Overall assessment: ✅ APPROVE.** No CRITICAL / HIGH / MEDIUM findings. Two optional LOW notes below; neither blocks merge.

### Verified clean
- **Correctness** — patch target `khora.extraction.extractors.llm.LLMEntityExtractor.extract_multi` is exactly the method the ingest pipeline invokes (`pipelines/tasks/extract.py`), the consumer zips results 1:1 with chunks, and the stub returns one result per text. `ExtractionResult()` default-constructs to empty entities/relationships/events. `extract`/`extract_batch` are not on the pipeline path, so no bypass.
- **Performance** — the stub eliminates exponential-backoff retry on a missing API key; every other LLM path is already off (embedder stubbed, events/facts disabled, reranking off). Measured locally: ~585s → ~7–8s for the file, 12 passed.
- **Test-quality** — no test asserts on extracted entities (assertions are on `result.chunks` filtered by `source_name`/`occurred_at`/`metadata.tag`). The constant-vector embedder makes the filter the sole narrowing force, so the tests still genuinely prove filter narrowing. Tests remain real sqlite_lance integration runs.
- **Security** — no secrets, no injection/network/deserialization/file/auth surface, no dependency changes. Monkeypatch is pytest-scoped and auto-reverted. Security-audit verdict: PASS.
- **Docs** — test-internal only; no public API, telemetry-contract, or doc updates required. In-file comment block is accurate and matches the existing `_patch_embedder` style.

### LOW (optional — not blocking)
- **L1 [robustness]** The stub intercepts `extract_multi` only. If the pipeline is later rewired to call a different extractor method, the live call could silently return (tests would still pass, just slow again). The measured speedup proves the stub is wired correctly *today*, and this matches the file's existing `_patch_embedder` convention (no call-assertion). A future call-count assertion would harden against silent regressions but is out of scope for this change.
- **L2 [style]** The stub uses `*args, **kwargs` while the real method has a keyword-only barrier. Standard and correct for a test stub; no action needed.

_Note: an initial "event/fact extraction status unclear" concern was a false positive (raised from the diff alone). `_no_event_fact_extraction()` is defined at line 174 (`events`/`facts` disabled) and wired at all 7 `remember()` call sites — confirmed independently by the performance review._

probablyrobot (approved):

PR #1107, opened 2026-06-11
Reset OpenTelemetry providers after OTLP smoke test to prevent cross-test tracer leak

filippetrovic:
## Summary
- The OTLP smoke test installed a real `TracerProvider` as a process-wide singleton but never reset it (`handle.shutdown()` only force-flushes). The leaked provider survived into later tests in the same `pytest -n auto` worker.
- Once a real provider is active, OpenTelemetry records exceptions on span `__exit__` by calling `str(exc)`. neo4j 6.x errors built via `Neo4jError._basic_hydrate` (a test-only shortcut for faking server codes) lack `_gql_status`, so `str()` raised `AttributeError` — masking the real `ClientError` a downstream engine test asserted on. The result was an order-dependent flake that surfaced only under the full parallel suite.
- **Fix:** restore the proxy providers in the smoke test's `finally` block.
- Unify the previously-duplicated OTel reset logic into a single shared `tests/test_helpers/otel.py` (it had been copied verbatim into the smoke test and the telemetry conftest).
- Add deterministic, order-independent regression tests that document the masking mechanism and assert the smoke test's teardown restores the proxy provider (they fail if the teardown is removed).
- **No production code change.** In production, neo4j errors are hydrated from real server responses that carry the GQL status fields, so `str(exc)` is safe; only the test's bare `_basic_hydrate` mock omitted them. The bug is purely a test-isolation defect.

## Test plan
- [x] Affected suite passes: smoke + isolation + dual_nodes + telemetry (155 passed)
- [x] Same set under `-n auto` (parallel) passes — mimics the original flake scenario
- [x] Regression fails without the teardown, passes with it (verified both directions)
- [ ] Full CI suite green

### DIFF
--- a/tests/integration/test_otel_smoke.py
+++ b/tests/integration/test_otel_smoke.py
@@ -23,6 +23,7 @@ from http.server import BaseHTTPRequestHandler, HTTPServer
 import pytest
 
 from khora.telemetry import bootstrap, trace_span
+from tests.test_helpers.otel import reset_khora_telemetry, reset_otel_globals
 
 pytestmark = pytest.mark.integration
 
@@ -52,25 +53,11 @@ def _free_port() -> int:
         return s.getsockname()[1]
 
 
-def _reset_otel_globals() -> None:
-    import opentelemetry.metrics._internal as _m
-    import opentelemetry.trace as _t
-    from opentelemetry.metrics._internal import _ProxyMeterProvider
-    from opentelemetry.trace import ProxyTracerProvider
-
-    _t._TRACER_PROVIDER_SET_ONCE = _t.Once()
-    _t._TRACER_PROVIDER = None
-    _t._PROXY_TRACER_PROVIDER = ProxyTracerProvider()
-    _m._METER_PROVIDER_SET_ONCE = _m.Once()
-    _m._METER_PROVIDER = None
-    _m._PROXY_METER_PROVIDER = _ProxyMeterProvider()
-
-
 def test_spans_reach_otlp_endpoint(monkeypatch: pytest.MonkeyPatch) -> None:
     """End-to-end: trace_span() with OTLP/HTTP ships to the endpoint."""
     # Reset state.
     monkeypatch.setattr(bootstrap, "_handle", None)
-    _reset_otel_globals()
+    reset_otel_globals()
     _CountingHandler.requests = []
 
     # Stand up a counting OTLP server on a free port.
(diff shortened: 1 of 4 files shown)


filippetrovic:

**Overall: ✅ APPROVE** — no blocking findings. Panel of 7 reviewers (Security, Performance, Correctness, Test, Docs, Observability, Devil's Advocate) + a dedicated security audit. The change is test-only (no `src/` changes).

### CRITICAL
_None._

### HIGH
_None._

### MEDIUM
- **M1 — [Test / Devil's Advocate]** `tests/integration/test_otel_smoke_isolation.py` — the regression tests call module-internal predicates `bootstrap._tracer_provider_is_proxy()` / `_meter_provider_is_proxy()` (leading underscore, not in `__all__`). Functional, but couples the tests to a private surface. _Lighter touch than promoting them to public API (which would expand the telemetry surface): add a brief comment that these are intentionally test-facing internal predicates._ Non-blocking.
- **M2 — [Devil's Advocate / Observability]** Duplication sweep is intentionally partial. This PR unifies the two in-scope copies of the OTel reset logic (smoke test + telemetry conftest) into `tests/test_helpers/otel.py`, but four other test files still carry their own local `_reset_otel_globals` — `tests/unit/telemetry/test_bootstrap.py`, `tests/unit/engines/vectorcypher/test_llm_rerank_gate.py`, `tests/unit/test_khora_recall_upgrade.py`, `tests/unit/filter/test_filter_telemetry.py`. These are a pre-existing condition outside this change's scope; recommend a **follow-up tidy PR** to converge them on the shared helper.

### LOW
- **L1 — [Observability]** `tests/test_helpers/otel.py` — `reset_otel_globals()` reaches into OTel private internals (`_TRACER_PROVIDER_SET_ONCE`, `_PROXY_TRACER_PROVIDER`, …). Consider a docstring note that this is test-only and breaks loudly at import time (not silently) if OTel internals shift.
- **L2 — [Test]** `tests/integration/test_otel_smoke_isolation.py` — optional clarifying comment that the async exception propagates through the `finally` cleanup block.

### 🔒 Security Audit — PASS
Test-only diff: no trust boundaries, endpoints, or dangerous sinks; no secrets, no weak crypto, no new dependencies. The OTel process-global manipulation is confined to test helpers (verified: zero reach-through into `src/`). Nothing to fix.

### Verified strengths (panel consensus)
- **Production-no-change claim independently re-verified correct.** Real neo4j 6.x errors are hydrated by the driver's bolt I/O path (`_hydrate_gql` / `_hydrate_neo4j`), which always calls `_init_gql` and sets `_gql_status`; only the test's bare `Neo4jError._basic_hydrate(...)` omits it. `dual_nodes.py` is untouched and the previously-reverted `exc._gql_status = "50N42"` symptom patch was **not** reintroduced.
- **Regression suite is a genuine tripwire.** The 3 tests are deterministic and order-independent; `test_smoke_test_resets_global_provider_in_teardown` fails if the smoke test's `finally`-block teardown is removed.
- **Fix is surgical and correct.** Teardown correctly placed in `finally`; `reset_khora_telemetry()` fully restores OTel globals + khora's cached `_TRACER`/`_METER` + `bootstrap._handle`; the shared-helper extraction removes verbatim duplication.

**Assessment: APPROVE.** The two MEDIUM items are maintainability/follow-up (not correctness or merge blockers); the LOWs are doc polish. No changes are required to merge.

probablyrobot (approved):

PR #1109, opened 2026-06-12
Bound entity extraction so it cannot hang on truncated-output recovery

### LINKED ISSUE #1113: Entity extraction can hang indefinitely on truncated-output recovery (bisection)
## Summary

`LLMEntityExtractor` can hang **indefinitely** during entity extraction when the model returns truncated output and the batch enters the bisection-recovery path. The process stays alive but makes zero progress — no timeout, no exception, no failure marker.

## Where

`src/khora/extraction/extractors/llm.py` — `extract_batch` → `_extract_multi_batch` → the truncation→bisection recovery (`_bisect_and_extract`), and the single-doc floor (`extract`).

## Root cause

The per-call `asyncio.wait_for(litellm.acompletion(...), ...)` added in #1059 bounds only the network call. It does **not** cover:

1. the bisection `asyncio.gather(...)` fan-out, nor
2. the shared `asyncio.Semaphore` **acquire** — `async with self._semaphore:` is entered *outside* the `wait_for`, and `Semaphore.acquire()` takes no timeout. A single extractor instance is shared across many concurrent documents, so one wedged permit-holder parks every other extraction coroutine on `acquire()`.

Two further factors make even the per-call deadline non-binding:

3. The calls run inside a tenacity `AsyncRetrying` loop whose retry predicate matched `CancelledError`, so a deadline's injected cancellation was *retried* instead of propagating.
4. The aiohttp transport builds its per-request `ClientTimeout` without `total` (only `sock_read`), so a connection that completes TLS then stalls is never cancelled at the transport layer — and `asyncio.wait_for` / `asyncio.timeout` cannot free a coroutine stuck in an await that ignores cancellation.

Net effect: a stalled / half-open LLM connection (or a starved semaphore permit) freezes extraction with the process alive.

## Reproduction

Stub `litellm.acompletion` so it swallows cancellation (or never returns) and run `extract_batch` over a few documents — extraction never completes. A deterministic regression covering this shape is added in `tests/unit/test_extraction_hang_guards.py`.

## Fix

Tracked in #1109:
- exclude `CancelledError` / `TimeoutError` from the extraction retry predicate so a deadline propagates;
- run each `litellm.acompletion` as a shielded task and abandon it on the per-call deadline (bounds even an uncancellable await);
- bound the semaphore acquire under a wall-clock deadline;
- add a coarse aggregate per-batch backstop with idempotent failure accounting;
- mark a timed-out document as failed (with a truthy, countable error) and continue the run instead of producing silently-empty output.

sale87:
## Summary

Fixes #1113. Follow-up to #1059. That PR wrapped every `litellm.acompletion` in `asyncio.wait_for`, but extraction could **still hang indefinitely** in the truncation→bisection path (process alive, zero progress). The deadline was **mis-scoped, not missing**:

- `asyncio.wait_for` bounds only the network call — not the bisection `asyncio.gather(...)` fan-out, nor the shared `asyncio.Semaphore` **acquire** (`async with self._semaphore:` is entered *outside* the `wait_for`; `Semaphore.acquire()` has no timeout). One wedged permit-holder, across a single extractor shared by many concurrent documents, parks every sibling on `acquire()`.
- The calls run inside a tenacity `AsyncRetrying` loop whose predicate matched `CancelledError`, so a deadline's cancellation was *retried* instead of propagating.
- The aiohttp transport builds its per-request `ClientTimeout` with no `total` (only `sock_read`), so a half-open socket is never cancelled at the transport layer — and `asyncio.wait_for` / `asyncio.timeout` cannot free a coroutine stuck in an await that ignores cancellation.

## Changes

- Exclude `CancelledError` / `TimeoutError` from the extraction retry predicate so a deadline propagates instead of being retried.
- `_bounded_acompletion`: run each `litellm.acompletion` as a shielded task and abandon it on the per-call deadline — bounds even an await that ignores cancellation.
- `_acquire_slot`: bound the shared-semaphore acquire under a wall-clock deadline.
- A coarse aggregate per-batch backstop with idempotent failure accounting.
- A timed-out document is marked failed (truthy, countable error), logged + counted, and the run continues instead of producing silently-empty output.

## Test plan

`tests/unit/test_extraction_hang_guards.py`: the retry predicate never retries cancellation/timeout; `extract_batch` returns within budget when `litellm.acompletion` swallows cancellation (the real hang shape); the semaphore acquire times out when starved; the aggregate budget fires on a stalled inner call; the breaker increments exactly once per timed-out batch. Locally these + `test_extraction_llm.py` + `test_llm_call_timeout.py` pass; `ruff` clean. Full suite runs in CI.

### DIFF
--- a/src/khora/extraction/extractors/llm.py
+++ b/src/khora/extraction/extractors/llm.py
@@ -5,6 +5,8 @@ from __future__ import annotations
 import asyncio
 import json
 import re
+from collections.abc import AsyncIterator
+from contextlib import asynccontextmanager
 from typing import TYPE_CHECKING, Any
 
 from loguru import logger
@@ -16,7 +18,7 @@ from tenacity import (
     wait_exponential,
 )
 
-from khora.config.llm import get_shared_session, llm_call_timeout
+from khora.config.llm import DEFAULT_LLM_TIMEOUT_S, get_shared_session, llm_call_timeout
 from khora.extraction.embedders._request_telemetry import (
     parse_rate_limit_headers,
     set_connector_attributes,
@@ -103,6 +105,19 @@ _MISSING_CREDENTIAL_SIGNALS = (
     "could not resolve auth",
     "authentication",
 )
+
+
+def _is_retryable_extraction_exc(exc: BaseException) -> bool:
+    """Retry predicate for extraction LLM calls (used by ``AsyncRetrying``).
+
+    Never retry cancellation or a timeout: a retried ``CancelledError`` would
+    defeat any surrounding ``asyncio`` deadline (the call is re-attempted instead
+    of unwinding) — exactly how the #1113 hang survived #1059's per-call
+    ``asyncio.wait_for``. Also never retry deterministic auth/credential failures.
+    """
+    if isinstance(exc, (asyncio.CancelledError, TimeoutError)):
+        return False
+    return not _is_nonretryable_auth_error(exc)
 
 
 def _is_nonretryable_auth_error(exc: BaseException) -> bool:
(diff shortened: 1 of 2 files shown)


sale87:
## Round-2 adversarial review (6 reviewers) → fixes applied

**Confirmed resolved (reproduced empirically):** round-1 HIGH #1 (tenacity defeating the deadline) and #2 (false-green test). All six reviewers: the hang is genuinely fixed — an uncancellable `acompletion` now returns control in ~0.2s.

**Round-2 findings, now fixed in `73bea7b1`→latest:**
- 🔴 **HIGH — falsy timeout error.** A per-call timeout raised a bare `TimeoutError` (`str()==''`, falsy), so `error_count`, `all_failed`, and the breaker all treated it as "no error" → silently-empty extraction (round-1 #3 via another door). Fixed: both except handlers now use `str(e) or type(e).__name__`, so a timeout is a truthy, countable error → WARNING fires, breaker increments, single-doc fallback runs.
- 🟠 **MEDIUM — budget undersized the linear fallback.** `_batch_deadline` was logarithmic but the single-doc fallback runs up to `batch_size` sequential calls under one aggregate timeout → completed work discarded on a slow-but-healthy backend. Fixed: budget is now `per_call * (batch_size + 1)`.
- Test strengthened to assert timed-out docs carry a truthy error.

**Flagged, not blocking (follow-up):** abandoned-task / connection-pool pressure under a *sustained* stuck backend (LOW–MEDIUM; not a hang — the breaker also trips and throttles); the budget is still loose at very high `timeout` (per-call `_bounded_acompletion` is the primary bound). Local: 68 tests pass, ruff clean.

msrdic (approved):

PR #1110, opened 2026-06-12
Make sqlite_lance LanceDB coarse temporal pre-filter upper bound inclusive

filippetrovic:
## Summary
- The `sqlite_lance` vector channel applies its temporal recency window in two layers: a coarse LanceDB pre-filter on `created_at`, then a precise SQLite refinement on `COALESCE(source_timestamp, created_at)`. The coarse pre-filter used a strict `<` on the upper bound while the SQLite refinement is inclusive (`<=`).
- Because the coarse pre-filter runs first and a post-filter can only drop rows (never restore them), a chunk with `source_timestamp` NULL and `created_at` equal to the upper-bound instant *exactly* was false-excluded at the LanceDB layer before the SQLite refinement could keep it — violating the superset-safe pushdown contract (the pushed candidate set must be a superset of the post-filtered set).
- Widen the coarse pre-filter upper bound to inclusive `<=`. LanceDB stores only `created_at` (not `source_timestamp`), so it cannot apply `COALESCE`; the inclusive bound keeps the coarse set a safe over-approximation while the SQLite refinement still narrows precisely. The lower bound (already `>=`), the SQLite refinement, and the BM25 channel are unchanged.
- Add at-boundary regression tests through the vector channel's storage-level recency window (`search_similar_chunks(..., created_before/created_after=bound)`): seed `source_timestamp` NULL with `created_at == bound` exactly and assert the row is returned, for both the upper and lower window bounds. The upper-bound test fails against the old strict `<` and passes with the inclusive `<=`.

### Known limitation (pre-existing, unchanged by this PR)
The coarse LanceDB pre-filter can only see `created_at`, so it cannot apply the `COALESCE(source_timestamp, created_at)` rule at that layer. A chunk whose `source_timestamp` is well *before* the bound but whose `created_at` is *after* it can still be coarse-excluded before the SQLite refinement runs. This assumes `created_at` and `source_timestamp` are roughly aligned at ingest (far-back-dated backfills are the outlier); it is a pre-existing property of a `created_at`-only ANN-layer filter and is not introduced or masked by this change.

## Test plan
- [x] New embedded regression tests pass (`tests/integration/test_chronicle_filter_embedded.py`, 8 passed); upper-bound guard verified to fail without the fix and pass with it
- [ ] Full Docker-backed integration + e2e suite and coverage gate run in CI (Docker is unavailable in this environment)

### DIFF
--- a/src/khora/storage/backends/sqlite_lance/vector.py
+++ b/src/khora/storage/backends/sqlite_lance/vector.py
@@ -485,7 +485,16 @@ class SQLiteLanceVectorAdapter:
             # ISO form since LanceDB's SQL parser accepts timestamp literals.
             where.append(f"created_at >= timestamp '{created_after.isoformat()}'")
         if created_before is not None:
-            where.append(f"created_at < timestamp '{created_before.isoformat()}'")
+            # Coarse pre-filter upper bound is INCLUSIVE (``<=``) on
+            # ``created_at`` to stay superset-safe: this LanceDB layer only
+            # stores ``created_at`` (not ``source_timestamp``), so it cannot
+            # apply the ``COALESCE(source_timestamp, created_at)`` rule. A
+            # strict ``<`` would false-exclude a chunk whose source_timestamp
+            # is NULL and created_at equals the bound exactly, before the
+            # SQLite refinement runs — and a post-filter can only drop rows,
+            # never restore them. The precise COALESCE-based bound is applied
+            # on the SQLite side below.
+            where.append(f"created_at <= timestamp '{created_before.isoformat()}'")
         where_sql = " AND ".join(where)
 
         await self._maybe_build_chunks_index()
--- a/tests/integration/test_chronicle_filter_embedded.py
+++ b/tests/integration/test_chronicle_filter_embedded.py
@@ -78,6 +78,8 @@ _FILTER_LB = "2026-01-01T00:00:00Z"
 # 2026-06-01T00:00:00Z is the ISO form of _IN_RANGE, so a chunk seeded with
 # source_timestamp == _IN_RANGE sits on the boundary of a $lte filter at this value.
 _FILTER_UB_AT_BOUND = "2026-06-01T00:00:00Z"
+# A lower-bound instant used by the paired at-boundary recency-window test below.
+_LB_BOUND = datetime(2026, 1, 1, tzinfo=UTC)
 
 
 def _filter_ast(wire: dict) -> Any:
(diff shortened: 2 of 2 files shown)


filippetrovic:

Reviewed by a 7-role team (security, performance, correctness, test, database, docs, devil's advocate) plus a dedicated security audit. The change widens the `sqlite_lance` vector-channel coarse LanceDB pre-filter upper bound on `created_at` from strict `<` to inclusive `<=` (one operator + rationale comment) and adds two at-boundary regression tests.

**Overall assessment: ✅ APPROVE** — no CRITICAL or HIGH findings. The fix is correct, surgically scoped, superset-safe, and aligned with the sibling backends.

### CRITICAL
- None.

### HIGH
- None.

### MEDIUM
- **M1 — [Security / Injection] (pre-existing, NOT introduced by this PR)** `src/khora/storage/backends/sqlite_lance/vector.py:497` — the LanceDB where-clause is built by f-string interpolation (`created_at <= timestamp '{created_before.isoformat()}'`). Assessed **safe**: `created_before` is a strongly-typed `datetime` and `.isoformat()` emits a fixed ISO-8601 string; interpolated UUIDs come from `uuid_to_text()` (hex). No attacker-controllable free text reaches the literal, and the namespace is independently enforced (parameterized) on the SQLite refinement layer (defense-in-depth). The `<`→`<=` change does not alter this surface. No action required for this PR; flagged only for awareness.

### LOW
- **L1 — [Test / Style]** `tests/integration/test_chronicle_filter_embedded.py` — minor naming asymmetry between `_IN_RANGE` (the upper-bound instant) and `_LB_BOUND` (the lower-bound instant). Cosmetic; not a defect.
- **L2 — [Docs / Comment]** `src/khora/storage/backends/sqlite_lance/vector.py:488-496` — the comment explains *why the bound is inclusive* (superset-safe) but not *why* the LanceDB layer lacks `source_timestamp` (it stores only `created_at` / embeddings). Sufficient for maintainers; optional to expand.

### INFO
- **Correctness / Database:** the `<`→`<=` widening is superset-safe — the coarse candidate set stays a superset of the post-filtered set, and the precise SQLite `COALESCE(source_timestamp, created_at) <=` refinement still narrows exactly (and correctly drops a row whose `source_timestamp` is just after the bound while `created_at == bound`). Lower bound (`>=`), the SQLite refinement, and the BM25 channel are untouched. The four pushdown sites are now consistent: `>=` lower / `<=` upper across pgvector, SurrealDB, and both sqlite_lance channels.
- **Test:** both new tests drive the storage recency window directly (`search_similar_chunks(..., created_before/created_after=bound)`) — the correct layer, since the engine-level recall-filter `compile_python` post-filter drops NULL-`source_timestamp` rows by documented design. The upper-bound test was verified to fail under the old strict `<` and pass with `<=`; the lower-bound test is a defensive contract guard.
- **Performance:** widening admits at most boundary-equal rows; the existing `limit*2` over-fetch absorbs them with no extra ANN cost. No N+1, no unbounded growth, no index impact.
- **Docs:** no public-doc drift (inclusive/exclusive is an internal storage-layer invariant, not a public API contract). No internal identifiers leaked into committed artifacts.
- **Devil's advocate:** merge as-is. The broader `created_at`-only coarse-layer limitation (a chunk whose `source_timestamp` is far before the bound but whose `created_at` is after it can still be coarse-excluded) is honestly documented and correctly scoped out — fixing it would require storing `source_timestamp` in LanceDB (schema + backfill + sync) for an outlier case, and this PR neither introduces nor worsens it.

### Security Audit
- **Verdict: PASS.** Scanned added/modified lines only. No injection (the single interpolation sink is type-safe per M1), no secrets/crypto, no auth/access-control surface, no file/path/SSRF/deserialization sinks, no dependency or lockfile changes. Nothing to auto-fix.

msrdic (approved):

PR #1111, opened 2026-06-12
Fail loud on recall filters over unbacked SurrealDB system keys

filippetrovic:
## Summary
- The SurrealDB skeleton store's `temporal_chunk` table backs only two of the denormalized document system keys with real columns (`occurred_at`, `created_at`). A recall filter on any of the other eight (`source_name`, `source_type`, `source_url`, `source_timestamp`, `external_id`, `content_type`, `source`, `title`) previously compiled to a predicate over a field the SCHEMAFULL table reads as `NONE` — total-false for every row — so recall **silently returned nothing** (and negations could leak excluded rows) instead of erroring.
- Gate the SurrealQL filter compiler's system-key pushdown on the declared `field_mapping` whitelist (mirroring the Weaviate compiler): a system key absent from the declared mapping is routed to `on_unsupported`. The store compiles in raise mode on both the vector and BM25 channels, so it now raises `RecallFilterUnsupportedError` for **any** operator (including `$exists`) on an unbacked key — fail loud, no silent wrong results.
- The store declares exactly the backed keys (`occurred_at`, `created_at`, plus the `metadata → metadata_` mapping) through one shared compile-context helper used at both compile sites. `occurred_at`/`created_at` and `metadata.*` filters are unaffected.
- Register the eight keys as expected-unsupported on the filter-conformance corpus's SurrealDB leg (asserting the raise, not silently skipping), add a drift gate pinning the declared key set to the table schema, and add store-level tests covering both channels plus positive controls.
- No schema columns added; no other backends' compilers touched.

## Test plan
- [x] Unit lane passes (`make test-unit`: 8728 passed, 35 skipped)
- [x] Filter compiler corpus + drift gate + conformance harness unit tests pass
- [x] SurrealDB conformance leg green (`-m filter_conformance`, embedded: 189 passed)
- [x] Store-level both-channel tests pass (vector + BM25 raise on the 8 keys; `occurred_at` range + `metadata.*` positive controls)
- [x] `make lint` + `make format` clean

### DIFF
--- a/docs/engines/skeleton-engine.md
+++ b/docs/engines/skeleton-engine.md
@@ -493,6 +493,21 @@ results = await engine.recall(query, namespace_id=ns_id, mode=SearchMode.HYBRID)
 # Use VectorCypher engine for graph-based queries
 ```
 
+### Deterministic recall filters
+
+The deterministic recall-filter API (the `filter_ast` argument to the backend
+`search`) is supported on the SurrealDB backend, but only over the system keys the
+`temporal_chunk` table backs with a real column: the two datetime keys
+`occurred_at` and `created_at`, plus any `metadata.<path>`. The other eight system
+keys (`source_name`, `source_type`, `source_url`, `source_timestamp`,
+`external_id`, `content_type`, `source`, `title`) are denormalized document fields
+that are **not** columns on `temporal_chunk`. A filter on one of them RAISES
+`RecallFilterUnsupportedError` rather than silently returning nothing: on the
+SCHEMAFULL table the missing field reads as absent, and because SurrealQL's
+absent-compare is total-false (`NONE = x` → `false`) the predicate would drop every
+row. Failing loud surfaces the unsupported key instead of returning a quietly empty
+result set.
+
 ## Configuration
 
 ### Via KhoraConfig
--- a/src/khora/engines/skeleton/backends/surrealdb.py
+++ b/src/khora/engines/skeleton/backends/surrealdb.py
@@ -34,6 +34,7 @@ from khora.telemetry import trace, trace_span
 if TYPE_CHECKING:
     from khora.config import KhoraConfig
     from khora.config.schema import SurrealDBConfig
+    from khora.filter import CompileContext
     from khora.filter.ast import FilterNode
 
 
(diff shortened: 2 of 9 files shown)


filippetrovic:

Seven independent reviewers plus a dedicated security audit examined this change. **Zero findings at any severity (CRITICAL 0 · HIGH 0 · MEDIUM 0 · LOW 0).**

| Domain | Verdict | Notes |
|---|---|---|
| Correctness | ✅ | Gate `path[0] not in self._declared` fires before operator dispatch *and* before consume, so it raises for **every** operator (incl. `$exists`) on the 8 unbacked keys and leaves them unconsumed in split mode; `occurred_at`/`created_at`/`metadata.*` still push. No nesting bypass ($and/$or/$not recurse to the leaf gate). Both read channels enforce it via the shared `_filter_compile_context()` helper. |
| Security | ✅ | The metadata-segment injection guard (`_SAFE_SEGMENT_RE`) is untouched; the new error embeds `path[0]` which is always a closed-set `SYSTEM_KEYS` token, never free user text. No secrets, no ADR-084 secret-typing regressions. The change *strengthens* posture by making a silent wrong-result impossible. |
| Tests | ✅ | All 8 unbacked keys assert the raise on **both** channels (vector + BM25), `$exists` on the 7 string keys, `$eq/$ne/$in/$nin` on the datetime key; positive controls (`occurred_at` range + `metadata.*`) on both channels; drift gate pins the declared set to the schema DDL; resilient error matching (intent substring + key-in-message). |
| Database / schema | ✅ | `_BACKED_SYSTEM_KEYS` = `{occurred_at, created_at}` matches the `temporal_chunk` columns that are also system keys (`source_system`/`author`/`channel`/`tags` are columns but not system keys → correctly excluded). No schema columns added; `metadata → metadata_` remap unchanged. |
| Performance | ✅ | O(1) per-clause dict lookup against a 3-entry mapping built once per compile pass; the conformance helper runs only at test-case construction, never on the query path. |
| Docs | ✅ | The backend doc states the limitation precisely and mechanism-accurately; compiler/store docstrings reflect the declared-whitelist behavior; no stale/contradictory docs. |
| Devil's advocate | ✅ | All call sites of the compiler now declare the backed keys (no caller broken); split mode is unreachable in production (`on_unsupported="raise"` hardcoded) and would be over-return-safe anyway; the conformance `expect_unsupported = … & backends` guard is watertight (no silent pass-through); the pre-existing metadata-quirk exclusion is orthogonal and correctly left untouched. |

### Security Audit
Scanned added/modified lines across the OWASP-aligned category set (injection, authn/z, crypto/secrets, access control, SSRF, deserialization, file/path, redirects, ReDoS, supply chain, boundary validation). **No findings.** No new trust boundary, route, sink, or dependency is introduced — the change is recall-filter compilation logic, tests, and docs. The single regex added (`DEFINE FIELD … (\w+) …`) is a linear-time pattern matched against a static schema constant in a test, not user input. **Verdict: PASS.**

### Overall
The implementation matches the intended design (declared-whitelist gate mirroring the Weaviate compiler; the SurrealDB store declares only its backed system keys), is surgical (no other backends touched, no columns added), comprehensively tested at three levels (unit compiler corpus, store-level both-channel, conformance corpus with the 8 keys registered as expected-unsupported), and the conformance leg is green. **Ready to merge.**

sale87 (approved):

PR #1112, opened 2026-06-12
Add canonical FilterPushdownReport and honest skeleton filter-pushdown reporting

### LINKED ISSUE #1069: recall().engine_info['filter']['pushed_down'] is always False on sqlite_lance even when the filter is fully pushed to SQL (engine=skeleton)
## Summary

`recall(query, filter=...).engine_info["filter"]["pushed_down"]` (the "honest filter pushdown status" added in #1022) reports `True` on the `postgres` backend but **`False` on `sqlite_lance` even when the filter was fully compiled into the SQLite `WHERE` clause**. It's a hardcoded backend check, not a read of what actually happened — so any caller relying on the signal (to decide whether a Python post-filter is needed, or for telemetry) gets the wrong answer on `sqlite_lance`.

The filter *results* are correct on both backends (see note) — this is a reporting bug, not a results bug.

## Environment

- **Khora**: `0.19.x` (reproduced on current `main`)
- **Engine**: `skeleton`
- **Backends**: `sqlite_lance` (mis-reports) vs `postgres` (correct)
- **Python**: `3.13`

## Root cause

`engines/skeleton/engine.py:670` hardcodes the flag to pgvector:

```python
pushed_down = ... and self._backend_type == "pgvector"
```

But `sqlite_lance` runs the filter in split mode and `compile_lance` (JSON1 present, the default) fully pushes the predicate into SQL — `CompiledFilter.consumed_keys == {the AST leaves}`. The adjacent code comment even anticipates the fix ("read `CompiledFilter.consumed_keys`"); the flag was just never wired to the non-pgvector path. The value is surfaced to the user verbatim as the engine's honest filter-handling summary.

## Reproduction

`recall(filter={"metadata.tier": "gold"})` on each backend; full script below. Verified output:

```
filter probed: {'metadata.tier': 'gold'}
pgvector      reports pushed_down = True
sqlite_lance  reports pushed_down = False
compile_lance(JSON1=present) consumed_keys = {'metadata.tier'}
AST leaves                                 = {'metadata.tier'}
  -> sqlite_lance FULLY pushed the filter to SQL: True
  -> yet reports pushed_down = False.   Source: engines/skeleton/engine.py:670
```

## Note (filter correctness itself is fine)

In the same run, the postgres (#1008) and SQLite-json_extract (#1045) filter compilers agreed with each other and with the `compile_python` oracle across **27 filter cases** (string/numeric eq, ranges, nested paths, `$in`/array-contains, `$exists`, null, `$ne`/`$nin`, `$date`, negation, system keys, etc.). So the only defect here is the `pushed_down` flag's value.

## Expected behavior

`engine_info["filter"]["pushed_down"]` reflects whether the predicate was actually pushed to the backend query — `True` on `sqlite_lance` when `compile_lance` consumed the filter keys.

## Actual behavior

`pushed_down` is `False` on every non-pgvector backend regardless of what the compiler did.

## Related (context)

- #1022 — "Report honest filter pushdown status in recall engine_info", the feature this flag belongs to. The flag is dishonest on `sqlite_lance`.

## Suggested fix

Derive `pushed_down` from `CompiledFilter.consumed_keys` (vs the filter AST's leaves) rather than `self._backend_type == "pgvector"`, so it's accurate on every backend whose compiler pushes predicates down.

---

## Full reproduction script (both backends; prints the parity matrix + the pushdown cross-check)

<details>
<summary>probe_skeleton_filter_parity.py</summary>

```python
#!/usr/bin/env python3
"""Cross-backend parity probe for the Skeleton recall-filter (#1012/#1008/#1045).

Drives the PUBLIC facade ``Khora.recall(query, filter=...)`` on
``engine="skeleton"`` against BOTH storage backends with an IDENTICAL adversarial
corpus and an IDENTICAL set of filters, and prints a side-by-side matrix:

    call | pgvector | sqlite_lance | oracle | verdict

The oracle is ``khora.filter.compilers.python.compile_python`` applied to the
decoded chunks — the same in-memory predicate the SQL/Cypher compilers are
documented to be checked against (filter/compilers/python.py module docstring:
"the **oracle** the SQL/Cypher compilers are checked against"). A backend whose
returned chunk-set differs from the oracle is mis-compiling the filter; a
backend-vs-backend difference is the user-facing divergence.

Architecture asymmetry under test (traced, not assumed):
  * pgvector: ``compile_postgres`` runs ``on_unsupported="raise"`` with NO
    compile_python post-filter — the emitted SQL is the SOLE arbiter of the
    row-set (engines/skeleton/backends/pgvector.py:459-465, 611-617).
  * sqlite_lance: ``compile_lance`` runs ``on_unsupported="split"`` AND a
    ``compile_python`` post-filter re-checks the FULL AST against every decoded
    chunk (engines/skeleton/backends/sqlite_lance.py:434, 537-538, 745-800).
So sqlite_lance is oracle-correct by construction; any divergence between the
two backends means the postgres SQL disagrees with the oracle.

Also compares, on the same corpus / fresh namespace per backend:
  1. plain recall (chunk count, #documents, #entities)
  2. recall mode=VECTOR/KEYWORD/HYBRID shape
  3. kb.stats(namespace) counts (chunks/entities/documents)
  4. temporal recall(start_time/end_time)
  5. forget(document_id) then re-recall / stats
  6. engine_info["filter"]["pushed_down"] across backends (#1022)

Run (portable; defaults match the workspace docker stack):

    python probe_skeleton_filter_parity.py

Env overrides:
    KHORA_DATABASE_URL   default postgresql://USER:PASS@HOST:5432/DBNAME
    KHORA_NEO4J_URL      default bolt://USER:PASS@HOST:7687
    KHORA_FILTER_PROBE_LOG  log file path (default: ./output.txt next to script)

Needs: a fresh pg DB with the current schema (skeleton uses skip_graph, neo4j
unused), and aiosqlite + lancedb for the embedded backend. The embedder is
stubbed (deterministic, content-derived) so no LLM/network is touched.
"""

from __future__ import annotations

import asyncio
import math
import os
import sys
import tempfile
from collections.abc import Callable
from contextlib import asynccontextmanager
from datetime import UTC, datetime, timedelta
from pathlib import Path
from typing import Any
from uuid import UUID

from loguru import logger

from khora.config import KhoraConfig
from khora.config.schema import SQLiteLanceConfig
from khora.filter import CompileContext, RecallFilter
from khora.filter.ast import parse_to_ast
from khora.filter.compilers.python import compile_python
from khora.khora import Khora
from khora.query import SearchMode

# --------------------------------------------------------------------------- #
# DSNs (env with workspace defaults).
# --------------------------------------------------------------------------- #

_PG_URL = os.environ.get(
    "KHORA_DATABASE_URL", "postgresql://USER:PASS@HOST:5432/DBNAME"
)
_NEO4J_URL = os.environ.get("KHORA_NEO4J_URL", "bolt://USER:PASS@HOST:7687")

_PG_EMBED_DIM = 1536  # pgvector schema hardcodes Vector(1536) (#925)
_LANCE_EMBED_DIM = 32  # small dim keeps LanceDB index build cheap

_LOG_PATH = os.environ.get(
    "KHORA_FILTER_PROBE_LOG", str(Path(__file__).resolve().parent / "output.txt")
)


# --------------------------------------------------------------------------- #
# Deterministic, content-derived stub embedder. Vocabulary is sized < dim so the
# shared query "alpha bravo charlie" reaches every seeded chunk on both backends
# (the FILTER is the only narrowing force, never retrieval reachability).
# --------------------------------------------------------------------------- #

_KEYWORD_SLOTS = {"alpha": 0, "bravo": 1, "charlie": 2, "delta": 3, "echo": 4}


def _embed_for(text_in: str, dim: int) -> list[float]:
    vec = [0.0] * dim
    vec[dim - 1] = 0.01  # baseline so an all-miss query still has a defined vector
    lower = text_in.lower()
    for kw, slot in _KEYWORD_SLOTS.items():
        if kw in lower:
            vec[slot] = 1.0
    norm = math.sqrt(sum(v * v for v in vec)) or 1.0
    return [v / norm for v in vec]


def _install_stub_embedder(dim: int) -> None:
    import khora.extraction.embedders.litellm as _mod

    async def _embed_batch(self: Any, texts: list[str]) -> list[list[float]]:
        return [_embed_for(t, dim) for t in texts]

    async def _embed(self: Any, text_in: str) -> list[float]:
        return _embed_for(text_in, dim)

    _mod.LiteLLMEmbedder.embed_batch = _embed_batch  # type: ignore[assignment]
    _mod.LiteLLMEmbedder.embed = _embed  # type: ignore[assignment]


# --------------------------------------------------------------------------- #
# Config builders — one per backend.
# --------------------------------------------------------------------------- #


def pgvector_config() -> KhoraConfig:
    return KhoraConfig(database_url=_PG_URL, neo4j_url=_NEO4J_URL)


def sqlite_lance_config(tmpdir: str) -> KhoraConfig:
    cfg = KhoraConfig()
    cfg.storage.backend = "sqlite_lance"
    cfg.storage.sqlite_lance = SQLiteLanceConfig(
        db_path=str(Path(tmpdir) / "khora.db"),
        lance_path=str(Path(tmpdir) / "khora.lance"),
        embedding_dimension=_LANCE_EMBED_DIM,
    )
    cfg.llm.embedding_dimension = _LANCE_EMBED_DIM
    cfg.storage.embedding_dimension = _LANCE_EMBED_DIM
    cfg.pipelines.chunk_size = 4096  # single-chunk docs -> deterministic
    return cfg


# --------------------------------------------------------------------------- #
# Adversarial corpus. Every row shares "alpha bravo charlie" so the vector
# channel returns the whole corpus; metadata / system columns vary so each
# filter bites differently. Date-bearing rows feed the temporal & $date cases.
# --------------------------------------------------------------------------- #

_BASE_TS = datetime(2024, 6, 1, 12, 0, 0, tzinfo=UTC)

# label -> dict(content, metadata, source_name, source_type, occurred_at_offset_days)
CORPUS: dict[str, dict[str, Any]] = {
    "gold_a": {"content": "alpha bravo charlie gold a", "metadata": {"tier": "gold", "score": 10, "labels": {"team": "x"}, "tags": ["red", "blue"], "active": True, "sent_at": "2024-03-15T00:00:00+00:00", "sub": {"k": 1, "j": 2}}, "source_name": "wiki", "source_type": "doc"},
    "gold_b": {"content": "alpha bravo charlie gold b", "metadata": {"tier": "gold", "score": 2, "labels": {"team": "y"}, "tags": ["green"], "active": False, "sent_at": "2024-07-20T00:00:00+00:00", "sub": {"j": 2, "k": 1}}, "source_name": "wiki", "source_type": "doc"},
    "silver": {"content": "alpha bravo charlie silver", "metadata": {"tier": "silver", "score": 5, "sent_at": "not-a-date"}, "source_name": "blog", "source_type": "doc"},
    "numstr": {"content": "alpha bravo charlie numstr", "metadata": {"tier": "bronze", "score": "10"}, "source_name": "blog", "source_type": "post"},
    "absent": {"content": "alpha bravo charlie absent", "metadata": {"other": "x"}, "source_name": "blog", "source_type": "post"},
    "nullt": {"content": "alpha bravo charlie nullt", "metadata": {"tier": None, "score": 7}, "source_name": None, "source_type": "post"},
    "arr": {"content": "alpha bravo charlie arr", "metadata": {"tier": "gold", "score": [1, 2]}, "source_name": "wiki", "source_type": "doc"},
}


async def _seed(kb: Khora, namespace_id: UUID, dim: int) -> dict[str, str]:
    """Remember the corpus; return label -> content (1 chunk per label)."""
    label_to_content: dict[str, str] = {}
    for i, (label, spec) in enumerate(CORPUS.items()):
        ts = _BASE_TS + timedelta(days=i)
        md = dict(spec.get("metadata") or {})
        # The skeleton engine resolves chunk.occurred_at from metadata["occurred_at"]
        # then source_timestamp (engine.py:355-368); the facade remember() exposes
        # no occurred_at kwarg, so drive it via source_timestamp for the temporal
        # window test. (We do NOT inject occurred_at into the user metadata blob so
        # it does not pollute the metadata.* filter matrix.)
        await kb.remember(
            content=spec["content"],
            namespace=namespace_id,
            title=label,
            source_name=spec.get("source_name"),
            source_type=spec.get("source_type"),
            source_timestamp=ts,
            metadata=md,
            entity_types=[],  # Skeleton refuses non-empty typed kwargs (#890)
            relationship_types=[],
        )
        label_to_content[label] = spec["content"]
    return label_to_content


# --------------------------------------------------------------------------- #
# Oracle: compile_python over decoded chunks. We reconstruct a minimal record
# object per returned/seeded chunk so the oracle can evaluate. We evaluate the
# oracle against the FULL seeded corpus (label -> synthetic record) so the
# expected in-scope label set is independent of either backend.
# --------------------------------------------------------------------------- #


class _OracleRecord:
    """Minimal record the compile_python predicate can read (attr + mapping).

    compile_python resolves system keys via getattr / source_document / mapping,
    and metadata via .metadata. We expose exactly those.
    """

    def __init__(self, label: str) -> None:
        spec = CORPUS[label]
        md = dict(spec.get("metadata") or {})
        # The skeleton ingest pipeline injects chunker bookkeeping keys into the
        # STORED metadata blob (start_char / end_char / chunk_index — verified by
        # reading khora_chunks.metadata directly). The oracle must mirror the
        # stored blob, not the user-supplied dict, or the whole-blob $eq case
        # disagrees with reality. Each corpus doc is a single chunk at offset 0.
        content = spec["content"]
        md.update({"start_char": 0, "end_char": len(content), "chunk_index": 0})
        self.metadata = md
        self.source_name = spec.get("source_name")
        self.source_type = spec.get("source_type")
        self.source_url = None
        self.external_id = None
        self.content_type = None
        self.source = None
        self.title = label
        i = list(CORPUS).index(label)
        self.occurred_at = _BASE_TS + timedelta(days=i)
        self.created_at = self.occurred_at
        self.source_timestamp = None
        self.source_document = None


def _oracle_labels(wire: dict[str, Any]) -> set[str]:
    """Labels the filter SHOULD match per the compile_python oracle."""
    rf = RecallFilter.model_validate(wire)
    ast = parse_to_ast(rf)
    pred = compile_python(
        ast, CompileContext(backend_target="khora_chunks", on_unsupported="raise")
    ).predicate
    return {label for label in CORPUS if pred(_OracleRecord(label))}


# --------------------------------------------------------------------------- #
# The filter matrix. Each entry: (name, wire-filter).
# --------------------------------------------------------------------------- #

FILTERS: list[tuple[str, dict[str, Any]]] = [
    ("eq metadata string", {"metadata.tier": "gold"}),
    ("nested metadata path", {"metadata.labels.team": "x"}),
    ("numeric compare gt", {"metadata.score": {"$gt": 5}}),
    ("numeric compare lte", {"metadata.score": {"$lte": 5}}),
    ("$in array-contains", {"metadata.tags": {"$in": ["blue", "purple"]}}),
    ("$in scalar set", {"metadata.tier": {"$in": ["gold", "silver"]}}),
    ("missing-key $exists false", {"metadata.tier": {"$exists": False}}),
    ("present-null {k:null}", {"metadata.tier": None}),
    ("$ne metadata string", {"metadata.tier": {"$ne": "gold"}}),
    ("bool eq true", {"metadata.active": True}),
    ("exact-array eq", {"metadata.tags": ["red", "blue"]}),
    ("system source_name eq", {"source_name": "wiki"}),
    ("system source_type eq", {"source_type": "doc"}),
    ("system source_name $ne", {"source_name": {"$ne": "wiki"}}),
    ("system source_name $in", {"source_name": {"$in": ["wiki", "blog"]}}),
    ("numstr not a number", {"metadata.score": {"$gte": 3}}),
    ("nested $exists true", {"metadata.labels.team": {"$exists": True}}),
    ("$nin metadata", {"metadata.tier": {"$nin": ["gold"]}}),
    # ---- HIGH-RISK: sqlite_lance DEFERS these to compile_python; pgvector
    #      pushes them into SQL with NO post-filter safety net (lance.py:434-455,
    #      pgvector raise-mode). A pg-SQL-vs-oracle mismatch shows up here. ----
    ("$date metadata gte", {"metadata.sent_at": {"$gte": {"$date": "2024-05-01"}}}),
    ("$date metadata lt", {"metadata.sent_at": {"$lt": {"$date": "2024-05-01"}}}),
    ("$date metadata eq", {"metadata.sent_at": {"$date": "2024-03-15"}}),
    ("object_equal dict eq", {"metadata.sub": {"k": 1, "j": 2}}),
    ("object_equal dict ne", {"metadata.sub": {"$ne": {"k": 1, "j": 2}}}),
    # Whole-blob $eq must equal the STORED blob exactly (incl. chunker keys), so
    # target the full stored blob of the `absent` row to prove the positive path.
    ("bare-blob metadata eq", {"metadata": {"other": "x", "start_char": 0, "end_char": 26, "chunk_index": 0}}),
    ("bare-blob no match", {"metadata": {"other": "x"}}),
    ("not(range) negation", {"metadata.score": {"$not": {"$gt": 5}}}),
    ("$in with $date", {"metadata.sent_at": {"$in": [{"$date": "2024-03-15"}, {"$date": "2024-07-20"}]}}),
]


# --------------------------------------------------------------------------- #
# Runners.
# --------------------------------------------------------------------------- #


@asynccontextmanager
async def _open(kb: Khora):
    await kb.connect()
    try:
        yield kb
    finally:
        await kb.disconnect()


async def _filter_labels(
    kb: Khora, namespace_id: UUID, wire: dict[str, Any], content_to_label: dict[str, str]
) -> set[str]:
    res = await kb.recall(
        "alpha bravo charlie",
        namespace=namespace_id,
        limit=50,
        mode=SearchMode.VECTOR,
        filter=wire,
    )
    return {content_to_label[c.content] for c in res.chunks if c.content in content_to_label}


async def run_backend(kb: Khora, dim: int) -> dict[str, Any]:
    """Run the full call matrix on one backend; return a results dict."""
    out: dict[str, Any] = {}
    ns = await kb.create_namespace()
    nsid = ns.namespace_id
    label_to_content = await _seed(kb, nsid, dim)
    content_to_label = {v: k for k, v in label_to_content.items()}

    # 1. plain recall
    plain = await kb.recall("alpha bravo charlie", namespace=nsid, limit=50, mode=SearchMode.VECTOR)
    out["plain_chunks"] = len(plain.chunks)
    out["plain_documents"] = len(plain.documents)
    out["plain_entities"] = len(plain.entities)
    out["plain_labels"] = {content_to_label.get(c.content) for c in plain.chunks}

    # 2. modes
    for mode in (SearchMode.VECTOR, SearchMode.KEYWORD, SearchMode.HYBRID):
        try:
            r = await kb.recall("alpha bravo charlie", namespace=nsid, limit=50, mode=mode)
            out[f"mode_{mode.name}_chunks"] = len(r.chunks)
        except Exception as exc:  # noqa: BLE001
            out[f"mode_{mode.name}_chunks"] = f"ERROR:{type(exc).__name__}:{exc}"

    # 3. stats
    st = await kb.stats(namespace=nsid)
    out["stats_chunks"] = st.chunks
    out["stats_documents"] = st.documents
    out["stats_entities"] = st.entities
    out["stats_repr"] = repr(st)

    # 4. temporal recall (occurred_at window): days 1..3 inclusive of base..base+2
    start = _BASE_TS - timedelta(hours=1)
    end = _BASE_TS + timedelta(days=2, hours=1)  # exclusive upper -> labels idx 0,1,2
    try:
        tr = await kb.recall(
            "alpha bravo charlie", namespace=nsid, limit=50, mode=SearchMode.VECTOR,
            start_time=start, end_time=end,
        )
        out["temporal_labels"] = {content_to_label.get(c.content) for c in tr.chunks}
    except Exception as exc:  # noqa: BLE001
        out["temporal_labels"] = f"ERROR:{type(exc).__name__}:{exc}"

    # 6. engine_info filter pushed_down (probe with one real filter)
    ei = await kb.recall(
        "alpha bravo charlie", namespace=nsid, limit=50, mode=SearchMode.VECTOR,
        filter={"metadata.tier": "gold"},
    )
    out["pushed_down"] = ei.engine_info.get("filter", {}).get("pushed_down")
    out["backend_reported"] = ei.engine_info.get("backend")

    # FILTER MATRIX — runs on the full corpus BEFORE forget() mutates it.
    filt: dict[str, set[str]] = {}
    for name, wire in FILTERS:
        try:
            filt[name] = await _filter_labels(kb, nsid, wire, content_to_label)
        except Exception as exc:  # noqa: BLE001
            filt[name] = f"ERROR:{type(exc).__name__}:{exc}"  # type: ignore[assignment]
    out["filters"] = filt

    # 5. forget: drop the "silver" document, re-stats + re-recall (LAST, after the
    # filter matrix, so deletion does not perturb the per-filter row-sets above).
    silver_doc_ids = {c.document_id for c in plain.chunks if c.content == label_to_content["silver"]}
    if silver_doc_ids:
        await kb.forget(next(iter(silver_doc_ids)), namespace=nsid)
    st2 = await kb.stats(namespace=nsid)
    out["stats_chunks_after_forget"] = st2.chunks
    out["stats_documents_after_forget"] = st2.documents
    after = await kb.recall("alpha bravo charlie", namespace=nsid, limit=50, mode=SearchMode.VECTOR)
    out["labels_after_forget"] = {content_to_label.get(c.content) for c in after.chunks}
    return out


async def main() -> int:
    logger.remove()
    logger.add(sys.stderr, level="INFO", format="{message}")
    logger.add(_LOG_PATH, level="INFO", mode="w", format="{message}")

    def line(s: str = "") -> None:
        logger.info(s)

    # ---- pgvector ----
    line("=" * 100)
    line("SKELETON CROSS-BACKEND FILTER PARITY PROBE  (pgvector vs sqlite_lance)")
    line(f"khora dsn(pg)={_PG_URL}")
    line("=" * 100)

    pg_out: dict[str, Any]
    _install_stub_embedder(_PG_EMBED_DIM)
    pg_kb = Khora(pgvector_config(), engine="skeleton", run_migrations=True)
    async with _open(pg_kb) as kb:
        pg_out = await run_backend(kb, _PG_EMBED_DIM)

    sl_out: dict[str, Any]
    _install_stub_embedder(_LANCE_EMBED_DIM)
    with tempfile.TemporaryDirectory() as td:
        sl_kb = Khora(sqlite_lance_config(td), engine="skeleton", run_migrations=True)
        async with _open(sl_kb) as kb:
            sl_out = await run_backend(kb, _LANCE_EMBED_DIM)

    # ---- report ----
    def fmt(v: Any) -> str:
        if isinstance(v, set):
            return "{" + ",".join(sorted(str(x) for x in v if x is not None)) + "}"
        return str(v)

    line()
    line("SHAPE / STATS / TEMPORAL / FORGET  (basic parity)")
    line("-" * 100)
    line(f"{'call':<32}{'pgvector':<26}{'sqlite_lance':<26}{'same?':<6}")
    line("-" * 100)
    basic_keys = [
        "plain_chunks", "plain_documents", "plain_entities", "plain_labels",
        "mode_VECTOR_chunks", "mode_KEYWORD_chunks", "mode_HYBRID_chunks",
        "stats_chunks", "stats_documents", "stats_entities",
        "temporal_labels", "pushed_down", "backend_reported",
        "stats_chunks_after_forget", "stats_documents_after_forget", "labels_after_forget",
    ]
    # backend_reported legitimately differs (each backend names itself); pushed_down
    # is the divergence under test (#1022) — neither counts as an unexpected DIFF.
    _expected_diff = {"pushed_down", "backend_reported"}
    basic_div = 0
    for k in basic_keys:
        pv, sv = pg_out.get(k), sl_out.get(k)
        same = pv == sv
        if not same and k not in _expected_diff:
            basic_div += 1
        mark = "OK" if same else ("note" if k in _expected_diff else "DIFF")
        line(f"{k:<32}{fmt(pv):<26}{fmt(sv):<26}{mark:<6}")

    line()
    line("FILTER MATRIX  (label-set returned; oracle = compile_python ground truth)")
    line("-" * 118)
    line(f"{'filter':<26}{'pgvector':<22}{'sqlite_lance':<22}{'oracle':<22}{'verdict':<24}")
    line("-" * 118)
    filt_div: list[str] = []
    pg_wrong: list[str] = []
    sl_wrong: list[str] = []
    for name, wire in FILTERS:
        pv = pg_out["filters"][name]
        sv = sl_out["filters"][name]
        oracle = _oracle_labels(wire)
        pv_ok = (isinstance(pv, set) and pv == oracle)
        sv_ok = (isinstance(sv, set) and sv == oracle)
        same = pv == sv
        verdicts = []
        if not same:
            filt_div.append(name)
            verdicts.append("BACKEND-DIFF")
        if not pv_ok:
            pg_wrong.append(name)
            verdicts.append("pg!=oracle")
        if not sv_ok:
            sl_wrong.append(name)
            verdicts.append("sl!=oracle")
        verdict = ",".join(verdicts) if verdicts else "OK"
        line(f"{name:<26}{fmt(pv):<22}{fmt(sv):<22}{fmt(oracle):<22}{verdict:<24}")

    # ---- pushed_down honesty cross-check (the one true divergence) ---- #
    # Recompile the same probe filter with compile_lance under JSON1=present and
    # show that the metadata leaf IS fully consumed (consumed_keys == AST leaves)
    # — i.e. sqlite_lance DID push it down, contradicting its False report.
    from khora.filter import SchemaCapabilities
    from khora.filter.compilers.lance import compile_lance as _cl

    probe_wire = {"metadata.tier": "gold"}
    _ast = parse_to_ast(RecallFilter.model_validate(probe_wire))
    _leaves = {".".join(c.path) for c in _ast.children}  # type: ignore[union-attr]
    _compiled = _cl(
        _ast,
        CompileContext(
            backend_target="khora_chunks", on_unsupported="split",
            schema_capabilities=SchemaCapabilities(sqlite_json1=True),
        ),
    )
    fully_pushed = _compiled.consumed_keys == _leaves

    line()
    line("=" * 100)
    line("SUMMARY")
    line("=" * 100)
    line(f"basic-parity divergences (excl. expected): {basic_div}")
    line(f"filter backend-vs-backend divergences: {len(filt_div)} -> {filt_div}")
    line(f"filters where pgvector != oracle:    {len(pg_wrong)} -> {pg_wrong}")
    line(f"filters where sqlite_lance != oracle: {len(sl_wrong)} -> {sl_wrong}")
    line()
    line("DIVERGENCE: engine_info['filter']['pushed_down']")
    line(f"  filter probed: {probe_wire}")
    line(f"  pgvector      reports pushed_down = {pg_out['pushed_down']}")
    line(f"  sqlite_lance  reports pushed_down = {sl_out['pushed_down']}")
    line(f"  compile_lance(JSON1=present) consumed_keys = {set(_compiled.consumed_keys)}")
    line(f"  AST leaves                                 = {_leaves}")
    line(f"  -> sqlite_lance FULLY pushed the filter to SQL: {fully_pushed}")
    line("  -> yet reports pushed_down=False. Source: engines/skeleton/engine.py:670")
    line("     `... and self._backend_type == \"pgvector\"` hardcodes non-pgvector to False.")
    line()
    line(f"(full log written to {_LOG_PATH})")
    return 0


if __name__ == "__main__":
    raise SystemExit(asyncio.run(main()))
```

</details>

filippetrovic:
## Summary
- Add a canonical, strongly-typed pushdown report for `RecallResult.engine_info["filter"]`. Previously the skeleton engine hard-coded `pushed_down` to `backend_type == "pgvector"`, so `sqlite_lance` always reported `pushed_down=False` even when filters were fully compiled into SQL (the bug in #1069). The flag is now derived from what each backend's compiler actually consumed.
- New public API (frozen Pydantic v2 models), exported from both `khora` and `khora.filter`, mirroring `RecallFilter`'s export treatment:
  - `FilterPushdownReport` — `pushed_down`, `post_filtered`, `pushed_keys`, `post_filtered_keys`, `channels`.
  - `FilterChannelReport` — per-channel `pushed_keys` / `post_filtered_keys`.
- New pure builder `build_filter_report(filter_ast, channel_inputs)` implementing the disposition rules: top-level keys are the intersection across gating channels and partition the constraint leaves; a defensive Python re-check sets `post_filtered` without demoting a pushed key (no-demote); constraint-free / no-filter recalls report `pushed_down=False` with empty key lists and one named empty channel entry.
- Each skeleton backend (pgvector, sqlite_lance, surrealdb, turbopuffer, weaviate) surfaces its plan from the **same compile its search actually ran** — no backend-name checks anywhere in the reporting path. `sqlite_lance` reuses a single authoritative compile (alias-independent, deterministic) for both the WHERE clause and the report.
- The recall facade now passes the engine's report through verbatim instead of flattening it to a `{engine, supported, pushed_down}` dict, so the full per-channel matrix reaches callers. Engines that do not report pushdown simply omit the key.
- Docs (`docs/api-reference.md`) and `CHANGELOG.md` updated; the public schema is pinned by a `model_json_schema()` snapshot test.

Fixes #1069

## Test plan
- [x] Unit tests pass — new `tests/unit/filter/test_report.py` covers intersection/partition semantics, no-demote, empty/no-filter carriers, `pushed_down` derivation, sorted/deterministic keys, dict↔model round-trip, frozen immutability, and a tight `model_json_schema()` snapshot.
- [x] Integration tests pass — `tests/integration/matrix/test_skeleton_sqlite_lance.py` reproduces the original report against a real `sqlite_lance` engine: with JSON1 the metadata+system-key filter now reports `pushed_down=True`; without JSON1 it honestly splits to `post_filtered_keys`. Every emitted report validates via `FilterPushdownReport.model_validate(...)`.
- [x] Full `make test` green (unit + integration), coverage above the configured threshold.
- [x] `make format` / `make lint` clean.


filippetrovic:
## Multi-perspective review — PR #1112

Reviewed across **security, correctness, performance, tests, docs, database/query, and a devil's-advocate** pass. **Overall: APPROVE — no blocking issues; CI is green (25/25).** No CRITICAL or HIGH findings. A few non-blocking MEDIUMs / follow-ups below.

### Clean dimensions
- **Security** — `consumed_keys` is reporting-only; it is never used to build WHERE/Cypher/filter predicates (no injection surface). Models are frozen, no secret-typed fields, no internal identifiers leaked.
- **Correctness** — partition property holds (each constraint leaf in exactly one list, or neither), NO-DEMOTE is correct, and there is no stale-plan leak (every backend resets its plan on every search path). All edge cases handled: no-filter / empty-AND / all-post-filtered / JSON1-absent. No backend-name logic remains in the reporting path. No placeholders.
- **Database / query** — the alias-independent `compile_lance` hoist is sound; the raise-contract (pgvector/surrealdb) and split-mode (weaviate) plans match real backend behavior; no query-path / WHERE-clause changes (plan-stashing only).
- **Docs** — all five fields + disposition rules are documented accurately; CHANGELOG entry is correct; public exports match the models.

### MEDIUM (non-blocking)
- **M1 — Performance** (`filter/report.py` partition loop): the `pushed_keys | post_filtered_keys` union is recomputed once per leaf. Pre-computing it once per channel removes N redundant unions. Impact is sub-microsecond per recall — a nice-to-have, not a blocker.
- **M2 — Test coverage gap**: the builder, the sqlite_lance integration path (JSON1 present *and* absent), and the facade pass-through are airtight. The **raise-contract** path on pgvector/surrealdb and the **split-mode report** on weaviate are not yet exercised by real-engine integration tests (only sqlite_lance is). A follow-up adding those would guard the "plan is set before the raise guard" invariant.
- **M3 — Facade shape change**: the facade now passes the engine's report through verbatim instead of the previous `{engine, supported, pushed_down}` wrapper. The dropped `supported` / facade-level `engine` keys were verified to be read nowhere in the codebase (only asserted by tests, which were updated) — so this is a safe, intentional canonicalization, not an accidental break. Worth a one-line CHANGELOG note that the wrapper was removed, for downstream awareness.

### Design notes (devil's advocate — context, not blockers)
- The **N-channel intersection builder** is intentionally general for a currently single-channel engine; it is shared infrastructure the sibling engine-adoption work builds on, so the generality is amortized rather than speculative.
- The **`_last_filter_plan` instance-state carrier** is a deliberate tradeoff to avoid changing the shared `search()` protocol signature across all backends; reviewers confirmed no stale-state leak. A future refactor could return the plan from the search call if that protocol is revisited.

### Security audit
No OWASP / injection / crypto / dependency concerns for this reporting-only change — `consumed_keys` is derived from already-validated filter-AST paths and never re-enters query construction. Nothing to auto-fix.

**Recommendation: APPROVE.** Merge-ready; M1–M3 are good follow-ups, not blockers.

filippetrovic:
Reviewed against #1069 and the new `report.py` semantics. The model/builder/test structure is right and the sqlite_lance repro now reports honestly. Requesting changes on two items, plus smaller fixes.

**1. (blocking) `_last_filter_plan` races under concurrent recalls.** The plan is written at the top of `_search_inner`/`search` and read back by the engine after `await temporal_store.search(...)` returns. Two overlapping recalls on the same store interleave: recall A stashes its plan, awaits its DB round-trips, recall B overwrites the attribute, and A's report is built from B's plan — the report can describe a different call's filter. Since the whole point of #1069 is an honest report, this needs a per-call handoff: return the `ChannelPlan` alongside the search results (or build the report inside the search call) instead of mutable instance state. Applies to all five backends.

**2. (blocking) sqlite_lance: the report compile is a third compile, not the one the WHERE ran.** `_search_inner` now runs a reporting-only `compile_lance`, while the actual WHERE fragments still come from the two per-pass compiles in `_build_ast_clause(alias=None/"c")`. The inputs are identical today, but the two `CompileContext(...)` constructions live in different methods and can drift independently (e.g. a capability flag added to one and not the other), at which point the report silently stops describing the executed WHERE. Make them share one source: extract a `_filter_compile_ctx(alias)` factory used by both `_build_ast_clause` and the plan, or capture `consumed_keys` from the pass compile itself. The "ONE authoritative compile … single source of truth for what BOTH passes push down" comment and the PR-body claim should be corrected either way — nothing else consumes that compile as written.

**3. (non-blocking) Contradictory partition docs.** `FilterPushdownReport`'s docstring (and `docs/api-reference.md`) say the two top-level lists *partition* the constraint leaves, while `build_filter_report`'s docstring and the gated-by-no-channel test establish that an ungated leaf lands in *neither* list. Reconcile (e.g. "partition the *gated* constraint leaves") or define the ungated disposition explicitly — unreachable for single-channel skeleton plans, but it matters for multi-channel engines.

**4. (non-blocking) `khora.py` comment accuracy.** The pass-through comment says chronicle writes the canonical `report.model_dump(mode="json")` — chronicle currently writes a hand-rolled `{pushed_down, post_filtered}` dict. Also, with the old facade synthesis removed, engines that don't emit a report no longer carry an `engine_info["filter"]` key at all — the body mentions it, but it deserves a CHANGELOG line since callers may key on presence.

**5. (non-blocking) Backend test coverage.** Report tests cover sqlite_lance (integration) and pgvector (stubs). The weaviate plan (`filter_leaf_keys(ast) - consumed_keys` + defensive recheck) is the most intricate of the five and has no test; surrealdb/turbopuffer are untested too.

**6. (nit)** `_filter_report`'s `assert model_validate(info) == model_validate(info)` is tautological — the validate call alone is the guard. Several test comments cite "decision 1/2/3/4", which doesn't resolve to anything in the repo or #1069 — reference the `docs/api-reference.md` section instead.

Heads-up: #1111 touches the same surrealdb `_search_inner` filter block — whichever lands second needs a rebase there; the raise-contract reading of `consumed_keys` remains valid under #1111's `field_mapping` change.

filippetrovic:
Thanks for the thorough review — all six addressed in `fbdf2de5`.

**1. (blocking) `_last_filter_plan` race — fixed.** Replaced the per-store instance attribute with a per-call `filter_plan_out` sink: the engine creates a fresh `list[ChannelPlan]` per recall, passes it into `search(...)`, and each backend appends exactly one plan built from the same compile its search ran. No shared mutable state, so two overlapping recalls on a shared store can't clobber each other's report. I made `filter_plan_out` an optional kwarg (default `None`) rather than changing the return type, so the out-of-scope `vectorcypher` caller and the ~15 backend `search()` tests are untouched, and it doesn't collide with #1111's surrealdb `_search_inner` work. `_last_filter_plan` is gone from all five backends.

**2. (blocking) sqlite_lance third-compile drift — fixed.** Extracted a single `_filter_compile_ctx(alias)` factory used by both `_build_ast_clause` (the per-pass WHERE) and the report compile in `_search_inner`, so a capability flag can't be added to one path and silently missed by the other. Corrected the misleading "ONE authoritative compile … single source of truth for BOTH passes" comment — it now states the plan and WHERE compiles share one context factory (deterministic + alias-independent `consumed_keys` ⇒ the report describes the executed WHERE). pgvector/surrealdb were already drift-free: they derive the plan from the same raise-mode compile that builds the WHERE.

**3. (non-blocking) partition docs — reconciled.** `FilterPushdownReport`'s docstring and `docs/api-reference.md` now say the lists *partition the **gated** constraint leaves*; a leaf no channel gates lands in neither (unreachable for single-channel skeleton, defined for multi-channel engines). The `model_json_schema()` snapshot was updated to match.

**4. (non-blocking) `khora.py` comment + CHANGELOG — fixed.** The pass-through comment no longer claims chronicle writes the canonical model (it writes its own `{pushed_down, post_filtered}` dict; canonical adoption is a follow-up). Added a `### Changed` CHANGELOG line: the facade no longer wraps the report, and engines that don't report pushdown now **omit** the `"filter"` key — callers must treat it as optional.

**5. (non-blocking) backend test coverage — added.** New plan tests for weaviate (the intricate split: date keys pushed, metadata → `post_filtered_keys`, `defensive_recheck`, NO-DEMOTE) and turbopuffer (empty plan when constraint-free; sink left untouched when the constraint raises). surrealdb's plan is byte-identical to pgvector's raise-mode path (`ChannelPlan(pushed_keys=consumed_keys)`), exercised by the engine stub test and the `conformance (surrealdb)` leg; pgvector's is covered by the same stub test.

**6. (nit) — fixed.** `_filter_report` now asserts `model_validate(info).model_dump(mode="json") == info` (a real round-trip identity, not the tautology). The "decision N" comments now reference the `engine_info["filter"]` rules in `docs/api-reference.md`.

Local `make test` is green (unit + integration, coverage above the 77% gate); ruff + ty clean.

filippetrovic:
Re-reviewed at b6b5d8c9. Both blockers resolved: the per-call `filter_plan_out` sink eliminates the shared-state race cleanly across all five backends (fresh list per recall, append-after-compile, raise-before-append verified — the sink stays empty when the compile raises), and `_filter_compile_ctx` makes the sqlite_lance report compile and the per-pass WHERE compiles structurally undriftable. Items 3–6 all addressed (partition docs reconciled to "gated constraint leaves", khora.py comment + CHANGELOG presence note, weaviate/turbopuffer plan tests, real round-trip assert, dangling comment references reworded). Good to merge once `test-integration` is rerun — the single failure there is the Neo4j keepalive timing flake (`test_keepalive_fires_pings_over_a_couple_intervals`, defunct-connection ping), untouched by this diff. One trivial leftover for whenever: `test_skeleton_sqlite_lance.py:625`'s comment still references the removed `_last_filter_plan` attribute.

msrdic (approved):

PR #1164, opened 2026-06-12
Denormalize document string keys in the sqlite_lance conformance runner

filippetrovic:
## Summary
- The recall-filter conformance corpus pruned every case that reads the *value* of one of the seven string document keys (`source_type`, `source_name`, `source_url`, `external_id`, `content_type`, `source`, `title`) off the `sqlite_lance` leg. The pruning existed because the conformance seeder builds chunk rows from the core `Chunk` model, which carries no string keys (they live on `Document`), so the denormalized chunk columns landed `NULL` and diverged from the python oracle.
- This was a seeder gap, not a backend gap: production `sqlite_lance` copies those keys onto each chunk row (the compiled predicate targets the denormalized chunk columns, no documents join). The conformance runner now mirrors production — and the existing postgres conformance runner — by fetching the seven keys off the parent `documents` row in one batched query and stamping them onto each `TemporalChunk` before insert.
- `_NO_STRING_KEY_BACKENDS` now contains only `weaviate` (whose collection schema genuinely lacks the columns). `sqlite_lance` joins `_SOURCE_TYPE_DEFAULTED_BACKENDS`: an unset `source_type` denormalizes to the `Document` default `"library"` while the oracle sees the key absent, so a `source_type` value leaf over a seed with an unset row is still pruned for that specific reason.
- Test-runner only: no production code changes. The override lives entirely in the conformance runner subclass.

Collect-count on the `sqlite_lance` leg rises from **122 → 187** (+65 string-key value cases now exercised).

## Test plan
- [x] `sqlite_lance` filter-conformance leg green — 187 passed (was 122 collected before)
- [x] `python` oracle filter-conformance leg unaffected — 208 passed
- [x] Filter unit guards green — 698 passed (`test_conformance_harness`, `test_conformance_catalog`, `test_system_keys_coverage`)
- [ ] Full CI matrix (including docker integration/e2e legs) green

### DIFF
--- a/src/khora/filter/conformance.py
+++ b/src/khora/filter/conformance.py
@@ -759,27 +759,25 @@ _OP_BACKENDS: frozenset[str] = _INMEM_BACKENDS | _LIVE_BACKENDS
 # The live backends that do NOT carry the seven string document keys
 # (``source_name`` / ``source_url`` / ``external_id`` / ``content_type`` /
 # ``source`` / ``source_type`` / ``title``) on the queryable row, so a predicate that
-# READS a string-key value is empty / divergent there and is pruned: the sqlite_lance
-# runner seeds through the core ``Chunk`` (the string keys live on the Document, not
-# the core chunk, so they land NULL), and the weaviate ``KhoraChunk`` collection
-# schema has no string-key columns at all. The surrealdb and cypher runners seed the
-# seven string keys verbatim onto the queryable row, and the postgres runner now
-# denormalizes them off the parent ``documents`` row onto each chunk (mirroring
-# production), so all three KEEP string-key value cases. (Postgres still prunes a
-# ``source_type`` value leaf when a seed record leaves ``source_type`` unset — see
-# :data:`_SOURCE_TYPE_DEFAULTED_BACKENDS`.)
-_NO_STRING_KEY_BACKENDS: frozenset[str] = frozenset({"sqlite_lance", "weaviate"})
+# READS a string-key value is empty / divergent there and is pruned: the weaviate
+# ``KhoraChunk`` collection schema has no string-key columns at all. The surrealdb and
+# cypher runners seed the seven string keys verbatim onto the queryable row, and the
+# postgres and sqlite_lance runners now denormalize them off the parent ``documents``
+# row onto each chunk (mirroring production), so all four KEEP string-key value cases.
+# (Postgres and sqlite_lance still prune a ``source_type`` value leaf when a seed
+# record leaves ``source_type`` unset — see :data:`_SOURCE_TYPE_DEFAULTED_BACKENDS`.)
+_NO_STRING_KEY_BACKENDS: frozenset[str] = frozenset({"weaviate"})
 
 # The live backends that DEFAULT an unset ``source_type`` to ``"library"``: the
-# postgres runner denormalizes the parent Document's ``source_type`` onto the chunk,
-# and the ``Document`` dataclass defaults ``source_type`` to ``"library"`` (so does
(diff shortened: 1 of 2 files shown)


filippetrovic:

7 independent reviewers (security, performance, correctness, test, database, docs, devil's advocate) plus a dedicated security audit. **Unanimous: APPROVE.** No CRITICAL / HIGH / MEDIUM findings.

Test-runner-only change; production code untouched. The conformance seeder now denormalizes the seven string document keys off the parent `documents` row onto each seeded chunk (one batched SELECT), mirroring production and the existing postgres conformance runner. The `sqlite_lance` leg now exercises string-key value predicates instead of pruning them (collect count 122 → 187, +65 cases) and the full suite stays green.

### Findings by severity

**CRITICAL / HIGH / MEDIUM:** none.

**LOW (optional, non-blocking)**
- **L1 [Devil's Advocate]** `_conformance_lance.py` `_fetch_document_keys` — the raw `f"SELECT {columns} FROM documents WHERE id IN ({placeholders})"` is slightly less self-evidently safe than the postgres twin's SQLAlchemy-core form. Mitigated: `columns` is static (built from the hardcoded `_DOC_STRING_KEYS` tuple), placeholders are `?` markers, ids bind positionally, and the `# noqa: S608` documents it. Raw aiosqlite is the file's existing house style (matches `run_live`).
- **L2 [Devil's Advocate]** `_fetch_document_keys` iterates `_DOC_STRING_KEYS` twice (SQL build + row unpack). Considered and rejected as a change: the dict-of-dicts shape is required for the `**doc_keys` unpack and keeps parity with the postgres twin — simplifying would lose that consistency.
- **L3 [Devil's Advocate]** Optional future-proofing: a unit test asserting the `uuid_to_text` ↔ `UUID(...)` round-trip between the SQLAlchemy write path and the aiosqlite read path. Currently covered transitively by the green conformance suite (both paths use the same encoding on the same file).

**INFO**
- **[Security audit / Injection]** `_conformance_lance.py:_fetch_document_keys` — SQL is assembled with an f-string, but the interpolated `columns` is a static module constant and all values bind positionally; the `document_ids` are internally-generated seed UUIDs, not untrusted input. No injection surface. No new endpoints, secrets, crypto, deserialization, file/redirect/SSRF sinks, or dependency changes.

### Per-domain notes
- **Security:** clean — static column list + positional binds; `# noqa: S608` justified.
- **Performance:** clean — single batched, deduped SELECT per chunk-batch; no N+1; bounded `IN`.
- **Correctness:** approved — NULL-vs-absent semantics align oracle and chunk row for `$eq`/`$ne`/`$in`/`$exists`; UUID hex round-trip correct; the two frozenset membership changes are logically consistent with `_backends_for_filter`.
- **Test:** approved — the +65 cases exercise real string-key value predicates; coverage guards unchanged; `external_id` correctly contributes 0 new cases (duplicate-`external_id` UNIQUE-constraint prune); source_type-default prune still excludes the right cases.
- **Database:** approved — hex-no-dashes UUID encoding matches on-disk format; named row access backed by `row_factory = aiosqlite.Row`; read-after-write consistency holds (documents committed before the chunk batch on the same file); zero schema/migration/ORM/pool impact.
- **Docs:** clean — all six comment/docstring sites accurate; no stale "sqlite_lance prunes string keys" phrasing; correct-and-unchanged sqlite_lance mentions (split-backend, created_at-stamp, seed-case Document path, duplicate-external_id) left intact.
- **Devil's Advocate:** MERGE — right diagnosis, right scope (weaviate stays pruned for its genuine schema gap), no over-engineering.

**Security audit verdict: PASS.**

probablyrobot (approved):

PR #1179, opened 2026-06-12
Emit canonical FilterPushdownReport from the Chronicle engine

filippetrovic:
## Summary
- Chronicle's `RecallResult.engine_info["filter"]` now emits the canonical `FilterPushdownReport` shape (top-level `pushed_down` / `post_filtered`, the `pushed_keys` / `post_filtered_keys` constraint-leaf partition, and a per-channel `channels` breakdown with a single `chunks` channel) instead of a hand-rolled two-field `{pushed_down, post_filtered}` dict.
- The report is built by the shared `build_filter_report` builder, fed from the engine's existing `plan_chronicle_filter` execution input — no backend-name checks and no parallel "reporting" compile that could drift from what the recall actually ran.
- **Behavior change:** `pushed_down` now uses ALL-leaves semantics (previously "any key pushed"). Chronicle is partial-pushdown by design — only the `source_timestamp` date bound folds into the recency window, while the full predicate is always re-checked in memory — so a multi-predicate filter now reports `pushed_down=false` with the pushed vs post-filtered keys split out, exposing the partial-pushdown truth the old boolean hid. The pushed date key stays in `pushed_keys` even though the post-filter re-checks it (no-demote), and `post_filtered` is `true` whenever a constraint-bearing filter is supplied.
- The carrier is written on every recall, including the no-filter and empty-filter cases (canonical empty carrier with the one `chunks` channel).
- CHANGELOG updated under the existing unreleased section, calling out both the report-shape change and the any→all `pushed_down` semantic correction (this is the released metric of record for callers).

## Test plan
- [x] Unit tests pass (`tests/unit/test_khora.py`, `tests/unit/filter/test_report.py`)
- [x] Recall tests pass (`tests/recall/test_filter_chronicle.py`, `tests/recall/test_chronicle_filter_composition.py`)
- [x] Embedded (sqlite_lance) Chronicle integration tests pass (`tests/integration/test_chronicle_filter_embedded.py`)
- [x] Every emitted report validated against the public `FilterPushdownReport` model on every path (no-filter, empty-filter, constraint-bearing)
- [ ] Postgres-backed integration tests (`tests/integration/test_chronicle_filter_pgvector.py`) — run in CI (no local Docker)


msrdic (approved):

filippetrovic:

**Overall: APPROVE** ✅

Six independent reviewers (security, performance, correctness, tests, docs, devil's advocate) examined the diff. The change is surgical (engine emission site + trimmed comment + 4 test files + CHANGELOG), reuses the canonical builder without duplication, and is correct across every disposition.

### Findings

**CRITICAL / HIGH:** none — no defects.

#### MEDIUM
- **[Tests]** `tests/recall/test_chronicle_filter_composition.py` — coverage did not include a `source_timestamp` predicate nested under `$or` / `$not` (which must NOT fold into the recency window and must land in `post_filtered_keys`). **Addressed** in follow-up commit `f28554e2`: two new tests assert `pushed_keys == []` / `post_filtered_keys` for both `$or` and `$not`, locking in the invariant the all-leaves `pushed_down` semantics depends on.

#### LOW
- **[Behavior/Docs]** `pushed_down` changes from "any leaf pushed" to "all leaves pushed". This is the intended correction and is documented in the CHANGELOG; `f28554e2` adds an explicit caller-migration note pointing `pushed_down` readers at a non-empty `pushed_keys`.
- **[Design]** A single `chunks` channel models a 4-retrieval-channel engine. This is correct: Chronicle applies one unified post-filter across the chunk candidate set (the four channels are read counters, not independent gating channels), and the canonical single-channel convention names the channel for that unified set. The inline comment documents the partial-pushdown rationale.

### Per-domain results
- **Security** — No findings. `engine_info["filter"]` carries only leaf-PATH key strings (`".".join(clause.path)`), never filter values or query text; no data exposure, no secrets, no new dependencies, no injection surface.
- **Performance** — No issues. `build_filter_report` runs once per recall (outside per-chunk/per-entity loops); `filter_pushed_keys` is reused from the already-computed plan (no second compile pass); `model_dump(mode="json")` on a small frozen model is cheap.
- **Correctness** — No defects. All four dispositions (partial / date-only no-demote / cross-axis post-filter-only / empty carrier) verified against the builder; scope and `None`-guards correct; NO-DEMOTE honored.
- **Tests** — Strong. `FilterPushdownReport.model_validate(...)` runs on every emitted path; canonical key strings are real (not guessed); dispositions covered across mocked / embedded sqlite_lance / pgvector / public-facade layers, plus the new nested-operator cases.
- **Docs** — CHANGELOG entry accurate and correctly placed under the unreleased `Changed` section; no documentation drift (the canonical schema reference already exists).

### Security Audit
No additional findings beyond the security review. The diff is a pure result-shape change with no I/O, auth, cryptography, secret handling, or new dependencies — the OWASP Top 10 surface is untouched. MEDIUM/LOW auto-fix: none required.

_All actionable review items were addressed in commit `f28554e2`._

PR #1180, opened 2026-06-12
Store denormalized document fields on the Weaviate KhoraChunk collection

filippetrovic:
## Summary
- Persist the eight denormalized document fields (`source_type`, `source_name`, `source_url`, `external_id`, `content_type`, `source`, `title`, `source_timestamp`) on the `KhoraChunk` collection. Previously these were never stored, so recall filters over them were silently wrong — a positive predicate matched **zero rows** and a negation was a no-op, even though the projection displayed the values hydrated from the relational row.
- The seven string keys are stored as **filterable-only** `TEXT` properties (`index_filterable=True`, `index_searchable=False` so provenance strings don't pollute BM25 hybrid-search scoring); `source_timestamp` is a `DATE`. They are defined once in a lazily-built property tuple (import-safe — `weaviate-client` is an optional dependency) shared by both collection creation and an **idempotent schema-reconciliation** path that adds any missing property to a pre-existing collection.
- `_chunk_to_properties` now writes the eight fields and `_object_to_chunk` reads them back, so the post-filter (the correctness safety net) can address them. Server-side pushdown stays date-only.
- **Conformance fidelity fix:** the filter-conformance weaviate leg now evaluates the post-filter against the **read-back object surface** rather than the declared seed mapping, so the leg verifies the store actually round-trips every filtered field (it previously stayed green even when the store dropped a field at write/readback). The obsolete string-key prune is removed now that the keys are carried on every backend.

## Test plan
- [x] Unit tests pass — `tests/unit/engines/skeleton/backends/test_weaviate_async.py` (47 passed; adds round-trip + idempotent-reconciliation coverage, asserting `index_searchable=False` on the TEXT properties)
- [x] Filter-conformance python leg passes — `KHORA_CONFORMANCE_BACKEND=python … -m filter_conformance` (208 passed)
- [x] Conformance unit guards pass — harness / catalog / system-keys-coverage (no changes; identical pre-existing env-only failures with the optional extra absent)
- [x] `ruff check` + `ruff format --check` clean on the changed files
- [ ] CI runs the weaviate Docker leg with a fresh container (exercises the new create path + the new properties end-to-end)

**Conformance case-count delta (weaviate leg):** 122 → 189 (**+67**). The added cases are string-key VALUE predicates that previously pruned the weaviate leg and now run against it — green only because the keys are seeded, stored, and read back correctly.

### DIFF
--- a/src/khora/engines/skeleton/backends/weaviate.py
+++ b/src/khora/engines/skeleton/backends/weaviate.py
@@ -66,6 +66,40 @@ COLLECTION_NAME = "KhoraChunk"
 # survivor set can still reach ``limit`` after the post-filter prunes.
 _FILTER_OVERFETCH = 4
 
+# The seven denormalized document string keys carried on every ``KhoraChunk`` object
+# (the eighth, ``source_timestamp``, is a DATE). Stored as filterable-only TEXT
+# (``index_searchable=False`` keeps them out of BM25 hybrid search) so the post-filter
+# can read them back; they do NOT push down (pushdown stays date-only).
+_DENORM_TEXT_KEYS: tuple[str, ...] = (
+    "source_type",
+    "source_name",
+    "source_url",
+    "external_id",
+    "content_type",
+    "source",
+    "title",
+)
+
+
+def _denorm_properties() -> tuple[Any, ...]:
+    """Build the eight denormalized document properties (single source: create + reconcile).
+
+    Lazily constructed because ``weaviate.classes.config`` is an optional import
+    (the backend module imports fine without ``weaviate-client`` installed). The
+    seven string keys are TEXT with ``index_searchable=False`` — a REQUIRED guard
+    against BM25 hybrid-search pollution — and ``index_filterable=True`` so the
+    post-filter can address them; ``source_timestamp`` is a DATE.
+    """
+    from weaviate.classes.config import DataType, Property
+
+    return (
+        *(
+            Property(name=key, data_type=DataType.TEXT, index_filterable=True, index_searchable=False)
+            for key in _DENORM_TEXT_KEYS
+        ),
+        Property(name="source_timestamp", data_type=DataType.DATE),
+    )
+
 
 @dataclass(frozen=True)
 class WeaviateBackendConfig:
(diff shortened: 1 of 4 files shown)


msrdic (approved):

filippetrovic:

A reviewer team (security, performance, correctness, tests, database/schema, docs, devil's advocate) plus a dedicated security audit reviewed this PR. **Overall: APPROVE** — no CRITICAL or HIGH findings; one MEDIUM advisory and a couple of LOW follow-ups, none blocking.

### Per-domain verdicts
| Domain | Verdict | Notes |
|---|---|---|
| Security | ✅ CLEAN | No new Pydantic secret-typed fields (`source_url` is a pre-existing non-credential provenance field); `metadata_json` decode is guarded; no injection — Property construction uses named params. |
| Performance | ✅ APPROVE | `index_searchable=False` correctly keeps provenance strings out of the BM25 corpus; reconciliation runs only on the cold `connect()`/existing-collection path (≤8 iterations), not per-write/per-query. |
| Correctness | ✅ CLEAN | `occurred_at`→`source_timestamp` coalesce matches the oracle record shape; write/read property mapping is symmetric incl. None/absent; reconciliation idempotent on `prop.name`; no dangling refs to the deleted prune constant; no placeholders. |
| Tests | ✅ APPROVE | Round-trip (populated + absent/None) and reconciliation (asserts `index_searchable=False`; zero adds on second connect) covered. The conformance leg now post-filters against the **read-back object surface**, genuinely catching a store that drops a field. |
| Database/Schema | ✅ APPROVE | TEXT + DATE typing correct for Weaviate v4; idempotent schema reconciliation is backwards-compatible (adds only missing props, no data loss); collection-config ops are tenant-independent. |
| Docs | ✅ CLEAN | All docstring/comment rewrites match code behaviour; stale "returns ZERO rows" limitation removed; no orphaned references. |
| Devil's Advocate | ☑️ Approve w/ notes | Storing the fields (vs read-time hydration) is the right call — avoids per-result joins. Two advisory items below. |

### MEDIUM
- **M1 — [Design]** `connect()` reconciliation — Adding missing schema properties to a pre-existing collection is defensive but borderline-YAGNI: CI always starts a fresh container (hits the clean-create path), so only dev volumes that weren't wiped benefit. The logic is idempotent and cheap, and it's covered by a unit test, so keeping it is reasonable; the alternative is to drop it and document a `docker compose down -v` step when upgrading a local volume. **Non-blocking — recommend keep as-is (self-healing, low cost).**

### LOW
- **L1 — [Performance/Ops]** Storage footprint of the seven string columns + one date across every chunk on every shard is unquantified. Not a concern at current scale; worth a follow-up metric if the collection grows large or this becomes a hot path.
- **L2 — [Style]** The `_reconcile_noop_collection()` test helper could be inlined, but it clarifies intent — optional.

### Security Audit
**Verdict: PASS.** Scanned the added/modified lines across all 16 categories. No injection (no query/template/shell construction; Property args are named constants), no secrets/weak crypto, no unsafe deserialization (the one `json.loads` is on controlled test-harness data and does not execute code), no SSRF/redirect/file/auth surface, and no dependency or lockfile changes. The two keyword hits were false positives (`DES` inside "decodes"/"encodes" in comments).

### Assessment
**APPROVE.** Implementation matches the spec precisely, scope is disciplined (storage backend + test harness only), and verification is strong (unit round-trip + reconciliation, the python conformance leg, and a +67-case expansion on the previously-pruned leg). The MEDIUM is an advisory design preference, not a defect.

probablyrobot (approved):

PR #1181, opened 2026-06-12
Emit canonical filter-pushdown report in VectorCypher engine_info

filippetrovic:
## Summary

The VectorCypher engine previously emitted **no** `engine_info["filter"]` entry, so callers had no way to know how a recall filter was handled across its multiple retrieval channels. It now surfaces the canonical `FilterPushdownReport` on **every** recall, built from each channel's *actual* compile (never a backend-name guess or a parallel "reporting" compile that could drift):

- **vector** and **bm25** report the `khora_chunks` constraint leaves they pushed into the shared Postgres `WHERE`.
- the **graph** (Cypher) channel reports the system-key slice it pushed down plus the residual `metadata.*` leaves it re-checks in memory.
- the **recency** channel reports its full-predicate in-memory post-filter.

All four channels gate fusion, so the shared builder's per-leaf partition keeps the report honest: a leaf that any gating channel re-checks in memory lands in the top-level `post_filtered_keys` even when the SQL channels pushed it (NO-DEMOTE applies only to a defensive full-predicate re-check). Only channels that actually executed under a filter appear in `channels`; a degraded channel (e.g. a graph fallback) is excluded rather than claiming a disposition it never performed. A no-filter recall yields an all-false report.

The per-channel plans are carried on a private result-metadata key that the engine pops **before** the public `engine_info` is assembled, so the internal carrier never leaks into the public payload. The shared report model and builder are reused unchanged.

A separate, pre-existing filter-correctness gap in the typed-entity-recent fast path (it ignores the recall filter entirely) is surfaced honestly by this report (`channels: {}`) and tracked separately — it is intentionally out of scope here.

## Test plan

- [x] Unit tests pass — 21 engine-level tests pinning the canonical payload shape: hybrid mixed-filter per-channel partition, graph-only mode, no-filter carrier, NO-DEMOTE (both gating-demote and defensive-no-demote), recency-present-only-when-fired, graph-fallback exclusion, typed-entity honest empty report, and edge cases (empty filter, system-only, metadata-only). Every emitted report is asserted to round-trip through `FilterPushdownReport.model_validate`.
- [x] Embedded-stack tests pass — 2 `sqlite_lance` matrix tests proving the report is honest end-to-end on the embedded compilers.
- [x] `make lint` and `make format` clean.
- [ ] CI: full integration + e2e + conformance matrix.


filippetrovic:

Reviewed by a six-reviewer panel (security, performance, correctness, tests, docs, devil's advocate) plus a dedicated diff-scoped security audit. **Overall assessment: APPROVE.** No blocking findings.

### CRITICAL
_None._

### HIGH
_None._ (One candidate HIGH was raised and adjudicated down — see H-adjudicated below.)

### MEDIUM
_None._

### LOW
- **L1 [Devil's Advocate] — recency channel "gating" narrative.** The inline comment could spell out *why* the recency channel gates: it contributes **new** recency-ordered candidates that the vector search never surfaced and can only enforce the filter on them via an in-memory post-filter (no SQL pushdown), so those leaves are genuinely post-filtered for the result set. This is distinct from a *defensive* re-check of already-pushed candidates (which would set `defensive_recheck=True` and not demote). The existing comment is accurate; this is a clarity nit only. No code change required.
- **L2 [Devil's Advocate] — test file size.** ~1k-line test module was scrutinized for padding; each test maps to a distinct channel combo / edge case / invariant. Judged proportionate, not redundant. No action.

### Adjudicated (raised as HIGH, resolved to non-blocking)
- **H-adjudicated [Devil's Advocate] — "recency is defensive, not gating."** Investigated against the fusion code: recency-contributed chunks merge into the vector pool and enter RRF (`merged_in = [rc for rc in recent_chunks if rc[0] not in existing_ids]`), so they can survive to the output. Because recency can only enforce the filter in memory, reporting those leaves as `post_filtered` is the **honest** outcome (the filter was not fully pushed to the DB). The proposed alternative (`defensive_recheck=True`, report `pushed_down=True`) would make the report *claim* full DB pushdown that did not happen. The correctness reviewer independently confirmed the classification. **Not a defect; not a blocker.**

### Per-dimension results
- **Security — no findings.** The new `engine_info["filter"]` payload surfaces only dotted constraint-leaf *key paths* (e.g. `metadata.tier`), never filter values, query text, chunk content, or secrets. The private `_filter_channel_plans` carrier is `pop`-ed before the public payload is assembled (guarded by a test).
- **Performance — no findings.** No redundant compile (graph reuses the captured cypher compile; vector/bm25 reuse the real sinks). Per-call dict/sink allocation is O(1) per recall; `model_dump`/`filter_leaf_keys` are O(1), not per-chunk. No-filter overhead negligible.
- **Correctness — no findings.** Gating/defensive classification matches the fusion path; only executed channels appear (graph excluded on fallback, recency only when fired, bm25 only when the fulltext path ran under a filter); pop-before-spread verified; report emitted on every recall; no placeholders.
- **Tests — no findings.** Every emitted report is validated via `FilterPushdownReport.model_validate`; NO-DEMOTE covered both ways; recency-presence ×4; graph-fallback honesty; typed-entity honest empty report; session fan-out single-capture; 5-way concurrent-recall isolation; embedded tests exercise the real lance compiler.
- **Docs — no findings.** CHANGELOG entry correct; `engine_info["filter"]` already documented in the API reference (the PR implements the already-published contract); comments accurate; no internal identifiers leaked.
- **Devil's Advocate — see L1/L2 and H-adjudicated.** Recommends merge as-is. Confirmed the per-call threading is the simplest viable approach for a multi-channel engine (a single shared sink can't work across the concurrent channel tasks).

## Security Audit

Diff-scoped audit across the 16 categories (injection, auth, crypto/secrets, access control, SSRF, deserialization, file/path, redirects, mass assignment, ReDoS, supply chain, …). The change adds no entry points and routes no untrusted input to any dangerous sink; no new dependencies; no file/network/deserialization operations.

- **INFO** — the payload surfaces user-supplied metadata *key names* (paths) back to the same caller in the result object (not a span/metric → no cardinality concern). Acceptable.

**Verdict: PASS** — no CRITICAL/HIGH/MEDIUM/LOW security findings.

---

A pre-existing, separately-tracked filter-correctness gap on the typed-entity-recent fast path (it does not enforce the recall filter) is **reported honestly** by this change (empty `channels`) and is intentionally out of scope here.

probablyrobot (approved):

Issue #569, opened 2026-05-13
Temporal Phase C: entity-anchored fast path for 'latest <typed-entity>' queries


probablyrobot:
Follow-on to #567 (Phase A — scoring fixes) and #568 (Phase B — ingestion contract). Phase C addresses the **entry-point bug**: for queries like "latest action items," `_vector_search_entities` matches conceptual entities ("Meeting", "Action Item Tracker") instead of the actual content, so the graph traversal starts from the wrong roots. No scoring fix downstream can rescue this.

## Problem

For VectorCypher RECENCY queries:

1. Aho-Corasick detects `TemporalCategory.RECENCY`.
2. Router boosts complexity to MODERATE; retrieval goes through `_vectorcypher_retrieve`.
3. **Entry entities found by `_vector_search_entities`** — vector search on the query embedding against typed-entity nodes.
4. For "action items," the closest entities semantically are conceptual ones ("Action Item Tracker"), NOT the actual action-item content. The traversal starts wrong.
5. `temporal_sort=True` on the Cypher side correctly does `ORDER BY c.occurred_at DESC` — but only over chunks anchored to the **wrong roots**.

The right shape for this query class is a typed-entity fast path:

```cypher
MATCH (a:Entity {entity_type:'ACTION_ITEM'})
WHERE a.namespace_id = $ns
  AND a.created_at >= datetime() - duration({days:14})
RETURN a ORDER BY a.created_at DESC LIMIT 20
```

Sub-50ms, no embedding, no fusion, no rerank.

### C1 — Decouple `prefer_current` from `temporal_sort`

Today in `RetrievalParams`, `prefer_current` follows `temporal_sort`. For ORDINAL ("which came first") `prefer_current=True` is wrong — you want historical entities, not currently-valid ones. Add a dedicated `prefer_current: bool = False` field:

- `STATE_QUERY` / `RECENCY` / `CHANGE` → `prefer_current=True`
- `ORDINAL` / `EXPLICIT` / `AGGREGATE` → `prefer_current=False`

Update the two call sites in `retriever.py` (~lines 669, 733) to use `_tp.prefer_current` instead of `_tp.temporal_sort`. Mechanical, ~10 lines.

### C2 — First-class `ACTION_ITEM` extraction

[...]

PR #1183, opened 2026-06-12
Gate VectorCypher typed-entity-recent fast path when filter_ast is present

filippetrovic:
## Summary

- The typed-entity-recent fast path (`_typed_entity_recent_retrieve`) runs a single typed-entity/status-only Cypher query and builds result chunks straight from the rows. It received the caller's filter but never applied it, so a recall carrying a metadata or document filter was silently returned **unfiltered**.
- The fast-path Cypher cannot enforce caller filters: chunk metadata is stored as a serialized JSON property on the graph node (not queryable as distinct columns), and the path hydrates chunks with `occurred_at` only.
- Fix: gate the fast path in the dispatch on `filter_ast is None`. Unfiltered typed-entity-recent recalls keep the single-Cypher fast path (recency-ordered, unchanged). Filtered recalls fall through to the full `_vectorcypher_retrieve` path, which enforces the filter on every channel (vector, BM25, graph) and emits the canonical per-channel filter-pushdown report.
- The fast path's `filter_ast` parameter and its internal fallback threading are left intact so the static filter-enforcement audit gate stays satisfied.
- Correctness-over-latency tradeoff: a filtered typed-entity-recent recall loses the single-Cypher recency optimization. This combination is rare; a follow-up could push system-key leaves into the fast-path Cypher to recover it.

## Test plan

- [x] Unit suite passes locally (`make test-unit`: 8845 passed, 35 skipped), including:
  - [x] Filtered typed-entity-recent recall skips the fast path and forwards the same filter to the full path (no `typed_entity_fast_path` marker on the result).
  - [x] Unfiltered typed-entity-recent recall still takes the fast path (latency regression guard).
  - [x] `mode=VECTOR/KEYWORD/GRAPH` dispatch unaffected by the new condition.
  - [x] Filter-report suite: a filtered typed-entity-recent recall emits a real non-empty per-channel report (validates against the report model); the unfiltered fast path carries the all-false no-filter report.
  - [x] Filter-enforcement audit gate passes (filter threading intact).
- [x] Lint/format clean (`make format`, `make lint`).
- [ ] Integration tests — validated in CI (Docker is not available in the authoring sandbox).

## Related issues

- #1181 — canonical filter-pushdown report in VectorCypher `engine_info["filter"]` (foundation for this gate).
- #569 — original typed-entity-recent fast path.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@ Format: versions match git tags (`git tag vX.Y.Z`). Versions before 0.5.1 were i
 ### Fixed
 
 - **VectorCypher recency channel now fires again on the pgvector stack** — the temporal store now implements `search_recent_chunks` (the channel had been silently dead since the temporal-store rewiring). The pgvector temporal store sorts `khora_chunks` by `COALESCE(occurred_at, source_timestamp, created_at) DESC` (served by a new `ix_khora_chunks_ns_recency` expression index), preserves the embedding so the caller's cosine relevance gate runs, and backends without the capability return the protocol default `[]`. Embedded (`sqlite_lance`) support is tracked in #1182.
+- **VectorCypher typed-entity-recent fast path no longer bypasses the recall filter; filtered recalls route to the full retrieval path.**
 - **Honest filter-pushdown reporting on the skeleton engine** (#1069): `recall(filter=...).engine_info["filter"]["pushed_down"]` was derived from a hardcoded `backend == "pgvector"` check, so it reported `False` on `sqlite_lance` even when the compiler fully pushed the predicate into the SQLite `WHERE` clause. The flag is now derived from what each channel's compiler actually consumed (`pushed_keys`) versus what it re-checked in memory (`post_filtered_keys`), so `pushed_down` is accurate on every backend whose compiler pushes predicates down. A defensive full-predicate re-check (the `sqlite_lance` path) sets `post_filtered=True` without demoting a fully-pushed leaf (NO-DEMOTE).
 
 ### Changed
--- a/src/khora/engines/vectorcypher/retriever.py
+++ b/src/khora/engines/vectorcypher/retriever.py
@@ -812,7 +812,19 @@ class VectorCypherRetriever:
             force_graph = mode == SearchMode.GRAPH
 
             # Step 3: Vector search for entry points
-            if not force_simple and not force_graph and routing.complexity == QueryComplexity.TYPED_ENTITY_RECENT:
+            if (
+                not force_simple
+                and not force_graph
+                and routing.complexity == QueryComplexity.TYPED_ENTITY_RECENT
+                and filter_ast is None
+            ):
+                # Gate (filter_ast is None): the fast-path Cypher cannot
+                # enforce caller filters — chunk metadata is a serialized
+                # JSON property on the graph node, not queryable columns — so
+                # a filtered recall would silently return unfiltered chunks.
+                # Filtered recalls take the full _vectorcypher_retrieve path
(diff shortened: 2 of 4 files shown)


filippetrovic:
## 🧑‍⚖️ Multi-perspective team review

Six independent reviewers (Security, Performance, Correctness, Test, Docs, Devil's Advocate) plus a dedicated security audit examined this PR. Scope: gate the typed-entity-recent fast path on `filter_ast is None` so filtered recalls route to the full enforcing path.

**Overall assessment: ✅ APPROVE** — correct, well-tested, secure, and accurately documented. One non-blocking design note (a possible latency follow-up) is recorded below.

### CRITICAL
- None.

### HIGH
- None.

### MEDIUM
- **[Devil's Advocate / Performance]** `src/khora/engines/vectorcypher/retriever.py:768` — Filtered typed-entity-recent recalls now drop from the single-Cypher fast path to the multi-channel full path. The "this combination is rare" assumption is asserted, not measured. This is the deliberate correctness-over-latency tradeoff (the fast-path Cypher cannot enforce metadata filters — chunk metadata is a serialized JSON property on the graph node), and the common unfiltered path is unchanged. **Recommendation (non-blocking):** track a follow-up issue to push system-key leaves (e.g. `source_name`, document keys) into the fast-path Cypher WHERE so the latency optimization survives for the pushable subset. Not required for this fix.

### LOW
- **[Devil's Advocate]** No latency benchmark accompanies the regression path; the guard tests assert routing/behavior, not timing. Quantifying the regression in the follow-up would validate the "rare combination" framing.
- **[Correctness — informational]** Two independent analyses disagreed on whether an explicitly-empty filter (`{}`) lowers to `None` vs. an empty `FilterNode`. Either way the gate is correct: a real filter always routes to the enforcing path, and an empty/no-op filter is harmless on either branch. No action needed.

### Reviewer coverage
- **Security:** No findings. Confirms the data-exposure hole is closed; gate uses identity (`is None`), not falsy coercion, so an empty `FilterNode` cannot bypass it; all four internal fallbacks thread `filter_ast`; namespace scoping intact; no injection (filter is a parsed/validated AST).
- **Performance:** No findings. Regression correctly scoped to filtered recalls only; unfiltered fast path identical; no N+1 / extra round-trips; reuses existing MODERATE/COMPLEX machinery.
- **Correctness:** No findings. Boolean gate placement and semantics correct; filtered intent preserved; fast-path `filter_ast` param + fallback threading unchanged (audit gate satisfied); no placeholders.
- **Test:** No findings. Tests exercise the gate through `retrieve()` (filtered → full path + same `filter_ast` + no fast-path marker; unfiltered → fast path; VECTOR/KEYWORD/GRAPH dispatch unaffected); the flipped report test asserts a real non-empty per-channel report via `FilterPushdownReport.model_validate`; no-filter carrier retained.
- **Docs:** No findings. CHANGELOG entry and the updated gate/docstring/no-filter-carrier comments are accurate; no internal identifiers leaked; only public GitHub refs (#569, #1181).
- **Devil's Advocate:** Merge, with the optional follow-up above.

### Security Audit
Scanned added/modified lines across all 16 OWASP-aligned categories (injection, authn/authz, crypto/secrets, access control, deserialization, file/path, SSRF, redirects, mass assignment, ReDoS, supply chain, boundary validation). The diff adds only a boolean dispatch condition, comments/docstrings, test mocks/assertions, and a CHANGELOG line — **no security-relevant sinks**.
- **Verdict: PASS** (no CRITICAL/HIGH/MEDIUM/LOW findings).

---
*Net: merge-ready. The single MEDIUM is a forward-looking latency optimization, not a defect in this change.*

probablyrobot (approved):

PR #1184, opened 2026-06-12
Normalize source_type to "library" in the filter-conformance corpus

filippetrovic:
## Summary
- An absent `source_type` is unreachable in production: the public `remember()` parameter, the `Document` dataclass field, and the `documents.source_type` column all default to `"library"` (the column is `NOT NULL`). The filter-conformance corpus, however, modeled an impossible absent state, which spuriously pruned two value-reading equality cases from the `postgres` and `sqlite_lance` legs.
- Default `SeedRecord.source_type` to a non-null `"library"`, byte-identical to the production `Document` contract. This makes an absent `source_type` unauthorable anywhere in the corpus and flows automatically into the oracle record mapping and every live seeder.
- With that default in place, the seed-aware `_SOURCE_TYPE_DEFAULTED_BACKENDS` prune is dead (its `any(rec.source_type is None ...)` guard is now constant-false), so it is removed along with its branch in `_backends_for_filter`. The two affected `$eq` cases now run on the `postgres` and `sqlite_lance` legs again.
- Simplify the seeder's `source_type` coalesce (the else-branch is now dead), drop the now-orphaned `Op` import, and update the related docstrings/comments to match the new rationale.
- **Test-harness only** — no production code, no `expected_ids`, and no oracle survivor sets change.

## Collection-count delta (vs. base)
Verified by loading the base branch's harness standalone and computing each version's per-case backend set:
- `postgres`: 187 → 189 (**+2**)
- `sqlite_lance`: 187 → 189 (**+2**)
- `python` / `chronicle` / `cypher` / `surrealdb` / `weaviate`: ±0
- Total cases unchanged (208). The two cases that gain the `postgres`/`sqlite_lance` legs are the implicit-AND and explicit-AND logic cases; their `expected_ids` are unchanged (`"library"` still fails `$eq "connection"`).

## Test plan
- [x] `ruff format` + `ruff check` clean on the touched files
- [x] `make lint` (ruff + ty typecheck) exits 0; no new diagnostics on the touched files
- [x] Python oracle leg green: `KHORA_CONFORMANCE_BACKEND=python pytest tests/integration/matrix/test_filter_conformance.py -m filter_conformance` → 208 passed
- [x] Grep gate: the removed constant has no remaining references in `src/` or `tests/`
- [ ] Full `filter-conformance` CI matrix (postgres / sqlite_lance / surrealdb / weaviate / cypher legs) green

### DIFF
--- a/src/khora/filter/conformance.py
+++ b/src/khora/filter/conformance.py
@@ -69,7 +69,7 @@ from khora.filter.compilers.weaviate import compile_weaviate
 # is the one production CompileContext builder; ``run_chronicle_filter`` is the
 # Chronicle date-bound-pushdown + full-AST post-filter applied to in-memory records.
 from khora.filter.execute import build_compile_context, iter_leaf_clauses, run_chronicle_filter
-from khora.filter.model import SYSTEM_KEYS, Op
+from khora.filter.model import SYSTEM_KEYS
 from khora.storage.coordinator import StorageCoordinator
 
 __all__ = [
@@ -156,8 +156,10 @@ class SeedRecord:
     vector channel returns the whole set and the filter is the only narrowing
     force). The remaining fields populate the eight denormalized document system
     keys, the three date columns, and the chunk ``metadata`` blob — the surface a
-    filter can address. A field left ``None`` / empty is simply absent from that
-    record (the missing-value semantics each compiler must agree on).
+    filter can address. A nullable field left ``None`` / empty is simply absent from
+    that record (the missing-value semantics each compiler must agree on);
+    ``source_type`` is non-null (it defaults to ``"library"``, mirroring the
+    ``Document`` dataclass default), so it is always present on every backend.
     """
 
     id: str
@@ -166,7 +168,7 @@ class SeedRecord:
     source_timestamp: datetime | None = None
     occurred_at: datetime | None = None
     created_at: datetime | None = None
-    source_type: str | None = None
+    source_type: str = "library"
     source_name: str | None = None
     source_url: str | None = None
     external_id: str | None = None
@@ -688,7 +690,7 @@ async def seed_case(coord: StorageCoordinator, case: ConformanceCase) -> dict[st
         doc = Document(
             namespace_id=namespace_id,
             content=record.content,
-            source_type=record.source_type if record.source_type is not None else "library",
+            source_type=record.source_type,
             source_name=record.source_name,
             source_url=record.source_url,
             external_id=record.external_id,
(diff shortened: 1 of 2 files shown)


filippetrovic:
## Consolidated PR Review

Multi-perspective review by a 6-reviewer panel (Security, Performance, Correctness, Test, Docs, Devil's Advocate) plus a dedicated security audit. Overall assessment: **APPROVE** — no blocking findings.

### CRITICAL
- None

### HIGH
- None

### MEDIUM
- None

### LOW
- **L1 [Devil's Advocate / Docs]** `src/khora/filter/conformance.py` (`_record_mapping`) — Optional clarity polish: a one-line comment noting that `source_type` is now non-null by construction (so it is always present in the oracle mapping, unlike the other string keys) would help future readers. **Non-blocking** and intentionally out of scope for this change, which deliberately keeps `_record_mapping` untouched (the new field default flows through it automatically). Recorded for visibility only.

### Findings by reviewer
- **Security** — No findings. Test-harness-only change; no injection, data-exposure, secret, or dependency surface.
- **Performance** — No findings. The change removes a branch from a per-case harness computation (`_backends_for_filter`); no runtime-path impact.
- **Correctness** — All checks pass: the deleted `_SOURCE_TYPE_DEFAULTED_BACKENDS` guard is now constant-false (the field can no longer be `None`); `elif root == "created_at"` → `if root == "created_at"` is semantically identical; the removed `Op` import is genuinely orphaned (0 remaining references, module imports cleanly); no `expected_ids` changed; no placeholders.
- **Test** — Coverage preserved. The two `$and` logic cases are correctly un-pruned onto the `postgres` and `sqlite_lance` legs (collection delta **postgres +2, sqlite_lance +2**, every other leg ±0, total 208); both keep their existing `expected_ids`; no assertion weakened.
- **Docs** — All rewritten docstrings/comments are accurate, reference no removed constant, and leak no internal identifiers.
- **Devil's Advocate** — Endorses the approach: normalizing the field default and removing the dead prune entirely (rather than leaving a guard) is the right scope; correctly leaves the sibling `created_at` prune untouched as separate scope. Only the optional L1 polish above.

## Security Audit
Diff-scoped audit across the OWASP-aligned categories. **Verdict: PASS.** No CRITICAL/HIGH/MEDIUM/LOW findings.
- **INFO** — A pattern scan flagged the substring "des" inside the word "encodes" in a docstring comment (the `DES`-cipher heuristic); this is a false positive, not a cryptographic finding.
- No new dependencies or manifest/lockfile changes; no network, file, deserialization, or injection sinks introduced.

### Overall: APPROVE
Test-harness-only change, zero production-code impact, fully verified. Merge once the `filter-conformance` CI matrix is green.

probablyrobot (approved):

PR #1185, opened 2026-06-12
Add cross-engine filter-report invariant conformance gate

filippetrovic:
## Summary
- Adds a CI gate asserting every engine's `RecallResult.engine_info["filter"]` obeys the engine-independent `FilterPushdownReport` invariants: the pushed / post-filtered key lists **totally and disjointly partition** the filter's constraint leaves; `pushed_down` derives from the list partition (**not** the `post_filtered` flag, which a defensive in-memory re-check legitimately sets even at full pushdown); every channel addresses only real leaves with **sorted** key lists; and a no-filter / `filter={}` recall is the canonical empty carrier.
- Runs **once per engine configuration**. Because `engine_info["filter"]` is emitted only on a *real recall*, the gate rides the end-to-end recall matrix (one engine config per leg) rather than the compile-level conformance matrix — that matrix never runs a recall and emits no report.
- Adds a **hermetic drift guard** tying the gate's covered lane set to the e2e matrix definition (and the backend-token table), so a new engine config cannot ship without invariant coverage. Includes falsifiability suites proving both the invariant helper and the drift guard reject malformed input.
- Pure test additions + one CI workflow edit + one CHANGELOG line — **no production-code changes**.

## Test plan
- [x] New drift-guard + coverage-gate + report-anchor unit tests pass (51 passed)
- [x] In-process e2e lanes pass with the gate active against real recalls: `vc_embedded` (32), `skeleton_sqlite_lance` (32), `skeleton_surrealdb` (28)
- [ ] Live-store e2e lanes (`vc_full`, `skeleton_pgvector`, `skeleton_weaviate`, `chronicle`) — validated in CI (require Postgres / Neo4j / Weaviate)
- [x] `ruff format` / `ruff check` clean; typecheck clean


filippetrovic:
## Multi-perspective review — cross-engine filter-report invariant gate

Six independent reviewers (correctness, test quality, security, performance, docs, devil's advocate) plus a dedicated security audit examined this PR. **Overall: APPROVE.** No CRITICAL/HIGH findings. The change is purely additive test code + one CI-workflow edit + one CHANGELOG line — no production-code, dependency, or lockfile changes.

### Verdicts by domain
- **Correctness — CLEAN.** All invariants (a–g) traced against `build_filter_report`: list-form `pushed_down`, the leafless early-return branch, and the one-directional `post_filtered` flag (no-demote) all match the builder. No conflation of the `post_filtered` bool with `post_filtered_keys`. No placeholders/stubs.
- **Test quality — CLEAN.** Genuine teeth: a valid control + 8 malformed-report cases each targeting a *specific* invariant, plus two falsifiability tests proving the drift guard fails RED on a synthetic extra token/leg. Anti-vacuity guard on the case corpus; per-lane node-id selection verified (no cross-lane bleed).
- **Security — CLEAN.** No secrets/credentials, no CI-yaml injection (matrix values are literals), no unsafe fixtures.
- **Performance — CLEAN.** pytest dedupes the all-leg path append (no double execution); invariant checks are O(leaf count); fixtures are function-scoped with fresh namespaces.
- **Docs — CLEAN.** CHANGELOG line is accurate, user-facing, and explains why the gate rides the e2e (real-recall) matrix. Comments accurate.
- **Security audit — CLEAN.** Test/CI/CHANGELOG only; no production code, no new dependencies, no crypto/secret/shell-exec surface.

### Non-blocking notes (LOW / informational)
- **L1** The drift guard couples three definitions (backend-token set ⇔ e2e lane map ⇔ CI matrix). This coupling is *intentional* — it is the mechanically-enforced "same schema, every engine config" guarantee this change exists to provide. A future PR that restructures the e2e matrix must update the mapping in lockstep — that is the guard firing by design, not a false positive.
- **L2** The explicit module path is appended to all seven lane legs' `select`, redundant for the six `-k`-filtered legs (the dir-glob already collects it). Intentional for static-analysis visibility and documented in an in-file comment; pytest dedupes, so no double run.
- **L3** Invariant (e) (channel-fold consistency) overlaps invariant (c) (partition). Kept as defense-in-depth: (c) checks the top-level partition; (e) checks the top level is correctly *derived from* the per-channel breakdown — a builder bug could pass (c) yet fail (e).
- **L4** The gate runs real recalls per lane (costlier than a unit test against hand-built dicts) — but real recalls are the only path that emits the report, so the fidelity is the point. Tradeoff is documented in the CHANGELOG.
- **L5** A pre-existing stale count comment on one lane in the e2e backend map is *outside this diff* (the harness change is purely additive) — left untouched to keep the PR surgical; worth a separate cleanup.

### Assessment: APPROVE
No changes required. The remaining gate is CI: the embedded lanes pass locally; the live-store lanes (Postgres/Neo4j/Weaviate) are validated in the e2e workflow.

probablyrobot (approved):

Issue #1182, opened 2026-06-12
Implement search_recent_chunks on the sqlite_lance temporal store (recency channel on the embedded stack)


filippetrovic:
## Context

The VectorCypher recency channel (`temporal_recency_channel_enabled`) sources its candidate pool via `TemporalVectorStore.search_recent_chunks`. An upcoming PR adds that method to the protocol as an optional capability (default: empty list) and implements it on the pgvector temporal store, so the channel fires on Postgres stacks.

On the embedded sqlite_lance stack the protocol default keeps the channel inert. A naive SQLite-only implementation would not be enough: SQLite rows are read with `embedding=None` (vectors live only in LanceDB), and the channel's cosine relevance floor (`temporal_query_relevance_floor`) drops embedding-less chunks — so it would silently contribute nothing.

## What's needed

- A recency-ordered SQLite read: `ORDER BY COALESCE(occurred_at, source_timestamp, created_at) DESC LIMIT n` (the `ix_khora_chunks_occurred` index already exists).
- Join the resulting chunk ids into the LanceDB `khora_chunks_vec` table to attach embeddings.
- Return `(Chunk, None)` tuples with `embedding` populated, matching the protocol shape.

## Acceptance

- With the channel enabled, `_recency_channel_chunks` produces candidates on the sqlite_lance stack (embedding-bearing, recency-ordered).
- An embedded-stack test proves the channel contributes candidates and respects the relevance floor.

filippetrovic:
I'll fix same issue for postgres path, what are your thoughts for sqllite_lance @probablyrobot? Do we want/need this on embedded stack?

PR #1186, opened 2026-06-13
Implement search_recent_chunks on the pgvector temporal store

filippetrovic:
## Summary
- The VectorCypher recency channel was dead on every stack. The channel calls `search_recent_chunks` on the wired temporal store, but no temporal-store backend implemented it (it existed only on the storage-layer pgvector backend, which is never the temporal store), so the channel always early-returned `[]` and its config knobs were no-ops. The channel is default-OFF, so this changes no default behavior.
- Add `search_recent_chunks` as a non-abstract optional-capability method on the `TemporalVectorStore` protocol (default body `return []`), mirroring the existing `search_fulltext` pattern. Backends without a recency-capable table contribute nothing and never appear in the filter report.
- Implement it on the pgvector temporal store: a pure recency sort over `COALESCE(occurred_at, source_timestamp, created_at)` (event-time → producer-time → ingest-time), with an optional `created_after` lower bound on the same axis and a `limit`. The query selects the full row so embeddings survive onto the returned chunks — the channel drops embedding-less chunks before the cosine gate, so losing the embedding would silently re-kill the channel.
- Add a matching B-tree expression index `ix_khora_chunks_ns_recency (namespace_id, (COALESCE(occurred_at, source_timestamp, created_at)) DESC)`, created idempotently on connect, whose expression is byte-identical to the `ORDER BY` so Postgres serves the sort from the index.
- Rewrite stale comments at the call site and in the channel docstring that claimed the channel read the relational chunks table and that the method was absent on every wired store. No control-flow change — the capability guard, embedding-drop guard, in-memory post-filter, and channel-plan recording are untouched.
- Embedded (sqlite_lance) support is deferred and tracked in #1182.

## Test plan
- [x] Unit tests pass (recency-channel suite extended with a protocol-default-store case: a real protocol implementer relying on the default `[]` yields no channel plan)
- [ ] Integration tests pass (new pgvector suite: store-level recency ordering on the COALESCE axis, embeddings populated on returned chunks, `limit`/`created_after` honored, namespace isolation; plus end-to-end recall asserting the recency channel appears in the filter report with all caller-filter leaves post-filtered) — run in CI against provisioned Postgres/Neo4j
- [x] `ruff check` + `ruff format` clean; full local unit run shows no regressions vs. base (pre-existing failures are confined to optional-extra backends not installed in this combo)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ Format: versions match git tags (`git tag vX.Y.Z`). Versions before 0.5.1 were i
 
 ### Fixed
 
+- **VectorCypher recency channel now fires again on the pgvector stack** — the temporal store now implements `search_recent_chunks` (the channel had been silently dead since the temporal-store rewiring). The pgvector temporal store sorts `khora_chunks` by `COALESCE(occurred_at, source_timestamp, created_at) DESC` (served by a new `ix_khora_chunks_ns_recency` expression index), preserves the embedding so the caller's cosine relevance gate runs, and backends without the capability return the protocol default `[]`. Embedded (`sqlite_lance`) support is tracked in #1182.
 - **Honest filter-pushdown reporting on the skeleton engine** (#1069): `recall(filter=...).engine_info["filter"]["pushed_down"]` was derived from a hardcoded `backend == "pgvector"` check, so it reported `False` on `sqlite_lance` even when the compiler fully pushed the predicate into the SQLite `WHERE` clause. The flag is now derived from what each channel's compiler actually consumed (`pushed_keys`) versus what it re-checked in memory (`post_filtered_keys`), so `pushed_down` is accurate on every backend whose compiler pushes predicates down. A defensive full-predicate re-check (the `sqlite_lance` path) sets `post_filtered=True` without demoting a fully-pushed leaf (NO-DEMOTE).
 
 ### Changed
--- a/src/khora/engines/skeleton/backends/__init__.py
+++ b/src/khora/engines/skeleton/backends/__init__.py
@@ -236,6 +236,32 @@ class TemporalVectorStore(Protocol):
         """
         return []
 
+    # Optional capability — backends that store ingested chunk content in
+    # their own table override this to expose a pure-recency channel for the
+    # VectorCypher recency path. Default returns an empty list so backends
+    # without a recency-capable table contribute nothing to the channel.
+    async def search_recent_chunks(
+        self,
+        namespace_id: UUID,
+        limit: int,
+        *,
+        created_after: datetime | None = None,
+    ) -> list[tuple[Chunk, float | None]]:
+        """Return the ``limit`` most-recent chunks in the namespace.
+
+        Pure recency sort with no semantic gate — the caller (the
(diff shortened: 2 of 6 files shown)


filippetrovic:
## Multi-Perspective Review Summary

**Overall assessment: APPROVE.** Seven independent reviewers (Security, Performance, Correctness, Test, Docs, Database, Devil's Advocate) plus a dedicated security audit. No CRITICAL or HIGH findings. One MEDIUM (declined with rationale) and one LOW (documented, out of scope).

### CRITICAL
- None.

### HIGH
- None.

### MEDIUM
- **[Correctness]** `search_recent_chunks(..., created_after=...)` — the parameter bounds the recency axis `COALESCE(occurred_at, source_timestamp, created_at)` (event-time first), so the name `created_after` is arguably broader than `created_at`. **Resolution: keep the name.** It mirrors the storage-layer `search_recent_chunks` sibling signature (parity contract) and the established call site, and both docstrings explicitly state the bound applies to the same COALESCE recency axis. Renaming would diverge from the sibling and the call site for no behavioral gain. Behavior is correct.

### LOW
- **[Design]** The temporal store sorts on a 3-way `COALESCE(occurred_at, source_timestamp, created_at)`, whereas the relational storage-layer backend uses a 2-way `COALESCE(source_timestamp, created_at)`. This is intentional and correct: the temporal `khora_chunks` table carries `occurred_at` (event-time) while the relational `chunks` corpus does not. The divergence is documented in code comments and the CHANGELOG. A future unification (backfilling `occurred_at` into the relational table) is out of scope here.

### Security Audit
- **PASS.** The new `search_recent_chunks` query is fully parameterized via SQLAlchemy Core (`namespace_id`, `created_after`, `limit` all bound; no string interpolation). Namespace isolation is enforced on every call and covered by a dedicated multi-tenant test. The new expression index uses static DDL. No new dependencies, no secrets, no auth/access-control surface. Two test-fixture items were inspected and confirmed benign (a throwaway dev-container Neo4j password matching the existing test convention; a schema-reset `DROP TYPE` whose identifier is sourced from the `pg_type` catalog, not untrusted input).

### Reviewer corroboration
- **Performance:** index expression byte-matches the `ORDER BY`, so Postgres serves the sort from the index; `LIMIT` always applied; full-row select is required (embeddings feed the cosine gate) and bounded; no N+1.
- **Test:** unit protocol-default case proves a capability-less backend records no channel plan; integration tests are non-vacuous — the channel-level test only passes when the recency channel genuinely fires and post-filters the caller filter leaf (all gates engaged: recency rows, cosine floor, filter survival, plan recorded).
- **Database:** index is idempotent (`IF NOT EXISTS`), single-statement, created at connect-time per the file's existing DDL pattern; correct leading-equality + sort composite shape.
- **Docs:** all three rewritten comment/docstring sites are accurate; no stale "relational chunks table" / "absent on every wired store" claims survive; CHANGELOG clean.
- **Devil's Advocate:** protocol-boundary placement, opt-in revival of a silently-broken channel, and the embedded-stack deferral (tracked in #1182) are all justified; no over-engineering.

probablyrobot (approved):

PR #1216, opened 2026-06-14
Drop absent-created_at seeds from filter-conformance corpus

filippetrovic:
## Summary
- `created_at` is always-present in production (NOT NULL column + default + writer stamping), so the filter-conformance corpus no longer models an absent `created_at`.
- The F-OP `created_at` `-5` seed row now sits at the miss instant instead of NULL; the `$lt` / `$lte` / `$in` expected sets pick it up accordingly. The other seven date operators are unchanged — the Python oracle leg recomputes every survivor set and would catch any mis-edit.
- Deletes the `_CREATED_AT_STAMP_BACKENDS` prune. It dropped every `created_at` case from the postgres / sqlite_lance / weaviate legs because those legs stamp `now()` on an absent value while the oracle saw the declared seed as absent — a divergence that can no longer occur once the corpus never leaves `created_at` unset.
- Replaces the `created_at` branch in `_backends_for_filter` with a mandatory guard that raises at corpus-construction time if any seed record leaves `created_at` unset (an unreachable production state), so a future hand-authored regression fails loudly instead of silently diverging.
- Removes the now-dead `_strip_stamped_created_at` helper and its call site from the neo4j conformance runner.
- `occurred_at` / `source_timestamp` stay nullable and keep their NULL `-5` rows. No production runtime code is changed — only the conformance corpus definition and the neo4j conformance runner.

## Collection-count impact
Re-enabling the `created_at` cases on the three stamping legs widens collection by exactly the 10 F-OP `created_at` cases + `F-DATES-and-compose`:

| backend | before | after | Δ |
|---|---|---|---|
| postgres | 190 | 201 | +11 |
| sqlite_lance | 190 | 201 | +11 |
| weaviate | 190 | 201 | +11 |
| python | 209 | 209 | 0 |
| chronicle | 209 | 209 | 0 |
| cypher | 209 | 209 | 0 |
| surrealdb | 190 | 190 | 0 |

Total case count is unchanged (209).

## Test plan
- [x] Python oracle leg: `KHORA_CONFORMANCE_BACKEND=python pytest tests/integration/matrix/test_filter_conformance.py -m filter_conformance` → **209 passed** (recomputes every survivor set; confirms the `lt` / `lte` / `in` edits)
- [x] Chronicle leg → **209 passed**
- [x] Grep gates clean: no remaining `_CREATED_AT_STAMP_BACKENDS` or `_strip_stamped_created_at` references in source
- [x] Construction guard verified: a `created_at` filter over a seed with an unset `created_at` raises `ValueError` at corpus build; an all-stamped seed does not; an `occurred_at` filter with an unset value does not (guard correctly scoped to `created_at`)
- [x] Unit guards: `test_conformance_harness` / `test_conformance_catalog` / `test_system_keys_coverage` → **692 passed** (the 9 failures are the optional `weaviate-client` extra being absent in the local venv — identical on the base branch, environmental)
- [ ] Full cross-store conformance matrix (postgres / sqlite_lance / weaviate / surrealdb / cypher live legs) — runs in CI

### DIFF
--- a/src/khora/filter/conformance.py
+++ b/src/khora/filter/conformance.py
@@ -758,19 +758,6 @@ _LIVE_BACKENDS: frozenset[str] = _TOTAL_BACKENDS | _SPLIT_BACKENDS
 # Every backend a metadata-only / occurred_at / source_timestamp case can run on.
 _OP_BACKENDS: frozenset[str] = _INMEM_BACKENDS | _LIVE_BACKENDS
 
-# The live backends that STAMP ``created_at`` to ``now()`` on insert when absent
-# (postgres + lance via the skeleton temporal store; weaviate via the same store
-# path), so an "absent" row cannot stay NULL there — its ``now()`` value satisfies a
-# lower-bound op and breaks the by-construction ``expected_ids``. A ``created_at``
-# predicate is pruned from these. The surrealdb and cypher runners KEEP ``created_at``
-# cases, but for different reasons: surrealdb leaves an absent ``created_at`` NULL
-# natively (explicit ``option<datetime>`` insert, no default), while the cypher
-# production writer ALSO stamps ``now()`` and the conformance seeder strips it back to
-# NULL post-write (``_strip_stamped_created_at`` in ``_conformance_neo4j.py``).
-# ``occurred_at`` / ``source_timestamp`` are user-supplied and stay NULL everywhere —
-# never pruned.
-_CREATED_AT_STAMP_BACKENDS: frozenset[str] = frozenset({"postgres", "sqlite_lance", "weaviate"})
-
 # The live backends whose runner seeds through ``seed_case`` — i.e. it writes one
 # ``Document`` per record into the relational ``documents`` table (which enforces
 # UNIQUE ``(namespace_id, external_id)``). A case whose seed has a duplicate non-NULL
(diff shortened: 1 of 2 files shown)


filippetrovic:

Reviewed by a 6-role team (security, performance, correctness, test, docs, devil's advocate) plus a dedicated security-audit pass. **Overall assessment: APPROVE.**

The change is test-corpus / test-harness only — no production runtime code. The Python oracle leg recomputes every survivor set, so the three changed expected-id literals (`$lt` / `$lte` / `$in`) are verified by construction.

### MEDIUM
- **[Docs]** `src/khora/filter/conformance.py` — a docstring carried a stray internal cross-reference that does not resolve in this repository. **Resolved in the latest commit** — replaced with the self-contained rationale (the always-present guarantee: NOT NULL column + default + writer stamping). No internal identifiers remain in the diff.

### LOW
- None.

### Findings by domain
- **Correctness** — APPROVE. Verified per-operator against the oracle: with the `created_at` `-5` row moved to the miss instant, `$lt`/`$lte` correctly gain it and `$in` gains it (the miss instant is in the operand); `$gt`/`$gte`/`$eq` correctly exclude it by value; `$ne`/`$nin` correctly keep it by value-mismatch (not NULL-inclusion). The construction guard uses `clause.path[0]` and runs over every leaf from `iter_leaf_clauses`, so composed/negated predicates cannot bypass it. No placeholders.
- **Test** — APPROVE. Re-enabling the 11 cases on the three stamping legs is genuine cross-backend coverage (those legs previously skipped every `created_at` predicate), not redundancy. The guard is exercised on every corpus build; a permanent unit test is unnecessary (self-documenting error message, mirrors the existing construction-error pattern). `_strip_stamped_created_at` removal leaves no orphaned references.
- **Security** — No findings. Test scaffolding with no runtime/trust-boundary surface; the deleted Cypher used parameterized binding; the new `ValueError` interpolates stable test identifiers only.
- **Performance** — No concerns; net positive. Deleting `_strip_stamped_created_at` removes one Cypher round-trip per seed during conformance seeding; the guard is an O(seed) check at collection time only.
- **Docs** — Docstrings/comments are coherent and accurate; no dangling references to the removed prune/helper; the PR description matches the diff (collection-count table correct).
- **Devil's Advocate** — Concurs. The hard `raise` (rather than a silent default) is the right call: auto-stamping would re-hide the exact oracle/stamping divergence this change removes. Guard is appropriately defensive, not over-engineered; scope is tight.

## Security Audit
**Verdict: PASS.** Diff touches only two test-corpus files; no injection, auth, crypto/secret, deserialization, file/path, SSRF, redirect, mass-assignment, or supply-chain surface in the added lines. No dependency or lockfile changes.

---
**Recommendation: APPROVE / ready to merge once CI is green.** The one MEDIUM was fixed in this PR before posting.

probablyrobot (approved):

PR #1219, opened 2026-06-15
Group filter-conformance seeds by external_id and drop the duplicate-external_id prune

filippetrovic:
## Summary
- Group filter-conformance seed records by their non-NULL `external_id` so the seeder writes **one** `Document` per distinct value (records with no `external_id` keep their own). This reproduces the production-normal shape where a single document chunks into N rows that share one denormalized `external_id`, instead of the previous 1:1 record→Document assumption that collided with the `documents` UNIQUE `(namespace_id, external_id)` constraint.
- Add a denorm-consistency guard at the case-construction choke point: records that share a non-NULL `external_id` must agree on the denormalized document fields (`source_type`, `source_name`, `source_url`, `external_id`, `content_type`, `source`, `title`, `source_timestamp`) plus `content`, since a single grouped `Document` can only carry one value for each. The guard raises a clear error naming the offending value and field if a group ever diverges.
- Remove the seed-mechanics prune that excluded duplicate-`external_id` cases from the Document-path backends. With the seeder grouping correctly, the 8 affected operator-coverage cases now run on every backend (the chunk-direct runners were never subject to the constraint and are un-pruned for free). Chunks remain 1:1 with seed records; expected survivor sets are unchanged.

## Test plan
- [x] Python conformance leg green: `KHORA_CONFORMANCE_BACKEND=python pytest ... -m filter_conformance` → 209 passed
- [x] Filter unit suites pass (`test_conformance_harness`, `test_conformance_catalog`, `test_system_keys_coverage`); the only failures are pre-existing environment skips where the optional `weaviate` extra is absent, identical on the base branch
- [ ] Full backend matrix via CI — the 8 `external_id` operator-coverage cases now collect on the postgres / sqlite_lance / weaviate legs (+8 each; other legs unchanged)

### DIFF
--- a/src/khora/filter/conformance.py
+++ b/src/khora/filter/conformance.py
@@ -684,9 +684,18 @@ async def seed_case(coord: StorageCoordinator, case: ConformanceCase) -> dict[st
     )
     namespace_id = ns.id
 
-    id_map: dict[str, UUID] = {}
-    chunks: list[Chunk] = []
-    for record in case.seed_records:
+    # Group seed records by their non-NULL ``external_id`` and write one Document per
+    # group (production-NORMAL: one document chunked into N pieces → N chunks sharing
+    # one denormalized ``external_id``). The ``documents`` UNIQUE ``(namespace_id,
+    # external_id)`` constraint is satisfied because each distinct non-NULL value owns
+    # exactly one Document. Records with ``external_id is None`` each get their own
+    # Document (NULL never violates UNIQUE). The §2 guard guarantees the shared denorm
+    # fields are identical across a group, so they are read off any member.
+    docs_by_external_id: dict[str, UUID] = {}
+
+    async def _document_id_for(record: SeedRecord) -> UUID:
+        if record.external_id is not None and record.external_id in docs_by_external_id:
+            return docs_by_external_id[record.external_id]
         doc = Document(
             namespace_id=namespace_id,
             content=record.content,
@@ -699,10 +708,17 @@ async def seed_case(coord: StorageCoordinator, case: ConformanceCase) -> dict[st
             title=record.title,
         )
         await coord.create_document(doc)
+        if record.external_id is not None:
+            docs_by_external_id[record.external_id] = doc.id
+        return doc.id
 
+    id_map: dict[str, UUID] = {}
+    chunks: list[Chunk] = []
+    for record in case.seed_records:
+        document_id = await _document_id_for(record)
         chunk_kwargs: dict[str, Any] = {
             "namespace_id": namespace_id,
-            "document_id": doc.id,
+            "document_id": document_id,
             "content": record.content,
             "chunk_index": 0,
             "embedding": fake_embedding(record.content, dim=EMBED_DIM),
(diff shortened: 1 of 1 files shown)


filippetrovic:

**Overall assessment: ✅ APPROVE**

A 6-reviewer agent team (Security, Performance, Correctness, Test, Docs, Devil's Advocate) plus a dedicated security audit reviewed this single-file test-infrastructure change. No CRITICAL, HIGH, MEDIUM, or LOW findings require action.

### Findings by severity

**CRITICAL** — none
**HIGH** — none
**MEDIUM** — none
**LOW** — none

### Per-domain summary

| Domain | Verdict |
|---|---|
| Security | No findings — keys are in-memory dedup only, never flow to a query sink; error message uses `!r` repr-formatting; no secret/injection/SSRF surface |
| Performance | Net win — grouping writes **one** `Document` per distinct `external_id` (fewer writes than the old 1:1 loop); guard is O(n) over fixed 4–5-record seeds (sub-µs) |
| Correctness | Grouping logic sound — one Document per distinct non-NULL `external_id`, fresh Document per NULL record, `create_document` called once per Document, `id_map` stays 1:1 record→chunk, guard at the correct choke point |
| Test | Coverage preserved & widened — the 8 `external_id` operator cases now run on the postgres / sqlite_lance / weaviate legs; `expected_ids` byte-identical; guard exercised by the tied seed pair; no stale references to the deleted symbols |
| Docs | Clean — no dangling references to the removed `_SEED_CASE_BACKENDS` / `_seed_has_duplicate_external_id`; reworded comments and the `_backends_for_filter` docstring exclusion list are accurate |
| Devil's Advocate | Approach is sound and minimal — grouping (not seed-dedup or backend-pruning) is the correct way to keep the production-normal duplicate-`external_id`-on-chunks shape under test |

### Discussed & resolved (not a defect)

- **`source_timestamp` in the group-consistency guard's field list** — one reviewer questioned whether `source_timestamp` belongs in the guarded field set, since the seeder stamps it on the chunk rather than the parent `Document`. Adjudicated: the field list deliberately mirrors the change's specification and enforces the production invariant that records grouped under one `external_id` represent N chunks of a **single** document, which in production share the denormalized `source_timestamp`. A seed declaring two divergent `source_timestamp` values for one `external_id` would model an impossible shape, so the guard's strictness is intentional and correct. No current case is affected (the only tied records both leave `source_timestamp` unset), and date-key cases carry no `external_id` so they never group. **No change made.**

## Security Audit

**Verdict: PASS** — scanned all changed lines against the 16 OWASP-aligned categories (injection, auth, crypto/secrets, access control, SSRF, deserialization, file/path, redirects, mass assignment, ReDoS, supply chain, boundary validation, …). Zero findings. The change is test-fixture seeding code with no trust boundary, untrusted input, or dangerous sink.

---

_Verification recap: python conformance leg green (209 passed); the 3 filter unit suites pass (only pre-existing optional-`weaviate`-extra skips, proven baseline-identical); per-leg case delta exactly +8 on each Document-path backend, others unchanged; `ruff`/`ty` clean. Full backend matrix runs in CI._

sale87 (approved):

probablyrobot (approved):

PR #1234, opened 2026-06-15
Fix exact-array $eq recall crash on scalar metadata node in sqlite_lance

filippetrovic:
## Summary
- The sqlite_lance filter compiler's exact-array equality emitter (`_md_exact_array`) ran the inner `json()` parse over the stored metadata node **unconditionally**. When a metadata path held a scalar string (e.g. `{"tags": "gold"}`) instead of an array, SQLite did not reliably short-circuit the `json_type(...) = 'array'` gate before evaluating the right operand, so `json('gold')` raised `OperationalError: malformed JSON` and aborted the **entire** recall query.
- Fixed by gating the comparison inside a `CASE WHEN json_type(...) = 'array' THEN <eq> ELSE 0 END`, wrapped in the unchanged `coalesce((...), 0)`. The inner `json()` parse now runs only when the node is a JSON array; a scalar (or absent) node reads false instead of erroring. Bind order (path, path, operand) is unchanged, so compiled params are unaffected.
- Behavior is otherwise identical: an exact ordered array still matches, a reversed array still does not (order-sensitive), and absent/wrong-type nodes still read false (so a wrapping `NOT` still includes them) — matching the Python oracle.
- Added one cross-engine conformance seed row holding a scalar at the array path (`id="ex-scalar"`, `{"tags": "gold"}`) so the previously-crashing emitter is now exercised against a scalar node on every backend. No expected row-sets changed — the scalar is excluded by both cases built on that seed, by construction.

### Emitter audit
`_md_exact_array` was the only emitter calling the `json()` **parse** function on a node not array-gated inside a `CASE`. `_md_array_contains` / `_md_contains` use `json_each` (scalar-safe). `_md_range` / `_md_null` and the `$exists` presence checks use only `json_type` / `json_extract` (never `json()`-parse). Object-equality dict cases defer to the post-filter and emit no SQL `json()` at all. No other ungated `json()`-parse call remains.

## Test plan
- [x] Unit tests pass (`make test-unit`: 9003 passed, 35 skipped)
- [x] Filter-conformance suite green, incl. the **sqlite_lance** leg (209 passed) — the regression guard. Reverting the fix reproduces `OperationalError: malformed JSON` on `F-ARRAY-exact`; restoring it passes.
- [x] Embedded e2e row-set legs pass (114 passed)
- [x] `make format` clean; `make lint` clean on changed files
- [ ] Integration legs requiring live Postgres/Neo4j/Weaviate run in CI (no Docker in the dev sandbox); the change is sqlite-only SQL text, and the cross-engine row-set invariant gate confirms parity across all backends in CI.


filippetrovic:
## 🔍 Multi-perspective review — consolidated

A reviewer team (security, performance, correctness, tests, database, docs, devil's advocate) plus a dedicated security audit reviewed this change independently.

**Overall assessment: ✅ APPROVE** — no CRITICAL / HIGH / MEDIUM findings. Surgical, spec-accurate fix.

### What was verified
- **Correctness** — The new `CASE WHEN json_type(...) = 'array' THEN <eq> ELSE 0 END` is truthiness-equivalent to the old `<gate> AND <eq>` form for every node shape: array-match → 1, array-non-match → 0, scalar-string node → 0 (was the crash), absent path (`json_type` → NULL) → 0, object/number/null node → 0. The `coalesce((...), 0)` wrapper is unchanged, and `NOT`-wrapping still includes absent/wrong-type rows (matches the Python oracle). Bind order (path, path, operand) is intact, so compiled params are unaffected.
- **Database / SQLite JSON1 semantics** — `CASE` gives true conditional evaluation, so the inner `json()` parse runs only on confirmed arrays; scalar/absent nodes take `ELSE 0` and never reach the parse. Change is `sqlite_lance`-only SQL text — the postgres / cypher / surrealdb / weaviate / python compilers are untouched, so the cross-engine row-set conformance invariant holds. No schema/migration/index impact.
- **Tests** — The single added seed (`ex-scalar`, a scalar at the array path) makes `F-ARRAY-exact` compile and run the previously-crashing emitter against a scalar node on the `sqlite_lance` leg. Zero `expected_ids` change confirmed: the two cases built on that seed (`F-ARRAY-exact`, `F-ARRAY-in-element`) both exclude the scalar by construction.
- **Security** — SQL is built from controlled tokens + bound `?` params; the path goes through `_jpath_bind`, the operand through `_bind`. No interpolation of user text, no new injection surface. Dedicated security audit: **PASS** (no findings; nothing touches auth, crypto, files, network, or dependencies).
- **Performance** — Equivalent per-row cost; the `CASE` actually avoids the inner `json()` parse on non-array rows. Metadata predicates are full-scan on this backend either way — no index/plan regression.
- **Docs** — CHANGELOG `### Fixed` entry is accurate and well-placed; the `_md_exact_array` docstring remains correct (it describes intent, not SQL syntax).
- **Devil's advocate** — Pressure-tested alternatives (text comparison, `json_quote`, Python-side type pre-check); the targeted `CASE` gate is the minimal correct fix, and the single-emitter scope is right (audit confirms no other ungated `json()`-parse site).

### LOW (informational, out of scope)
- A pre-existing CHANGELOG entry (unrelated to this PR, already on the base branch) embeds an internal tracker reference. Not introduced here; flagged only for a future cleanup pass.

_No changes requested._

probablyrobot (approved):

PR #1235, opened 2026-06-15
Run threadable boolean-composition filter cases through real ingest on the multi-channel e2e lanes

filippetrovic:
## Summary

The end-to-end recall-filter suite reconciles real `remember()` → `recall(filter=)` results against the shared conformance corpus. Until now the boolean-composition family of filter cases (`$and` / `$or` / `$nor` / `$not`, De Morgan, distributivity, and — critically — a single logical op spanning a system key **and** a metadata key) was exercised only by the isolated compiled-predicate executor, never through the real multi-channel orchestration + post-filter-combine path.

That combine path (push the system disjunct into the channel query, post-filter the metadata disjunct, then combine the results) exists only on the multi-channel engines — the full graph stack (PG + Neo4j) and the PG-only chronicle stack. A wrong combine on a mixed `$or` would slip past both the isolated executor (it does its own splicing) and the previous e2e suite (which selected no boolean-composition cases). This change closes that gap on exactly those two lanes.

This is a pure test-coverage addition — no production code changes, and no new corpus cases authored (it reuses the existing boolean-composition corpus).

- Add `engine_logic_cases(cases)` to the e2e harness: selects the subset of boolean-composition corpus cases that are seedable through the real ingest path. A case is kept iff every filter leaf is threadable (a system key the harness threads, `source_timestamp`, or any `metadata.*` path), it carries no exact-instant `source_timestamp` `$eq`/`$in` equality (an engine-boundary the row-set selector already excludes — range comparisons stay in), it has a non-empty expectation, and its seeds carry no duplicate reconciliation key. The exact-instant check walks the real parsed AST leaves, so a timestamp equality nested under `$not`/`$and`/`$or` is still caught.
- Add `lane_logic_cases(backend)` resolver mirroring the existing row-set resolver, with the same anti-vacuity empty-raise guard (a mis-mapped lane token or a corpus shrink fails RED at collection rather than parametrizing a vacuously-green lane).
- Add `test_logic_reconciliation_graph` and `test_logic_reconciliation_chronicle`, each parametrized over the threadable subset for its lane. Both run real ingest and reconcile survivors by set-equality (ranking-independent) against the declared expectation and the conformance oracle.
- Extend the existing lane-selection pins so the resolved boolean-composition selection per lane is pinned to the shipped selection — a token slip can't silently drop the cases.
- The embedded / single-path lanes are intentionally left unchanged: the value here is the multi-channel post-filter-combine path, which only the two multi-channel lanes have.

## Test plan

- [x] Selector returns the expected non-empty threadable subset (excludes the case that filters on a date key with no ingest keyword; includes the mixed system+metadata `$or`).
- [x] Both lanes collect the threadable cases (verified under the e2e marker).
- [x] Lane-selection pins hold via the resolver; row-set selection counts unperturbed.
- [x] `ruff format --check` and `ruff check` clean on the changed files; `make lint` exit 0 with no new findings.
- [x] New live-lane tests skip cleanly (not error) on a host without the live stores.
- [ ] Live reconciliation lanes pass in CI (PG + Neo4j provisioned).

### DIFF
--- a/tests/e2e/_harness.py
+++ b/tests/e2e/_harness.py
@@ -33,10 +33,12 @@ from khora.filter.conformance import (
     f_coerce_cases,
     f_dotkey_cases,
     f_exists_cases,
+    f_logic_cases,
     f_objeq_cases,
     f_op_cases,
 )
-from khora.filter.execute import filter_leaf_keys
+from khora.filter.execute import filter_leaf_keys, iter_leaf_clauses
+from khora.filter.model import Op
 from khora.filter.report import FilterPushdownReport
 from khora.query import SearchMode
 
@@ -58,6 +60,16 @@ _REMEMBER_STRING_KEYS: tuple[str, ...] = (
 # an F-OP case). Everything ``remember`` can thread plus the user-supplied
 # ``source_timestamp`` date column and the free-form ``metadata`` blob.
 _ENGINE_ROWSET_KEYS: frozenset[str] = frozenset({*_REMEMBER_STRING_KEYS, "source_timestamp", "metadata.tier"})
+
+# The leaf root segments an F-LOGIC composition may address and still be seedable
+# through ``Khora.remember()``: the five string document keys, the user-supplied
+# ``source_timestamp`` date column, and the ``metadata`` root (a ``metadata.<path>``
+# leaf's first segment — the chunk row carries the whole blob, so any sub-path is
+# threadable). ``occurred_at`` / ``created_at`` are absent — ``remember`` exposes no
+# keyword for them, so an F-LOGIC case composing one of those date keys (e.g.
+# ``F-LOGIC-range-bracket`` on ``occurred_at``) is not threadable through the engine
+# lane and stays covered by the conformance executor suite.
+_ENGINE_LOGIC_KEYS: frozenset[str] = frozenset({*_REMEMBER_STRING_KEYS, "source_timestamp", "metadata"})
 
 # A generous recall ceiling so the whole seed set is returned and the filter is
 # the only narrowing force (conformance seeds are a handful of records each).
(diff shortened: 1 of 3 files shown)


filippetrovic:
## Multi-Perspective Review — PR #1235

**Scope:** pure test-coverage addition — 3 files under `tests/e2e/` (`_harness.py`, `test_filter_rowset_graph.py`, `test_filter_rowset_chronicle.py`), +174 / −1. No production code, no new corpus cases, no dependency changes.

**Team:** Security · Performance · Correctness · Test · Docs · Devil's Advocate, plus a dedicated security audit.

Every reviewer approved/endorsed. No CRITICAL, HIGH, or MEDIUM findings. All findings are LOW and documentation-only; none block merge.

---

### CRITICAL
- None.

### HIGH
- None.

### MEDIUM
- None.

### LOW
- **[Docs]** PR text / `test_filter_rowset_*.py` — the change closes the multi-channel post-filter-combine gap for the **threadable** subset of boolean-composition filters (every leaf seedable through `remember()`); compositions touching `occurred_at`/`created_at` remain out of scope by design. Wording could say "threadable boolean-composition" rather than "boolean-composition" to be precise. Non-blocking.
- **[Devil's Advocate]** `_harness.py` — `engine_logic_cases` repeats the three seed-level guards from `engine_rowset_cases` (`expected_ids is None`, `_has_duplicate_external_id`, threadability). This is deliberate (F-OP keys off `exercises` tags; F-LOGIC must walk the real AST via `iter_leaf_clauses`, so the two selectors can't share one body) but is a minor future-drift risk. Accepted as-is.
- **[Devil's Advocate / Test]** the new logic lane-selection pin (`lane_logic_cases(token)` == `_*_logic_cases()`) is partially tautological: the F-LOGIC corpus is backend-agnostic, so a token slip where both tokens share the same F-LOGIC set would not diverge the id-sets. The real anti-vacuity guarantee is the Layer-1 empty-raise (verified to fire RED on an empty selection). The pin is still justified — it mirrors the rowset sibling and catches corpus shrinks — and the code comment is honest about its scope. Kept intentionally.
- **[Docs — informational, pre-existing, out of scope]** `_harness.py` `_E2E_BACKEND_MAP`: `"skeleton_pgvector": ("postgres", True),  # 46` — the `# 46` count comment is stale (`rowset_cases("postgres", include_system_keys=True)` returns 81). Confirmed via `git log` that this PR does **not** touch or introduce that line. Track separately if desired; not addressed here per surgical scope.

---

### Verification (confirmed by reviewers + lead)
- Selector returns exactly **17/18** threadable F-LOGIC cases — excludes `F-LOGIC-range-bracket` (`occurred_at`, not `remember`-threadable); includes the mixed system+metadata `F-LOGIC-mixed-or` (the core combine-path case) and `F-LOGIC-not-range` (a `$not`-over-range, correctly kept).
- Exact-instant `source_timestamp` `$eq`/`$in` exclusion walks the real parsed leaves (`iter_leaf_clauses`), recursing into `$not`/`$and`/`$or` — **not** via `_is_exact_timestamp_match` (which never matches an F-LOGIC shape). Range ops and `$ne`/`$nin` correctly retained. Empty-path access is short-circuit-guarded.
- Both lanes parametrize the 17 cases and reconcile by set-equality against `case.expected_ids` **and** `conformance.oracle_survivors(case)` (valuable cross-check, not redundant).
- Row-set selection unperturbed (81/81 cypher/chronicle); embedded/skeleton lanes untouched; `ruff format`/`ruff check`/`make lint` clean on the changed files.
- New live-lane tests inherit the module `pytestmark` PG/Neo4j skip guards — skip cleanly without Docker (199 skipped, 0 errors locally) and run in CI under PG + Neo4j.
- **Security audit:** no secrets, no injection sinks, no dependency/network/deserialization surface. Clean.

### Scope-correctness (Devil's Advocate)
Standalone selector (vs folding into `rowset_cases`) is the right call — it keeps the shipped row-set counts and pins unperturbed. Limiting to `vc_full` + `chronicle` is correct: those are the only **multi-channel** lanes with the push-system-disjunct / post-filter-metadata-disjunct / combine path this PR is designed to exercise; skeleton/embedded lanes use a single-channel path and are rightly excluded.

_No changes required to merge._

probablyrobot (approved):

PR #1236, opened 2026-06-15
Push recall filter into recency-channel SQL to stop provenance filter leak

### LINKED ISSUE #1223: Khora.recall() returns results that don't match a source_name/source_type filter, but reports the filter as applied (vectorcypher recency channel)
## Summary

With the vectorcypher recency channel on, `Khora.recall(query, filter=...)` using a
`source_name` or `source_type` filter returns chunks that don't match the filter —
and the result says the filter was applied. A caller who filters out a source still
gets chunks from it, with nothing to flag that the filter was ignored.

Why it happens: the recency channel fetches recent chunks, then rebuilds each one
into the chunk object it returns. The rebuilt object drops the `source_name` and
`source_type` fields. When the channel checks each chunk against the filter, those
fields are empty, so the check can't tell which chunks match — a `!=`/`$nin` filter
keeps everything (including rows it should drop), and a `==`/`$in` filter drops
everything. Either way the filter does nothing, yet the channel records it as
applied.

v0.20.0 added a report (in `engine_info["filter"]`) that says, per filter, whether
recall actually applied it — `unenforced_keys` lists the ones it couldn't. This bug
makes that report wrong: `unenforced_keys` comes back empty while the filter
silently leaks, defeating the guarantee the report was added to give.

## Environment

- Khora 0.20.0 (commit `6f931ed4`)
- Storage backend: postgres+neo4j (the recency channel reads the temporal chunk
  table; the bug is engine-level and backend-independent)
- Engine: vectorcypher, recency channel on (`temporal_recency_channel_enabled=True`)
- Python 3.13, macOS

## Reachability

A normal caller hits this through the public facade:

`Khora.recall(query, filter={"source_name": {"$ne": "secret"}})` on a recency- or
change-style query, with `temporal_recency_channel_enabled=True` (a user-set
`KhoraConfig` field), runs the vectorcypher recency channel. The channel fetches
recent chunks, checks each one against the caller's filter in memory, and records
whether it applied the filter — that record is what the public
`engine_info["filter"]` report is built from. The repro runs that channel directly
with a stub data source (so it's deterministic and needs no DB) and builds the
report the same way the engine does.

The cause is structural: when the channel rebuilds each fetched chunk into the
object it returns, it doesn't copy `source_name`/`source_type`, so the in-memory
filter check never sees those values.

## Reproduction

No DB needed; deterministic. Builds a stored chunk with `source_name="secret"`, runs
the recency channel under `filter={"source_name": {"$ne": "secret"}}`, and reads
back `engine_info["filter"]` the same way the engine builds it. Exits 1 (+
`REPRODUCED`) when the non-matching chunk survives while the report shows the filter
as applied.

```python
#!/usr/bin/env python3
"""Vectorcypher recency channel: a source_name/source_type filter is reported as
applied but is not, so a chunk that does not match the filter is returned while
engine_info["filter"].unenforced_keys comes back [].

Reachability (public entry -> bug):
  Khora.recall(query, filter={"source_name": {"$ne": "alpha"}}) with a
  RECENCY/CHANGE query, temporal_recency_channel_enabled=True
    -> VectorCypherRetriever._vectorcypher_retrieve (the path Khora.recall
       drives for the vectorcypher engine)
    -> ._recency_channel_chunks(filter_ast=..., filter_channel_plans=...)
       [retriever.py:1786] -- the channel under test
    -> build_filter_report(filter_ast, filter_channel_plans)
       [filter/report.py] -- the public engine_info["filter"] payload

This probe drives ._recency_channel_chunks directly (the channel the
recall path invokes) with a stubbed temporal store, then builds the public
engine_info["filter"] report the same way the engine does. No DB needed;
deterministic.

Run:  uv run python probe_vc_recency_filter_leak.py
"""
from __future__ import annotations

import asyncio
import sys
from datetime import UTC, datetime
from uuid import uuid4

from khora.engines.skeleton.backends import TemporalChunk, temporal_chunk_to_chunk
from khora.engines.vectorcypher.retriever import RetrieverConfig, VectorCypherRetriever
from khora.filter import RecallFilter, parse_to_ast
from khora.filter.report import build_filter_report


class _StubVectorStore:
    """Temporal store stub: returns one recent chunk whose provenance VIOLATES
    the caller filter. search_recent_chunks converts a TemporalChunk (which has
    source_name) to a core Chunk (which does not) -- mirroring the real
    pgvector backend (skeleton/backends/pgvector.py:704)."""

    def __init__(self, tc: TemporalChunk) -> None:
        self._tc = tc

    async def search_recent_chunks(self, *, namespace_id, limit, created_after=None):
        # Exactly what the real backend does: TemporalChunk -> Chunk (drops
        # source_name/source_type). Returns (chunk, None).
        return [(temporal_chunk_to_chunk(self._tc), None)]


class _StubEmbedder:
    model_name = "stub"
    dimension = 3

    async def embed(self, text: str):  # pragma: no cover - unused here
        return [1.0, 0.0, 0.0]


async def main() -> int:
    ns = uuid4()
    # A chunk whose source_name is "beta" -- it MUST be excluded by a
    # filter source_name != "alpha"? No: != alpha should INCLUDE beta but the
    # leak we show is for source_name == "alpha" (beta should be DROPPED) and
    # source_name != "alpha" (beta correctly kept, but a chunk with
    # source_name=="alpha" would be WRONGLY kept). We demonstrate the $ne leak:
    # filter source_name != "secret" must DROP a chunk whose source_name=="secret".
    tc = TemporalChunk(
        id=uuid4(),
        namespace_id=ns,
        document_id=uuid4(),
        content="quarterly revenue rose",
        embedding=[1.0, 0.0, 0.0],  # cosine 1.0 vs query -> passes relevance floor
        occurred_at=datetime(2026, 6, 14, tzinfo=UTC),
        created_at=datetime(2026, 6, 14, tzinfo=UTC),
        source_name="secret",  # <- provenance the filter wants to EXCLUDE
        source_type="confidential",
    )

    chunk = temporal_chunk_to_chunk(tc)
    print("step 1 - provenance survives TemporalChunk -> Chunk conversion?")
    print(f"   TemporalChunk.source_name = {tc.source_name!r}")
    print(f"   Chunk has source_name attr = {hasattr(chunk, 'source_name')}")
    print(f"   Chunk has source_type attr = {hasattr(chunk, 'source_type')}")

    cfg = RetrieverConfig()
    cfg.temporal_recency_channel_enabled = True
    cfg.temporal_query_relevance_floor = 0.0  # let the chunk through the cosine gate

    retriever = VectorCypherRetriever(
        vector_store=_StubVectorStore(tc),
        neo4j_driver=None,
        embedder=_StubEmbedder(),
        config=cfg,
    )

    # Caller filter: "exclude chunks whose source_name is 'secret'".
    rf = RecallFilter.model_validate({"source_name": {"$ne": "secret"}})
    filter_ast = parse_to_ast(rf)

    channel_plans: dict = {}
    surviving = await retriever._recency_channel_chunks(
        query_embedding=[1.0, 0.0, 0.0],
        namespace_id=ns,
        temporal_filter=None,
        filter_ast=filter_ast,
        filter_channel_plans=channel_plans,
    )

    leaked = [ch for (_id, _s, ch) in surviving]
    print("\nstep 2 - did the recency channel keep the filter-violating chunk?")
    print(f"   caller filter           : source_name != 'secret'")
    print(f"   chunk's real source_name: {tc.source_name!r}  (MUST be dropped)")
    print(f"   chunks surviving filter : {len(leaked)}")

    # Build the public engine_info["filter"] report exactly as the engine does.
    report = build_filter_report(filter_ast, channel_plans)
    print("\nstep 3 - what does engine_info['filter'] CLAIM about source_name?")
    print(f"   recency ChannelPlan post_filtered_keys : {sorted(channel_plans['recency'].post_filtered_keys) if 'recency' in channel_plans else '(none)'}")
    print(f"   report.unenforced_keys                 : {report.unenforced_keys}")
    print(f"   report.post_filtered_keys              : {report.post_filtered_keys}")

    leak = len(leaked) == 1  # the secret chunk survived a filter that should drop it
    dishonest = report.unenforced_keys == [] and "source_name" in report.post_filtered_keys

    print()
    if leak and dishonest:
        print("REPRODUCED: recency channel returned a chunk that does NOT match "
              "source_name != 'secret' (source_name was dropped when the chunk was "
              "rebuilt, so the filter check sees an empty value and != keeps it), "
              "while engine_info reports unenforced_keys==[] and source_name in "
              "post_filtered_keys -- recall says the filter was applied while it leaks.")
        return 1
    print(f"NOT reproduced -- leak={leak} dishonest_report={dishonest} "
          f"(surviving={len(leaked)}, unenforced_keys={report.unenforced_keys})")
    return 0


if __name__ == "__main__":
    sys.exit(asyncio.run(main()))
```

Run:

```
uv run python probe_vc_recency_filter_leak.py
```

## Expected behavior

A chunk whose `source_name` is `"secret"` is excluded by
`filter={"source_name": {"$ne": "secret"}}` on every channel. If the recency
channel can't actually apply a filter, `engine_info["filter"].unenforced_keys`
should list it, so the caller knows that filter wasn't enforced.

## Actual behavior

The chunk that should be excluded is returned, and the report says the filter was
applied:

```
step 1 - provenance survives TemporalChunk -> Chunk conversion?
   TemporalChunk.source_name = 'secret'
   Chunk has source_name attr = False
   Chunk has source_type attr = False

step 2 - did the recency channel keep the filter-violating chunk?
   caller filter           : source_name != 'secret'
   chunk's real source_name: 'secret'  (MUST be dropped)
   chunks surviving filter : 1

step 3 - what does engine_info['filter'] CLAIM about source_name?
   recency ChannelPlan post_filtered_keys : ['source_name']
   report.unenforced_keys                 : []
   report.post_filtered_keys              : ['source_name']

REPRODUCED: recency channel returned a chunk that does NOT match source_name != 'secret'
(source_name was dropped when the chunk was rebuilt, so the filter check sees an empty
value and != keeps it), while engine_info reports unenforced_keys==[] and source_name in
post_filtered_keys -- recall says the filter was applied while it leaks.
```

The channel records every filter as applied no matter which ones it could actually
check, so the `source_name`/`source_type` filter is reported as applied even though
the chunk it checked has no such field. Fix: check chunks against the stored object
that still carries `source_name`/`source_type`, or only record a filter as applied
when it was actually checked.

filippetrovic:
## Summary

The VectorCypher recency channel reported a `source_name`/`source_type` (provenance) filter as enforced while silently leaking it. On a RECENCY/CHANGE query with the pgvector temporal store and `temporal_recency_channel_enabled=True`, `recall(query, filter={"source_name": {"$ne": "secret"}})` returned chunks whose `source_name` **was** `"secret"`, while `engine_info["filter"].unenforced_keys == []` and `source_name` appeared in `post_filtered_keys` — i.e. the filter leaked and the report claimed it was applied.

**Root cause:** the recency channel rebuilt each fetched chunk into a public `Chunk` that structurally cannot carry provenance fields, then ran an in-memory post-filter against that blank object. Every provenance leaf resolved to `None`, so a `$ne` filter kept rows it should drop (leak) and a `$eq` filter dropped rows it should keep (silent recall loss) — yet the channel unconditionally credited every leaf as enforced.

**Fix:** push the recall filter down into the `khora_chunks` SQL `WHERE` clause, mirroring the vector channel:

- `TemporalVectorStore.search_recent_chunks` gains backward-compatible `filter_ast` / `filter_plan_out` keyword params (default `[]` impl unchanged for backends without a recency table).
- `PgVectorTemporalStore.search_recent_chunks` compiles `filter_ast` with the same raise-mode `khora_chunks` compile the vector path uses, AND-s the predicate into the recency `WHERE`, and reports an honest `ChannelPlan(pushed_keys=...)`.
- The recency channel drops its in-memory post-filter, records the plan from the store only when it produced gating rows, and re-raises `RecallFilterUnsupportedError` so an unpushable leaf fails loud (matching the vector channel) while operational faults still degrade to `[]`.

A filter-violating recent chunk can no longer be fetched, and `engine_info["filter"]` reports `pushed_keys` honestly with `unenforced_keys == []`.

Fixes #1223.

## Test plan

- [x] Unit tests pass — `tests/unit/engines/test_recency_channel.py` (new cases: `$ne` excludes the violating chunk and records no plan when nothing survives; `$eq` returns it with `source_name` in `pushed_keys` and `post_filtered_keys == []`; the leaked-chunk regression that fails on pre-fix code)
- [x] Filter report / compile / skeleton-filter unit suites stay green
- [ ] Integration test passes in CI — `tests/integration/test_vectorcypher_recency_channel_pg.py::test_recency_channel_source_name_filter_no_leak` (real Postgres; seeds a violating recent chunk, asserts it is absent and the report is honest). Requires the docker pg+neo4j stack, exercised in CI.


filippetrovic:
## Consolidated Review — PR #1236

A multi-perspective review team was dispatched across 7 domains (security, performance, correctness, database, tests, docs, devil's advocate). The background reviewer agents did not return findings in this run, so the lead consolidated the review from a direct, per-dimension verification of the diff. Findings below are grouped by severity.

**Overall assessment: APPROVE** — no CRITICAL or HIGH findings. The change is a faithful, minimal port of the established vector-channel pushdown pattern; the integration test that reproduces the original leak passes in CI.

### CRITICAL
- None.

### HIGH
- None.

### MEDIUM
- **[Correctness / Observability] `retriever.py` `_recency_channel_chunks` — operational-fault path returns `[]` without recording a structured degradation.** The generic `except Exception: return []` (pre-existing; not introduced here) silently drops the recency channel on a DB/network fault. The vector + BM25 channels still enforce the filter and carry the report, so there is no honesty violation, but per the project's failure-observability convention a caught operational fault on a retrieval channel should record a degradation signal. Pre-existing behavior — recommend a follow-up rather than blocking this fix.

### LOW
- **[Docs] Architecture docs may reference the recency channel's old in-memory post-filter behavior.** The code docstrings were updated, but `docs/architecture/*` prose describing the recency channel as "post-filtering" may now drift. Non-blocking; worth a follow-up sweep.
- **[Test] Stub-based unit coverage simulates the SQL pushdown rather than exercising real SQL.** This is appropriate (no DB in the unit lane) and is backstopped by the real-Postgres integration test (`test_recency_channel_source_name_filter_no_leak`), which runs in the `vc_full`/`vc_embedded` CI legs.

### Verification performed (per dimension)
- **Security:** `namespace_id == namespace_id` is unconditionally the first `WHERE` condition on the recency SQL — no cross-namespace path is opened. The filter predicate is compiled via `compile_postgres` (SQLAlchemy bound parameters, no string interpolation → no injection). No secrets introduced.
- **Performance:** Mirrors `_search_inner` exactly; the `ORDER BY COALESCE(occurred_at, source_timestamp, created_at) DESC LIMIT` continues to match the recency index expression. No N+1 or extra round-trip.
- **Correctness:** `compile_postgres(on_unsupported="raise")` raises exactly `RecallFilterUnsupportedError`; the channel re-raises it (matching the vector channel's fail-loud contract) ahead of the generic operational-fault handler. The plan is recorded only when the channel produced gating rows; a leaf the recency channel skips is still credited by the gating vector/BM25 channel, so no false `unenforced_keys` entry arises.
- **Database:** Full-row select preserves `embedding`; the compile context (`khora_chunks`, raise mode) matches the vector path; no schema/migration change.
- **Tests:** New unit + integration regressions fail on pre-fix code and pass post-fix; CI conformance/skeleton/vectorcypher legs are green.
- **Devil's advocate:** The raise-mode contract change (vs the prior split-mode degrade) is intentional and consistent with the vector/BM25 channels, which already use raise mode — an unpushable leaf now fails loud uniformly rather than silently no-op'ing on the recency axis. Scope is tight (legacy `storage/backends/pgvector.py` recency method untouched).

_References GitHub issue #1223._

filippetrovic:
## Consolidated Review — Update (reviewer team findings)

The 7-domain reviewer team returned its findings. All reviewers **APPROVE**; no CRITICAL/HIGH correctness or security blockers. Two non-blocking items were actionable and have been **addressed in this PR** (commit follows).

| Domain | Verdict | Notable finding |
|---|---|---|
| Security | APPROVE | Predicate parameterized via SQLAlchemy bound params; `namespace_id` unconditionally in WHERE — no injection, no cross-namespace path. |
| Performance | APPROVE | Index-served `ORDER BY COALESCE(...) DESC LIMIT` preserved (`ix_khora_chunks_ns_recency`); no seq-scan/N+1 risk vs the vector path. |
| Database | APPROVE | Full-row select preserves `embedding`; compile context matches `_search_inner`; no schema/migration/FK impact. |
| Correctness | APPROVE | Exception ordering correct; plan-gating partition cannot mislabel a leaf `unenforced`; operational-fault `[]` degrade is intentional (pre-existing) and safe. |
| Tests | APPROVE | Regressions genuinely fail on pre-fix code; stubs faithfully model the SQL pushdown. Flagged: no test for the unsupported-leaf raise path. |
| Docs | APPROVE | Docstrings accurate; no repo-docs drift. Flagged: one stale "post-filtered chunks" comment. |
| Devil's advocate | APPROVE | SQL pushdown is the necessary fix (public `Chunk` cannot carry provenance); raise-mode is consistent with the vector channel; scope is tight. |

### Fixes applied in response
- **[Test]** Added `test_unsupported_filter_leaf_propagates_fail_loud` and `test_operational_fault_degrades_to_empty` — pinning the fail-loud (`RecallFilterUnsupportedError` re-raise) vs operational-fault (degrade to `[]`) contract this change introduced.
- **[Docs]** Corrected the stale plan-assembly comment that still described the recency channel as recording "post-filtered chunks" (it records its SQL-pushdown plan for surviving gating chunks).

### Deferred (non-blocking, by design / convention)
- Wider operator coverage (`$in`/`$nin`, compound multi-leaf filters) — coverage expansion, not a correctness gap; the core `$ne`/`$eq` + integration paths are covered.
- The raise-mode contract is documented in this PR's description; per repo convention the CHANGELOG is updated at release/version-bump time, not mid-cycle.

**Overall: APPROVE.** No correctness/security blockers; the regression that reproduced the original leak passes against real Postgres in CI.

_References GitHub issue #1223._

filippetrovic:
## Follow-up: two recency-channel review items addressed

Implemented the two non-blocking items surfaced in review (commit `ac2ef6eb`):

1. **Plan-recording symmetry** — the recency `ChannelPlan` is now recorded only when `filter_ast is not None` (symmetric with the vector/BM25 channels). A no-filter recall no longer appends a benign empty `"recency"` entry; the report was already honest, this removes the asymmetric uncovered branch.
2. **Degradation observability (failure-observability contract)** — an operational fault (DB/network) in `search_recent_chunks` now records a structured `Degradation` on `RecallResult.engine_info["degradations"]` and increments a new `khora.vectorcypher.recency_channel.degraded_total` counter, matching the bm25/rel_fetch/cypher_expand convention. A `RecallFilterUnsupportedError` still propagates (fail-loud); only operational faults degrade to `[]`. The new metric is registered in the telemetry contract.

Added unit coverage for both: the no-filter no-record path, the operational-fault degradation record, and the fail-loud propagation (alongside the existing leak/no-leak and `$ne`/`$eq` cases).

_References GitHub issue #1223._

probablyrobot (approved):

PR #1245, opened 2026-06-15
Add property-based differential fuzzer for recall-filter compilers

filippetrovic:
## Summary

The recall-filter conformance harness pins a hand-authored *catalog* of cases — precise but finite, covering only the shapes a human thought to enumerate. This adds a complementary property-based fuzzer that *generates* filters with Hypothesis and checks the same oracle contract on every draw, so a divergence the catalog never enumerated surfaces as a shrunk counterexample instead of an escaped bug.

New file `tests/unit/filter/test_filter_fuzz.py` (`@pytest.mark.unit`, runs in the existing unit-test CI job — no workflow change). Two properties over one shared, validity-biased filter-generation strategy:

- **Property A — differential (primary).** For a generated valid filter over a fixed ~18-record seed, the Python oracle (`compile_python`) and the real `sqlite_lance` read path (`compile_lance` server-side prefilter ∘ the `compile_python` split post-filter) must keep the same record ids. 300 examples, plus a separate 8-seed sweep (real `pytest` parametrization with distinct Hypothesis seeds) to confirm the agreement is stable across independent PRNG streams, not an artifact of one. `sqlite_lance` is the differential target because it is a genuinely independent compiler (chronicle's post-filter *is* the oracle, so it cannot disagree).
- **Property B — metamorphic (oracle-only).** Four equivalence-preserving rewrites must not change the oracle row-set: implicit-AND ⇔ `$and`, `$nor` ⇔ `$not($or)`, De Morgan, and double-negation. A dedicated strategy emits only the transformable shapes (a generic strategy gated by `assume` would trip Hypothesis's `filter_too_much`).

Anti-false-green guards so the differential cannot pass vacuously:
- seed-materialization assertion (every seed record has a row; the id mapping is 1:1),
- a strict-subset discrimination check (some draws keep a non-trivial subset, not all/none),
- a deferred-leaf post-filter agreement test (proves the split post-filter is actually wired and applied, exercising the exact-array-on-scalar-metadata path end-to-end).

The fuzzer ships green with no carve-out: the previously-known bare-list `$eq` exact-array crash on a scalar metadata node is already fixed upstream, and the strategy + seed exercise that path for real.

Test-only addition; no production code changes.

## Test plan
- [x] `tests/unit/filter/test_filter_fuzz.py` passes (18 cases, ~3.3s)
- [x] Full unit suite green (`make test-unit`: 9027 passed, 35 skipped)
- [x] `make format` clean, `make lint` clean
- [ ] CI: full integration + e2e suite

### DIFF
--- a/tests/unit/filter/test_filter_fuzz.py
+++ b/tests/unit/filter/test_filter_fuzz.py
@@ -1,0 +1,904 @@
+"""Property-based differential fuzzer for the recall-filter compilers.
+
+The recall-filter conformance harness (``khora.filter.conformance``) pins a
+hand-authored *catalog* of cases: each is lowered through the real validator +
+``parse_to_ast`` + every backend compiler, and the surviving row-set is asserted
+against the Python oracle (``compile_python``). That catalog is precise but
+finite — it covers the shapes a human thought to enumerate.
+
+This module is the complementary force: it *generates* filters with Hypothesis
+and checks the same oracle contract on every draw, so a divergence the catalog
+did not happen to enumerate surfaces as a shrunk counterexample instead of an
+escaped production bug. Two properties:
+
+**Property A — differential (primary).** For a generated valid filter over a
+single FIXED seed, the Python oracle and the real ``sqlite_lance`` read path
+(the ``compile_lance`` server-side prefilter ∘ the ``compile_python``
+post-filter) must keep the SAME ``SeedRecord`` ids. sqlite_lance is the
+differential target — NOT chronicle, whose post-filter *is* ``compile_python``
+(the oracle itself), so it cannot disagree the way an independent compiler can.
+A separate sweep re-runs Property A under eight distinct Hypothesis seeds to
+confirm the agreement is stable, not an artifact of one PRNG stream.
+
+**Property B — metamorphic (oracle-only).** Four logical rewrites that MUST
+preserve the oracle row-set: implicit-AND ⇔ ``$and``, ``$nor`` ⇔
+``$not($or)``, De Morgan, and double-negation. A divergence here is a
+``parse_to_ast`` / ``compile_python`` normalization bug, caught without any
+store. A dedicated strategy emits only the transformable shapes (a generic
+strategy gated by ``assume`` would trip Hypothesis's ``filter_too_much``).
+
+The bare-list-``$eq`` exact-array-on-scalar-metadata crash fixed in PR #1234 is
+in scope and exercised with no carve-out: the strategy freely draws bare-list
+metadata operands, and the seed carries both array and scalar metadata nodes.
+"""
+
+from __future__ import annotations
+
+import tempfile
+from collections.abc import Mapping, Sequence
+from datetime import UTC, datetime
+from functools import lru_cache
+from pathlib import Path
+from typing import Any
+from uuid import UUID
+
+import pytest
+from hypothesis import HealthCheck, assume, given, seed, settings
+from hypothesis import strategies as st
+
+from khora.db.session import run_migrations
+from khora.filter import (
+    CompiledFilter,
+    RecallFilter,
+    RecallFilterUnsupportedError,
+    RecallFilterValidationError,
+)
+from khora.filter.ast import FilterNode, parse_to_ast
+from khora.filter.conformance import (
+    ConformanceCase,
+    LanceExecutor,
+    PythonExecutor,
+    SeedRecord,
+    _record_mapping,
+    seed_case,
+)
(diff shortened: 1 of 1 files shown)


filippetrovic:
## Consolidated Review — PR #1245 (multi-perspective agent team)

**Overall assessment: ✅ APPROVE.** Seven independent reviewers (Security, Performance, Correctness, Test-quality, Database, Docs, Devil's Advocate) plus a dedicated security audit. No CRITICAL or HIGH findings. Two MEDIUM (one cosmetic-docs, one advisory-maintainability) and a couple of LOW/FYI items. This is a clean, well-guarded, test-only addition (one new file, zero production changes).

### MEDIUM
- **M1 [Docs]** `tests/unit/filter/test_filter_fuzz.py` — `test_post_filter_deferred_leaf_still_agrees`: the docstring/name describe a *post-filter-deferred* leaf, but the leaf it uses (bare-list `$eq` exact-array on `metadata.tags`) is SQL-pushed via the `#1234` `_md_exact_array` CASE gate, **not** deferred to the Python post-filter. The assertions are correct (oracle == lance; `r01` array matches, `r05` scalar does not; strict subset), so this is a wording/naming-accuracy nit, not a behavioral bug. Post-filter application is genuinely exercised elsewhere (the general Property A run draws deferred `metadata.$date` leaves, and `test_match_everything_keeps_all_rows` guards row-loss). Recommend renaming to reflect the exact-array path it actually covers, or swapping in a truly-deferred leaf. *Non-blocking.*
- **M2 [Maintainability — Devil's Advocate]** Over-engineering opinion: the strict-subset discrimination guard (`_SURVIVOR_SET_SIZES` + `test_a_meaningful_fraction_of_draws_are_strict_subsets`) and the three separate Property-B shape strategies could be trimmed (~100 lines). *Lead note:* we are intentionally **keeping** the discrimination guard — it defends against the differential's specific false-green mode (both compilers trivially returning all/none on every draw, which a pure `oracle == lance` assertion would pass vacuously), and the Test reviewer independently validated it as a real hard assertion. The three Property-B strategies are kept because each transform needs a distinct shape (distinct-keys for implicit-AND, lists for `$nor`, pairs for De Morgan). Recorded as advisory, not adopted.

### LOW / FYI
- **L1 [Maintainability]** `_seeded_store()` — a docstring note on the loop-thread re-entrancy hazard would help future contributors (the code already resolves the `@lru_cache` singleton on the caller thread and only submits the read coroutine to the loop thread; this is correct as-is).
- **L2 [Resource]** The embedded SQLite store is a process-lived singleton left to interpreter shutdown (standard for the conformance harness pattern; ~10 MB temp dir). Acceptable; an `atexit` close would be tidier.

### Security Audit — Verdict: ✅ PASS
Diff-scoped audit across the 16 categories: **no findings.** The only SQL sinks are a static `SELECT COUNT(*)` and a parameterized `execute(sql, args)` with positional binds (column token + `?` placeholders + compiler-generated predicate; no untrusted interpolation). No secrets, no unsafe deserialization, no outbound requests, no regex/ReDoS, no dependency/manifest changes. `tempfile.mkdtemp` is used safely. (Pattern hits for `raw(`/`DES` were false positives — substrings of `draw(` and `nodes`/`sides`.)

### What the reviewers confirmed (highlights)
- **Correctness:** lance runner composes SQL prefilter **and** applies the `compile_python` post-filter on survivors; chunk-UUID↔seed-id mapping correct; `predicate == "1"` handled; only legitimate `assume`-skips (invalid draw, oracle-unsupported); **no carve-out** (the `#1234` fix is exercised for real); seed invariant honored (`created_at` always set; `source_timestamp` ⇒ matching `occurred_at`).
- **Test-quality:** both properties non-vacuous; anti-false-green guards are real hard assertions; the 8-seed sweep is genuine `pytest` parametrization with distinct `@seed()` streams; Property B's 4 transforms each use a dedicated transformable-shape strategy (no `filter_too_much`); health-check suppressions justified.
- **Performance:** example budget bounded (~3.3s locally); store built exactly once via module `@lru_cache`; one cheap scoped SQL per draw; bounded recursion; per-worker singletons safe under xdist.
- **Database:** embedded coordinator + read path mirror the real conformance `run_live`; re-entrancy safe; no production schema/ORM/migration change.
- **Open-source compliance:** no internal IDs/ADR refs anywhere; only the allowed `#1234` GitHub reference appears in code.

_Reviewed by an automated multi-perspective agent team. MEDIUM items are non-blocking; M1 (docstring/name accuracy) is a worthwhile pre-merge or follow-up tidy._

sale87 (approved):

PR #1246, opened 2026-06-15
Add channel × filter-operator coverage matrix for post-filter retrieval channels

filippetrovic:
## Summary

Closes a whole class of filter-enforcement leaks (the GitHub #1223 family) with a behavioral coverage matrix plus a durable registry meta-test — not a one-off regression.

- **Matrix (× filter operators):** forces each VectorCypher *post-filter* retrieval channel to FIRE under leak-prone operators and asserts it actually ENFORCED the filter (not merely threaded it). Rows: recency (`_recency_channel_chunks`), graph metadata over-fetch (`_fetch_chunks_from_entities`), session fan-out, CHANGE/version-history. Columns: provenance-key `$eq`/`$in`/`$ne`/`$nin`, `$exists`, `$ne`/`$nin` missing-inclusion, metadata present-null vs absent.
- **Per-cell invariants (mechanism-agnostic):** violating chunk absent (and satisfying chunk present on over-drop columns); `unenforced_keys == []`; filtered key ∈ `pushed_keys ∪ post_filtered_keys` with the total-disjoint partition invariant; and a **load-bearing non-vacuity guard** proving the channel really fired (so a "pushed-down, not post-filtered" cell still proves enforcement). This stays correct whether a channel pushes the filter into SQL or post-filters in memory.
- **Registry meta-test (hermetic, runs in the unit lane):** partitions the filter-enforcement audit gate's filter-aware method set into post-filter seams vs documented exclusions; a new filter-aware channel that is neither classified nor excluded fails the gate, and every declared (row × column) cell must have a collected test. This is the durable guard that stops the class from silently regressing when a new channel lands.
- **Pre-fix proof:** a self-contained monkeypatch reproduces the pre-fix recency leak; the recency × provenance-`$ne` cell fails on that behavior and passes on current code — demonstrating the matrix catches #1223. Nothing is reverted in committed source. Expected pre-fix failure:
  `AssertionError: filter-violating chunk leaked through the 'recency' channel: doc external_id='violating', value at ('source_name',)='leakdoc', leaf op=Op.NE, operand='leakdoc'`
- **Test-only:** no `src/` changes. Registry + meta-tests live in `tests/recall/` (hermetic, unit lane); the live channel-firing cells live in `tests/integration/` (run on the existing PG+Neo4j integration lane). No CI workflow change.

## Test plan
- [x] Hermetic registry/partition meta-tests pass in the unit lane (`tests/recall/test_channel_filter_operator_matrix.py`)
- [x] Cell-presence guard verified load-bearing (removing a cell turns the meta-test red)
- [x] `make format` / `make lint` clean; no `src/` changes
- [ ] Integration cells execute on the live PG+Neo4j lane (recency/graph/session/CHANGE + recency pre-fix proof) — they self-skip without infra and run on CI

### DIFF
--- a/tests/integration/test_channel_filter_operator_matrix.py
+++ b/tests/integration/test_channel_filter_operator_matrix.py
@@ -157,9 +157,13 @@ async def vectorcypher_recency_kb(monkeypatch: pytest.MonkeyPatch) -> AsyncItera
 
 # ===========================================================================
 # RECENCY ROW — the #1223 seam. ``_recency_channel_chunks`` pushes every leaf
-# into the khora_chunks WHERE (post-#1236), so the pinned leaf is PUSHED and the
-# recency channel's ``post_filtered_keys`` is empty (asserted in the driver's
-# MECHANISM block).
+# into the khora_chunks WHERE (post-#1236), so when the recency channel records a
+# ChannelPlan the pinned leaf is PUSHED and its ``post_filtered_keys`` is empty
+# (asserted in the driver's MECHANISM block, which the floor may skip). The
+# always-on recency guard is the ``search_recent_chunks`` RETURN-boundary spy:
+# non-vacuity (>= 1 returned row) + #1223 enforcement (the violating doc is never
+# in the returned rows), floor-INDEPENDENT — see ``_install_recency_spy`` /
+# ``_assert_recency_fired``.
 # ===========================================================================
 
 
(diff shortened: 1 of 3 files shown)


filippetrovic:

Six independent reviewers (correctness, tests, security, performance, docs, devil's advocate) plus a security audit. **CI is fully green** (the live PG+Neo4j integration lane passes all 11 matrix cells + the pre-fix proof). No blocking issues found.

**Overall assessment: APPROVE.** Findings below are non-blocking polish — mostly doc wording and one efficiency nit.

### Strengths
- **Correctness:** registry partition is exact and disjoint (2 post-filter seams + 7 documented exclusions = 9 filter-aware methods); the mechanism-agnostic per-cell contract holds (pinned key ∈ `pushed_keys ∪ post_filtered_keys`, never hard-requires post-filter); `$ne`/`$nin` missing-inclusion semantics correct.
- **Non-vacuity is load-bearing:** each row proves its channel fired *and* enforced — recency via a floor-independent return-boundary capture, graph via the fetch spy + recorded plan, session via ≥2 fan-out searches, change via a distinct decomposed rewrite. The cell-presence meta-test (AST source-scan over all four rows) makes silent under-coverage impossible and runs in the no-infra unit lane.
- **Pre-fix proof** faithfully reproduces the pre-#1236 leak via an auto-undone, scoped monkeypatch (no revert in committed source) and pins the failure message so it can't pass for the wrong reason.
- **Security:** test-only, no `src/` changes; `neo4j_password` typed `SecretStr`; backdating SQL fully parameterized; all seeded content synthetic. Security audit added nothing.

### Findings

**MEDIUM**
- **M1 (docs):** the column-B `$exists` oracle note (`tests/recall/test_channel_filter_operator_matrix.py` `_OPERATOR_CLASSES["B"]`) is correct but could separate the two ideas more crisply — *system/provenance-key `$exists` compiles to constant-true (not a real presence test); only metadata-key `$exists` is a genuine presence test* — and cite where the three-compiler behavior is verified.
- **M2 (perf):** `_backdate_recency` opens its own `create_async_engine`/`dispose` per recency cell (~1s total across the run). Reusing the already-connected `kb` engine would remove it. Minor on a ~7-min lane.
- **M3 (docs):** no "extending the matrix" checklist — adding a new seam/column touches ~6 places; a short subsection would help future contributors (the partition meta-test already catches the registry half).

**LOW**
- **L1 (tests):** the session/change module's `pytestmark` gates on `_pg_reachable()` but not `_neo4j_reachable()` (the recency/graph module checks both). Cosmetic — both run in the same PG+Neo4j CI lane — but worth aligning for local robustness.
- **L2 (correctness/devil's-advocate):** the recency non-vacuity is layered (store-boundary capture *plus* the universal `result.chunks` enforcement check). This is intentional and not a gap, but a one-line comment noting the dual signal would aid maintenance.
- **L3 (docs):** minor terminology drift ("seams" vs "sense") and a past-tense mechanism note that could point at the pre-fix proof.

### On the complexity objection (devil's advocate)
The devil's advocate flagged the registry + 7 meta-tests + shared cell driver as overengineered (~1.7k lines for 11 cells) and suggested dropping the registry. Noting for the record: the registry meta-test (a new filter-aware channel must be consciously classified as post-filter seam or documented exclusion) and the pre-fix proof are the **core deliverables** of this change — a behavioral matrix alone would not stop the bug class from silently regressing when a new channel lands, which is the whole point. The legitimate residual is that the shared `_assert_channel_cell` driver is large; factoring it is a reasonable future cleanup, not a merge blocker. The reviewer themselves conceded the new-channel-backstop goal is sound and that the two "HIGH" vacuity concerns are "not a real gap."

_None of the findings block merge; they're captured here for an optional touch-up or follow-up._

sale87 (approved):

PR #1250, opened 2026-06-16
Declare numpy as a direct core dependency

msrdic:
## Summary
- Add `numpy>=2.1` to `[project] dependencies` in `pyproject.toml`. numpy is imported unconditionally on the eager `import khora` path (`hooks/embedding_filter.py`) but until now was only declared **transitively** (it rode in via `sentence-transformers → torch → numpy`). Declaring it directly fixes a latent undeclared-dependency bug and keeps `import khora` self-sufficient if torch ever leaves the default install.
- Floor rationale: `requires-python >=3.13` ⟹ `numpy>=2.1` is the first release shipping cp313 wheels. The fleet already resolves to numpy 2.4.4, so there is no 1.x→2.x ABI migration risk.
- Add a regression test (`tests/unit/test_core_dependencies.py`) that parses `[project].dependencies` and asserts numpy is a declared **core** dependency with a `>=2.1` floor — guarding against it silently reverting to transitive-only.
- `CHANGELOG.md`: note the change under the unreleased `### Changed` section.
- No source behavior changes. The unconditional import is untouched; the guarded numpy uses (`_accel.py`, `pgvector`, surrealdb, `temporal_detection`, `embedders/litellm`) stay guarded.

## Test plan
- [x] `make format` clean; `make lint` (ruff + ty) exits 0
- [x] New regression test passes; full unit/recall/security suite green (8923 passed, 0 failed)
- [x] `python -c "import khora"` succeeds
- [ ] CI integration/e2e pass (rebuilds the Rust extension fresh)


msrdic:

Reviewed by a six-perspective team (Security, Performance, Correctness, Test, Docs, Devil's Advocate) plus a dedicated security audit. The change adds `numpy>=2.1` to `[project] dependencies`, a regression test, a CHANGELOG entry, and the corresponding `uv.lock` edges.

**Overall assessment: ✅ APPROVE.** No CRITICAL / HIGH / MEDIUM findings. A small, well-scoped, surgical dependency-declaration change.

### Findings by severity
_None at CRITICAL / HIGH / MEDIUM._

**LOW**
- **L1 (Docs):** The new `### Changed` CHANGELOG bullet has no issue/PR reference, while sibling bullets cite one. Acceptable for a pure dependency-declaration; non-blocking.

### Per-dimension verdicts
- **Correctness:** `numpy>=2.1` is the correct floor for `requires-python >=3.13` — 2.1.0 is the first numpy release shipping cp313 wheels (a looser `>=2.0` would let a resolver pick a wheel-less 2.0.x on 3.13). Placement in core `[project] dependencies` is right: `import khora` reaches an unconditional `import numpy as np` in `hooks/embedding_filter.py` on the eager import path, so an extra would let a minimal install `ImportError`. `uv lock --check` passes; no source behavior change.
- **Test:** `tests/unit/test_core_dependencies.py` meaningfully guards two invariants — numpy is a declared *core* dependency, and its floor is `>=2.1`. It fails if numpy is dropped from core deps or the floor regresses. `packaging`-import fallbacks are defensively sound.
- **Security:** No findings. No secrets/credentials in the diff; no source-execution change. Supply-chain impact is neutral — numpy was already in the resolved graph (locked 2.4.4, no relevant CVE); promoting transitive→direct only makes the existing requirement explicit.
- **Performance:** No findings. numpy was already resolved transitively, so install footprint, resolution, and import-time cost are unchanged. Locked version unchanged at 2.4.4.
- **Docs:** CHANGELOG entry is correctly placed under `### Changed`; the pyproject rationale comment matches surrounding style. No internal identifiers in added lines.
- **Devil's Advocate:** No blocking concerns. Core-deps (not an extra, not transitive) is the correct call given the unconditional import. The floor does not conflict with other deps' numpy constraints; resolution for the crewai/google-adk extra combos is unaffected.

### Security audit
OWASP/dependency/crypto pass over the diff: no source code executes, no secrets, no new network/IO surface. Sole vector is supply-chain (numpy), which is unchanged at the locked 2.4.4 — no actionable findings.

sale87 (approved):

PR #1334, opened 2026-06-22
Add regression guard for vectorcypher-skeleton engine isolation

msrdic:
## Summary
- Add `tests/unit/test_engine_isolation.py`, an amendable regression guard asserting that `khora.engines.vectorcypher` (and its submodules) imports nothing from `khora.engines.skeleton`. The two engines are independent implementations of the memory-engine protocol and must not reach into each other's internals; this test locks that boundary in so an *accidental* re-coupling trips loudly instead of landing silently.
- Two independent, complementary checks back the guard:
  - **Runtime probe:** a fresh subprocess imports the `vectorcypher` package plus each submodule and asserts no `khora.engines.skeleton*` key appears in `sys.modules`, naming any offenders. This catches transitive/lazy re-coupling at import time.
  - **Static AST scan:** every `.py` under the `vectorcypher` package is parsed and checked for any `import` / `from ... import` (including relative and `TYPE_CHECKING`-guarded) that resolves to `khora.engines.skeleton`, failing with the exact file, line, and offending module. This catches direct source imports even when guarded behind `TYPE_CHECKING`.
- Test-only change: no production code is modified. It is deliberately an **amendable guideline tripwire**, not a hard import-linter contract — a developer may consciously relax it (with justification in the same diff) if the isolation rule itself changes. No new dependency or config is added (pure stdlib `ast` / `subprocess`).

## Test plan
- [x] `tests/unit/test_engine_isolation.py` passes on the current tree (2 passed)
- [x] Negative case verified: reintroducing a `khora.engines.skeleton` import makes **both** tests fail with a clear, offender-naming message, then reverted
- [x] `ruff format` / `ruff check` clean; `make lint` exits 0
- [ ] Full CI suite green

## Summary by CodeRabbit

* **Tests**
  * Added a regression test to ensure the vectorcypher engine and its submodules remain isolated and do not load or reference the forbidden skeleton engine components, validated via both runtime import checking and a static source scan.

### DIFF
--- a/tests/unit/test_engine_isolation.py
+++ b/tests/unit/test_engine_isolation.py
@@ -1,0 +1,142 @@
+"""Regression guard for vectorcypher → skeleton engine isolation.
+
+Asserts that ``khora.engines.vectorcypher`` and all of its submodules import
+nothing from ``khora.engines.skeleton``. Two independent checks back this up:
+a runtime ``sys.modules`` probe in a fresh subprocess, and a static AST scan of
+every source file under the vectorcypher package.
+"""
+
+# AMENDABLE GUIDELINE TRIPWIRE — not a hard import-linter contract.
+#
+# The engine-isolation rule says the vectorcypher engine must stand on its own
+# and must not reach into the skeleton engine's internals: the two engines are
+# independent implementations of the memory-engine protocol, and coupling one
+# to the other erodes that boundary. This test locks that property in so an
+# *accidental* re-coupling (a stray ``from khora.engines.skeleton import ...``
+# added during unrelated work) trips loudly in CI rather than landing silently.
+#
+# A developer MAY consciously relax this guard if the isolation rule itself
+# changes — but only deliberately, with a justification in the same diff that
+# edits this file. It is a guideline, not a wall: the point is that the change
+# is *seen*, not that it is forbidden.
+
+from __future__ import annotations
+
+import ast
+import importlib.util
+import subprocess
+import sys
+from pathlib import Path
+
+import pytest
+
+_FORBIDDEN_PREFIX = "khora.engines.skeleton"
+_VECTORCYPHER_PACKAGE = "khora.engines.vectorcypher"
+
+# Submodules imported individually by the runtime probe so that a skeleton
+# import hiding in any one of them (not just the package __init__) is caught.
+# Keep in sync with the package's top-level submodules; the static AST scan
+# (rglob over every .py) is the backstop if this drifts.
+_SUBMODULES = (
+    "engine",
+    "retriever",
+    "dual_nodes",
+    "fusion",
+    "ppr_retrieval",
+    "router",
+    "temporal_detection",
+)
+
+
+def _vectorcypher_dir() -> Path:
+    """Locate the vectorcypher package source directory via its spec."""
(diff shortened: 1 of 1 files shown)


msrdic:

Reviewed across six perspectives (security, performance, correctness, test quality, docs, devil's advocate) plus a dedicated security audit. **Overall: APPROVE** — one HIGH correctness issue was found and fixed during review; all other perspectives clean.

### HIGH (resolved)
- **H1 — [Correctness] Relative-import resolution off-by-one for package `__init__.py`** (`tests/unit/test_engine_isolation.py`). The static AST scan anchored relative-import resolution on the full module name. For non-package modules this happened to be correct, but for a package `__init__.py` (where the module name *is* the package) it was off by one: `from ..skeleton import x` in `vectorcypher/__init__.py` resolved to `khora.skeleton` instead of `khora.engines.skeleton`, so a skeleton import added there would not be flagged (false negative). **Fixed** in the follow-up commit: resolution now uses the importing module's *package* (`_package_for`) with `package.rsplit(".", level - 1)[0]`, matching `importlib._bootstrap._resolve_name`. Verified against an importlib oracle on package/module/nested cases, and a new negative check confirms `from ..skeleton` in `__init__.py` now correctly fails the scan, naming the offender. (The runtime `sys.modules` probe already covered this case, so the merged guard is defense-in-depth.)

### MEDIUM (acknowledged — out of scope for this guard)
- **M1 — [Devil's Advocate] Transitive coupling via a third module** is not detected by either check. This is a coordinator-level concern, not a `vectorcypher`→`skeleton` boundary violation, and belongs in a separate follow-up if desired.
- **M2 — [Devil's Advocate] The "amendable guideline" framing relies on reviewer diligence** — a developer can relax the guard with a weak justification. This is intentional by design: the rule is a visible tripwire, not an immutable contract; both the import and the test edit surface in the diff.

### LOW (acknowledged)
- **L1 — [Devil's Advocate] Deliberate dynamic obfuscation** (`importlib.import_module("…skeleton")` inside an uncalled function) can evade both checks. Accidental re-coupling — the stated target — is virtually always a plain `import` statement, which is caught. Determined evasion is a code-review concern.

### Clean perspectives
- **Security (reviewer + dedicated audit): PASS.** The only sinks are `subprocess.run([sys.executable, "-c", <fully static script>])` (`# noqa: S603` justified — no untrusted input) and `ast.parse(path.read_text())` over package files located via `find_spec`/`rglob` (no path traversal). No secrets, network, deserialization, or new dependencies.
- **Performance: APPROVE.** One short-lived subprocess import + an AST walk over a single package; negligible CI cost.
- **Test quality: APPROVE.** Deterministic, well-named, mirrors the existing `test_import_surface.py` subprocess+AST harness, and is non-tautological (genuinely fails on reintroduced coupling).
- **Docs / OSS hygiene: APPROVE.** Docstring and amendable-tripwire comment are clear; no ticket IDs, ADR numbers, or internal identifiers anywhere in the file, commit messages, or PR text.

**Assessment: APPROVE** — the HIGH finding is fixed and verified; remaining items are documented, out-of-scope limitations inherent to an amendable accidental-recoupling tripwire.

probablyrobot (approved):

PR #1356, opened 2026-06-23
Add tests for temporal-store storage relocation

msrdic:
## Summary
- Add regression coverage for the recent relocation of the temporal vector-store layer (the `TemporalVectorStore` protocol, the `create_temporal_store` factory, and the per-backend implementations) into `khora.storage.temporal`.
- `tests/unit/test_backend_shims.py` — the legacy `khora.engines.skeleton.backends.*` submodule paths still resolve, emit `DeprecationWarning`, and re-export classes that are object-identical (`is`) to their new home in `khora.storage.temporal`. The warning is asserted via a clean subprocess under `-W error::DeprecationWarning` (module caching makes an in-process second import unreliable), with an in-process identity check.
- `tests/unit/test_temporal_no_optional_dep_leak.py` — importing `khora.storage.temporal` pulls in no optional backend dependencies (`pyarrow`, `lancedb`); verified in a fresh subprocess so the parent interpreter's already-imported modules cannot mask a leak.
- `tests/integration/test_temporal_store_resource_reuse.py` — `StorageCoordinator.temporal_store()` reuses the coordinator's shared engine/connection/handle (asserted by object identity, not a returned-store smoke check) rather than opening a second one, and returns a fresh, uncached, connected store per call. Backends without available infrastructure skip loudly with a clear reason rather than passing vacuously.

These complement the existing engine-isolation regression guard (`tests/unit/test_engine_isolation.py`) and the embedded temporal end-to-end lane, which were both confirmed to still pass unchanged.

## Test plan
- [x] New unit tests pass (`test_backend_shims.py`, `test_temporal_no_optional_dep_leak.py`)
- [x] Engine-isolation guard passes unchanged
- [x] Embedded temporal e2e lane passes unchanged
- [x] `make format` clean, `make lint` clean
- [ ] Integration test runs against live backends in CI where infra is available; skips loudly otherwise

## Summary by CodeRabbit

* **Tests**
  * Added integration coverage confirming that repeated temporal-store requests reuse existing backend resources (shared SQLite/LanceDB handle, shared PostgreSQL engine, shared SurrealDB connection) while still producing distinct store instances.
  * Added unit tests to verify relocated legacy shims emit deprecation warnings on import and re-export the same symbols as the new locations.
  * Added unit tests ensuring `khora.storage.temporal` avoids eagerly importing optional backend SDKs.

### DIFF
--- a/tests/integration/test_temporal_store_resource_reuse.py
+++ b/tests/integration/test_temporal_store_resource_reuse.py
@@ -203,6 +203,7 @@ class TestSurrealDBResourceReuse:
         await conn.connect()
         relational = SurrealDBRelationalAdapter(conn)
         coord = StorageCoordinator(relational=relational, vector=None)
+        await coord.connect()
         try:
             assert coord._relational._conn is conn
 
--- a/tests/integration/test_temporal_store_resource_reuse.py
+++ b/tests/integration/test_temporal_store_resource_reuse.py
@@ -1,0 +1,220 @@
+"""Behavioral guard: ``StorageCoordinator.temporal_store`` reuses shared resources.
+
+The temporal-store relocation keeps ``StorageCoordinator.temporal_store`` a thin
+factory whose whole job is to hand the new store the coordinator's *existing*
+connection so the two never fork a second handle / engine / connection. The unit
+tests in ``tests/unit/test_coordinator_temporal_store.py`` pin the forwarded
+kwargs with mocks; these tests prove the real thing on a live stack: the store
+the coordinator returns holds the SAME underlying resource object the
+coordinator already owns (``is``-identity), and two calls return DISTINCT store
+objects that nonetheless share that one resource.
+
+The embedded sqlite_lance path runs unconditionally (no external service). The
+pgvector path is gated on Postgres reachability and the surrealdb path on its
+optional SDK; both skip loudly when their backend is absent rather than passing
+silently.
+"""
+
+from __future__ import annotations
+
+import os
+import socket
+from pathlib import Path
+from urllib.parse import urlparse
+
+import pytest
+
+from khora.config import KhoraConfig
+from tests.integration._sqlite_lance_fixtures import build_sqlite_lance_coordinator
+
+try:
+    import aiosqlite  # noqa: F401
+    import lancedb  # noqa: F401
+
+    _HAS_EMBEDDED = True
+except ImportError:
+    _HAS_EMBEDDED = False
+
+try:
+    import surrealdb  # noqa: F401
+
+    _HAS_SURREALDB = True
+except ImportError:
+    _HAS_SURREALDB = False
+
+
+_DATABASE_URL = os.environ.get(
+    "KHORA_DATABASE_URL",
+    "postgresql+asyncpg://khora:khora@localhost:5434/khora",
+)
+if _DATABASE_URL.startswith("postgresql://"):
+    _DATABASE_URL = _DATABASE_URL.replace("postgresql://", "postgresql+asyncpg://", 1)
+elif _DATABASE_URL.startswith("postgres://"):
+    _DATABASE_URL = _DATABASE_URL.replace("postgres://", "postgresql+asyncpg://", 1)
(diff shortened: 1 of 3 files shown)


msrdic:

Six reviewers examined this test-only change across security, performance, correctness, test-quality, docs, and an adversarial (devil's-advocate) lens. The diff adds three test files guarding the temporal vector-store relocation (protocol + factory + backend impls moved into `khora.storage.temporal`, with legacy submodule shims). Overall assessment: **APPROVE.** No CRITICAL or surviving HIGH findings; the one valid rigor gap was addressed in-branch.

### Reviewer verdicts
- **Security** — CLEAN. Subprocess calls use arg-lists (no shell injection), module names are hardcoded, no secrets in fixtures, no eval/exec/pickle. Reachability check uses `urlparse` + timed socket. (CI `security` check also green.)
- **Performance** — APPROVE. The per-shim subprocesses are load-bearing (module caching makes an in-process re-warn impossible) and the single clean-interpreter subprocess is necessary for an unpolluted `sys.modules`. 60s timeouts, no unbounded waits; integration tests use the embedded fixture / in-memory mode / reachability-gated live infra.
- **Docs** — PASS. Excellent self-documenting docstrings explaining *why* subprocess isolation is used and the lazy-vs-transitive dep boundary. No internal identifiers leaked.
- **Test-quality** — PASS. All five intended scenarios covered; gated backends skip loudly; clean async/cleanup discipline. (Marker note below.)
- **Correctness / Devil's-advocate** — findings adjudicated below.

### Findings

**M1 (Correctness/Adversarial) — lazy-import guard could pass vacuously — ADDRESSED.**
`test_temporal_no_optional_dep_leak.py` asserted `lancedb`/`weaviate`/`turbopuffer` are absent after import, but an SDK that isn't installed can never appear regardless of lazy-vs-eager imports, so those assertions risk being vacuously true — the "silently passed where infra is absent" anti-pattern. Fixed in `670b4704`: the test now skips loudly when the headline dep (`lancedb`) is absent (guard would be wholly vacuous) and prints the installed-vs-absent split so CI logs show which assertions were meaningful. (In this env/CI all three SDKs are installed, so the assertions do bite.) Note an *eager* impl-module import of an absent SDK was already caught — it raises `ImportError` at package import and trips the non-zero-exit assertion.

**L1 (Test) — new unit files omit `@pytest.mark.unit`.**
Considered and intentionally left as-is: the marker is applied in only ~44% of `tests/unit/` files and is omitted by the two most-relevant peers — the engine-isolation regression guard and the sibling coordinator temporal-store test. There is no selective marker-filtered unit lane that would skip them (the full suite collects and runs them — `test-unit` CI lane green). Matching the immediate siblings is the locally-consistent, surgical choice.

**L2 (Adversarial) — shim identity test theoretical cleanup gap — dismissed.**
The concern (a cached re-import hides a missing warning and `pytest.warns` passes silently) does not hold: `pytest.warns` raises `DID NOT WARN` if no matching warning fires, and `khora.storage.temporal` imports nothing from the shims (the engine-isolation rule), so importing the new home cannot re-cache a shim. The warning is *independently* guaranteed by the separate `-W error::DeprecationWarning` subprocess test.

**L3 (Adversarial) — resource-reuse test uses `is` on a private `_handle`.**
This is precisely the reuse assertion the change set out to make (identity proves no second handle/engine/conn was opened), and the reviewer concedes it correctly detects the bug. A functional round-trip could be added as future hardening, but the identity check is the intended contract.

**L4 (Test) — surrealdb skip reason could be more explicit** about optional-dependency vs. unavailable-service. Minor wording.

### CI
All required checks green on the reviewed commits — including `vc_embedded` (behavior-parity lane), `test-unit`, `test-integration`, the full `conformance`/`skeleton` matrix, `lint`, `security`, and `examples-smoke`.

_Security audit (OWASP / deps / crypto): nothing applicable to a test-only diff that adds no runtime code paths, network calls, or dependencies; Security reviewer CLEAN and CI `security` check green._

sale87 (approved):

PR #1403, opened 2026-07-01
[codex] Reuse VectorCypher extraction state

Arsenije:
## Summary

This PR caches one shared LLMEntityExtractor per effective extraction model inside VectorCypherEngine and reuses it for skeleton extraction. This keeps semaphore and circuit-breaker state across staged/background document indexing instead of resetting extraction state for every document. The root cause was that the VectorCypher staged path already reused the embedder but still constructed fresh LLM extraction state through extract_entities unless an explicit shared extractor was provided. Validation: uv run pytest tests/unit/engines/test_vectorcypher_engine.py --no-cov -q; uv run pytest tests/unit/test_extraction_llm.py --no-cov -q.


Arsenije:
Closing because this work is being moved to https://github.com/Arsenije/Uru.

Issue #857, opened 2026-05-28
`Khora.recall()` returns empty `entities` / `relationships` on VectorCypher + `sqlite_lance`, while the same call on `postgres+neo4j` correctly surfaces them


damirkrstanovic:
## Summary

On the embedded `sqlite_lance` backend with the **vectorcypher**
engine, `Khora.recall(query, namespace=ns)` returns `chunks` correctly
but `result.entities == []` and `result.relationships == []`. The same
call against the production `postgres+neo4j` stack returns chunks +
entities + relationships inline.

The embedded graph IS populated: `kb.list_entities(...)` returns the
extracted nodes and `kb.find_related_entities(...)` traverses edges
on `sqlite_lance` exactly as it does on `postgres+neo4j`. Only the
`recall()` projection path is broken on the embedded backend.

## Environment

- **Khora**: v0.17.2
- **Storage backends tested**: `sqlite_lance` (embedded) — bug
  reproduces. `postgres+neo4j` (production stack) — works correctly.
- **Engine**: `vectorcypher`
- **Python**: 3.13.11
- **OS**: macOS 25.5.0 (Darwin)

## Reproduction

```python
"""Repro: VectorCypher recall returns empty entities on sqlite_lance, full on pg+neo4j."""
import asyncio
import os
import tempfile
from pathlib import Path

from khora import Khora
from khora.config import KhoraConfig

_PG = os.environ.get("KHORA_DATABASE_URL", "postgresql://khora:khora@localhost:5434/khora")
_NEO = os.environ.get("KHORA_NEO4J_URL", "bolt://neo4j:pleaseletmein@localhost:7688")
_LLM = {"model": "gpt-4o-mini", "api_key_env": "OPENAI_API_KEY",
        "embedding_model": "text-embedding-3-small", "embedding_dimension": 1536}


def sqlite_lance_config(tmpdir):
    return KhoraConfig.model_validate({
        "storage": {"backend": "sqlite_lance", "sqlite_lance": {
            "db_path": str(tmpdir / "khora.db"), "embedding_dimension": 1536}},
        "llm": _LLM,
    })


def postgres_neo4j_config():
    return KhoraConfig.model_validate({
        "storage": {"backend": "postgres", "embedding_dimension": 1536},
        "database_url": _PG, "neo4j_url": _NEO, "llm": _LLM,
    })


async def probe(label, config):
    async with Khora(config, engine="vectorcypher", run_migrations=True) as kb:
[...]

PR #1454, opened 2026-07-07
Add between_entity_ids filter to list_relationships across storage backends

### LINKED ISSUE #1451: list_relationships: between_entity_ids filter (push #857 recall relationship projection into the store)
The #857 simple-path projection in VectorCypher recall fetches up to 1,000 relationships for the whole namespace (`list_relationships(namespace_id, limit=1000)`) and keeps, in Python, only edges whose both endpoints are in the matched entity set. This is wasteful and unbounded relative to the recalled set.

Companion to #1449 (the entity half — `source_chunk_ids` overlap filter on `list_entities`). Same decision: no new protocol methods — add a keyword-only `between_entity_ids: list[UUID] | None = None` filter to the existing `list_relationships`.

Semantics:
- `None` (default): no filter (byte-identical to today).
- Non-empty list: return only relationships whose source AND target entity ids are both in the list (both-endpoints-in-set).
- Empty list `[]`: return `[]`.
- Composes with AND against existing conditions; applied before limit/offset.

Pushes the #857 relationship projection down into every storage backend (Neo4j, Memgraph, pgvector, sqlite_lance, SurrealDB, Neptune, AGE) + coordinator + protocol, and updates the retriever call site to drop the Python-side endpoint filter.

filippetrovic:
## Summary
- Add a keyword-only `between_entity_ids: list[UUID] | None = None` filter to `list_relationships` across every storage backend (Neo4j, Memgraph, pgvector, SQLite/LanceDB, SurrealDB, Neptune, AGE), the storage protocol, and the coordinator. When a non-empty list is supplied, only relationships whose **source and target entity ids are both in the list** are returned (both-endpoints-in-set). `None` (default) is byte-identical to prior behavior; `[]` returns `[]`. The filter composes with existing conditions and is applied before `limit`/`offset`.
- Push the simple-path relationship projection down into the store: the VectorCypher retriever now passes the recalled entity ids as `between_entity_ids` instead of fetching the namespace-wide relationship list and filtering endpoints in Python. This is the companion to #1449 (the entity half, `source_chunk_ids` on `list_entities`) — with both halves now pushed into the backend, the projection no longer pulls the whole namespace over the wire.
- Graph backends (Neo4j/Memgraph) push the predicate into the Cypher `WHERE`, binding the id list only when the clause is active; pgvector/SQLite/SurrealDB push it into the query; Neptune/AGE filter in-method (openCypher subset support unverified, no CI lane).
- Add backend behavior tests (sqlite_lance, pgvector, Neo4j, SurrealDB) and retriever unit tests pinning the pushdown contract (no-filter → all edges, `[A,B]` → only A→B, `[A]` → `[]`, `[]` → `[]`).

Fixes #1451

## Test plan
- [x] Unit tests pass (retriever simple-path projection + expansion-flow caller)
- [x] Embedded backend behavior tests pass locally (sqlite_lance, SurrealDB memory mode)
- [ ] Integration tests pass in CI (pgvector, Neo4j — require Docker)
- [x] `ruff format --check`, `ruff check`, and `ty` clean; `check_optional_imports` clean

## Summary by CodeRabbit

* **New Features**
  * Relationship lookups can now be limited to connections between a specific set of entities using an optional entity-set filter.
  * Results now only include edges where **both endpoints** are in the selected set; explicitly empty sets return no relationships.
* **Bug Fixes**
  * Improved cross-backend consistency by pushing the entity-set constraint into backend queries and avoiding redundant post-filtering.
* **Tests**
  * Added/updated integration and unit tests to verify the new filter behavior (including relationship-type compatibility and empty-set cases).

### DIFF
--- a/src/khora/engines/vectorcypher/retriever.py
+++ b/src/khora/engines/vectorcypher/retriever.py
@@ -3345,15 +3345,11 @@ class VectorCypherRetriever:
             # return value hardcoded ``entities=[]`` and ``relationships=[]``,
             # so backends without a Neo4j driver (sqlite_lance, surrealdb,
             # postgres-only) surfaced empty entity lists from ``recall()``
-            # even when the graph was populated. Entities are pushed down via
-            # ``source_chunk_ids`` overlap (#1448); relationships still fetch
-            # the namespace-wide list (capped at 1000) and filter by overlap
-            # with the recalled chunk ids in Python.
-            # TODO: replace the namespace-wide relationship list + Python
-            #       filter with a per-backend ``list_relationships_by_chunk_ids``
-            #       query method once it lands across sqlite_lance / surrealdb /
-            #       pgvector (perf follow-up to #857). The entity half is now
-            #       pushed down via ``list_entities(source_chunk_ids=...)``.
+            # even when the graph was populated. Both halves are now pushed
+            # down: entities via ``list_entities(source_chunk_ids=...)`` overlap
+            # (#1448), relationships via ``list_relationships(between_entity_ids=...)``
+            # so only edges whose BOTH endpoints are in the recalled entity set
+            # come back (#1451) — no namespace-wide list + Python filter.
             entities_with_scores: list[tuple[Entity, float]] = []
             relationships_with_scores: list[tuple[Relationship, float]] = []
             if chunk_results and self._storage is not None:
@@ -3380,20 +3376,20 @@ class VectorCypherRetriever:
                         entities_with_scores.append((entity, score))
 
                 if entities_with_scores:
+                    matched_entity_ids = [e.id for e, _ in entities_with_scores]
                     with trace_span(
                         "khora.vectorcypher.simple_projection.list_relationships",
                         namespace_id=str(namespace_id),
                     ) as rel_span:
                         try:
-                            all_relationships = await self._storage.list_relationships(namespace_id, limit=1000)
+                            all_relationships = await self._storage.list_relationships(
+                                namespace_id, limit=1000, between_entity_ids=matched_entity_ids
+                            )
                         except Exception as e:
                             logger.warning(f"#857 simple-path relationship projection failed: {e}")
                             all_relationships = []
                         rel_span.set_attribute("relationship_count", len(all_relationships))
-                    recalled_entity_ids = {e.id for e, _ in entities_with_scores}
-                    for rel in all_relationships:
-                        if rel.source_entity_id in recalled_entity_ids and rel.target_entity_id in recalled_entity_ids:
-                            relationships_with_scores.append((rel, 1.0))
+                    relationships_with_scores = [(rel, 1.0) for rel in all_relationships]
 
             # Honest per-channel filter-pushdown plans for the simple path. Each
             # plan is the one its channel appended to the sink from the BACKEND's
--- a/src/khora/storage/backends/age.py
+++ b/src/khora/storage/backends/age.py
@@ -792,21 +792,31 @@ class AGEBackend(GraphBackendBase):
         namespace_id: UUID,
         *,
         relationship_type: str | None = None,
+        between_entity_ids: list[UUID] | None = None,
         limit: int = 1000,
         offset: int = 0,
     ) -> list[Relationship]:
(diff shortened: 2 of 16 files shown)


filippetrovic:
## 🔍 Multi-Perspective Review — PR #1454

Reviewed by a 7-role team (security, performance, correctness, test, docs, database, devil's advocate) plus a focused security audit. This change adds a `between_entity_ids` filter to `list_relationships` across all storage backends, mirroring the merged `source_chunk_ids` filter on `list_entities` (#1449) and pushing the #857 simple-path relationship projection into the store.

**Overall assessment: ✅ APPROVE** — correct, secure, performant, and well-tested; faithfully mirrors the approved #1449 pattern. No blocking findings. Two raises came in at HIGH/CRITICAL and were adjudicated down (details below).

### CRITICAL
_None._

### HIGH
_None confirmed._ Two findings were raised at HIGH/CRITICAL and reclassified:

- **[Correctness/Performance] Neptune & AGE fetch-without-`LIMIT` when `between_entity_ids` is set → reclassified MEDIUM (accepted design).** These two backends deliberately skip query pushdown (openCypher subset support is unverified and they have no CI lane), so the filter runs in-method: fetch → keep rows where both endpoints are in the set → slice `[offset:offset+limit]`. This preserves filter-before-pagination semantics and mirrors the merged sibling #1449 exactly. A pre-filter cap was explicitly evaluated and rejected upstream because it reintroduces an arbitrary-truncation correctness bug. In practice the id list is bounded by the caller (`list_entities(limit=1000)`, and the recalled-entity cardinality is typically far below that). No change required.
- **[Database] sqlite_lance "double-bind" / "pagination after filtering" → not a defect.** The two `params.extend(id_texts)` calls are required by positional `?` binding (each placeholder needs its own value in order — they cannot be reused). Bind count is bounded (≤ 2,000, far under SQLite's 32,766 ceiling). Filter-then-slice preserves the filter-before-`limit`/`offset` contract.

### MEDIUM
- **M1 [Performance] Neptune/AGE unbounded in-method fetch** — accepted tradeoff (see above). *Optional follow-up:* add a `matched_entity_ids` cardinality attribute to the projection span/metric to detect if the bounded-caller assumption ever breaks.
- **M2 [Docs] Prose docs not updated** — the protocol docstring and all in-code comments are accurate and complete; `docs/architecture/storage-backends.md` / `docs/api-reference.md` don't yet mention the new filter. Out of scope for this change (the sibling #1449 merged without it); *optional follow-up doc PR.*
- **M3 [Devil's Advocate] Empty-list guard duplicated across backends** — `if between_entity_ids is not None and not between_entity_ids: return []` appears in each backend. This is intentional (defense-in-depth: backends can be called directly, not only via the coordinator) and matches #1449. *Optional:* add a coordinator-level short-circuit in addition.

### LOW
- **L1 [Database/Performance] Index assumption** — pgvector filters on `source_entity_id`/`target_entity_id`; entity `id` is the primary key on the graph backends (already indexed). Confirm the pgvector columns are covered; sibling index work is tracked separately.
- **L2 [Devil's Advocate] Cross-backend binding style differs** (conditional params dict vs chained `.where()`) — not a bug; watch for drift.
- **L3 [Test] Optional test strengthening** — explicit self-loop case, `between_entity_ids` + `relationship_type` composition, and mocking an out-of-set edge in the retriever verbatim test. Current coverage (4 contract cases × 4 CI backends, exact set-equality assertions) is strong.

### 🔒 Security Audit
- **Injection:** N/A — every backend binds ids as query parameters (Cypher `$between_entity_ids`, SQLAlchemy `.in_()`, positional `?`, SurrealDB `$between_eids`); no string interpolation of values.
- **Sensitive data / secrets:** none introduced; no internal identifiers leaked.
- **Input trust:** entity ids originate from the trusted storage layer (recalled entities), not user free-text.
- **Dependencies / crypto / authz:** unchanged.
- **Verdict:** CLEAN.

### Verdict
**APPROVE.** No blocking findings. All MEDIUM/LOW items are optional follow-ups; the two HIGH/CRITICAL raises were adjudicated as an already-decided, documented design tradeoff (mirrors merged #1449) and a non-defect. Merge-ready pending CI.

probablyrobot (approved):

Issue #1457, opened 2026-07-09
Filtered recall: entities/relationships bypass the caller filter on the VectorCypher graph path


filippetrovic:
## Summary

`RecallResult.entities` and `relationships` bypass the caller `filter=` on the VectorCypher graph path. A date-keyed filter (`occurred_at`/`created_at`) deterministically forces that path, so every date-filtered recall is affected. Verified by repro on v0.21.1 and current main (v0.22.2), embedded sqlite_lance, no external services.

## Reproduction

Two documents ingested (5 entities extracted), embedded `sqlite_lance`, `engine="vectorcypher"`:

```python
r = await kb.recall("pagination fix", namespace=ns,
                    filter={"occurred_at": {"$gte": "2027-01-01T00:00:00Z"}})
# chunks == []           — correct, nothing matches
# entities == [2 items]  — LEAK: entities returned despite 0 matching chunks
# relationships == [2]   — LEAK
# documents == [1 doc]   — the leaked entities pull their (filter-excluded) doc into documents

r = await kb.recall("pagination fix", namespace=ns, filter={"source": "github"})
# chunks == [] and entities == []  — consistent, because this routes to the simple path
```

A deterministic pytest repro (stub embedder/extractor, `uv run --extra sqlite-lance pytest -s --no-cov`) accompanies the fix PR.

## Mechanism (anchors on current main)

1. `filter_constrains_date_key` synthesizes `TemporalSignal(category=EXPLICIT, source="api")` — `engines/vectorcypher/engine.py:2401-2407`.
2. The router boosts any temporal-flagged query to MODERATE → `use_graph=True` → `_vectorcypher_retrieve` (`query/router.py`, "temporal signal boosted to MODERATE").
3. On the graph path, `_vector_search_entities` (`engines/vectorcypher/retriever.py:1547`) receives no `filter_ast`; the full-AST in-memory post-filter re-checks the **chunk** channel only; `entity_results` (entry entities + graph expansion) is returned unfiltered and projected 1:1 into `RecallResult.entities`.
4. The simple path is consistent only because its #857 projection derives entities from already-filtered chunk ids.

`engine_info["filter"].unenforced_keys` stays `[]` throughout: the report universe is channel self-registration, and the entity projection never registers a `ChannelPlan` — so the honest-report gate cannot see this defect class.

The score-collapse symptom originally reported alongside this (RRF-scale chunk scores under date filters) was fixed by #1433 in v0.22.2 and is **not** part of this issue.

[...]

Issue #1458, opened 2026-07-09
Chronicle recall: entities bypass the caller filter (entity channel receives no filter_ast)


filippetrovic:
Sibling of #1457, discovered by source trace while verifying that issue's engine scope. Not yet reproduced empirically; the mechanism is identical.

## Mechanism (anchors on current main)

- `_entity_channel` finds query-similar entities via `storage.search_similar_entities(namespace_id, query_embedding, limit=10)` — no `filter_ast` (`engines/chronicle/engine.py:2856`).
- Those entities are recorded into `entity_hits` and surfaced 1:1 as `RecallResult.entities` (`engine.py:2299-2311`, `:2396`). No provenance/filter check between the entity search and the result.
- The chunk post-filter ("the full filter is always enforced by the in-memory post-filter") covers chunks only; the filter report registers a single `"chunks"` ChannelPlan (`engine.py:2376-2390`), so `unenforced_keys == []` while entities bypass the filter.
- Entity-referenced documents are appended to `documents[]` via the producer invariant (`engine.py:2335-2347`), so filter-excluded documents ride along, same as #1457.

Skeleton is NOT affected: it always returns `entities=[]` (`engines/skeleton/engine.py:647`).

## Contract

Same as #1457 (ADR-092 result-surface amendment): entity filter compliance is ∃-over-provenance — a returned entity must have ≥1 provenance chunk (`source_chunk_ids`) satisfying the filter.

## Fix

Reuse the shared ∃-over-provenance helper introduced by #1457's fix PR: apply it to `entity_hits` before `recall_entities` assembly under `filter_ast is not None`, register an `entities` ChannelPlan, fail-closed with a `Degradation` when provenance cannot be fetched (`storage.get_chunks_batch`, `coordinator.py:1344`). Until then, the surface-coverage report rule from #1457's first PR will (correctly) flag Chronicle filtered recalls with entities as `unenforced_keys ≠ []`.

PR #1462, opened 2026-07-09
Compute filter report unenforced_keys from result surfaces

filippetrovic:
## Summary
- `build_filter_report` gains optional keyword-only `surface_sizes` / `covered_surfaces` params: when a result surface (chunks/entities/relationships) comes back non-empty but is **not** covered by the filter's channels, every constraint leaf of the filter is forced into `unenforced_keys` (out of `pushed_keys` / `post_filtered_keys`) and `pushed_down` is cleared — keeping the three lists a total, disjoint partition. With `surface_sizes=None` the builder is byte-for-byte its legacy self, so existing callers are unaffected.
- The VectorCypher, Chronicle, and Skeleton engines now pass their result-surface sizes. The VectorCypher **graph path** and Chronicle's **entity channel** honestly report the entity/relationship surface as *unenforced* when the caller filter is only applied to the chunk channel — so a filter that leaks unfiltered entities is now visible in `engine_info["filter"].unenforced_keys` instead of silently reporting `[]`.
- This is the **report-completeness** step: it makes the leak visible (red) via the honest report. The actual entity/relationship surface filtering is a follow-up; the simple path is already compliant by derivation (its entities come from already-filtered chunks) and stays green.

## Test plan
- [x] Unit tests: builder surface-coverage rule (uncovered non-empty → all leaves unenforced; covered → legacy; `surface_sizes=None` → byte-for-byte legacy; empty surface → inert; disjointness)
- [x] Graph-path and Chronicle entity-bearing leak pinned as `@pytest.mark.xfail(strict=True)` (they encode the clean report the follow-up fix restores → flip to xpass then); simple-path asserts `unenforced_keys == []` (coverage-by-derivation) and stays green
- [x] `make format` / `make lint` clean; `make test-unit` green for the affected suites (0 XPASS)
- [ ] Integration + e2e lanes (CI: live Postgres/Neo4j) — includes the cross-engine filter-report invariant gate

Refs #1457

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved filter-report accuracy by explicitly tracking which result surfaces (chunks/entities/relationships) are covered across Chronicle, Skeleton, and VectorCypher recall modes.
  * Prevented filter “leaks” by disabling filter enforcement when a required surface is uncovered, keeping filter-pushdown invariants consistent.

* **Tests**
  * Added regression coverage for entity-surface leak scenarios under date filters (including graph-path cases), with affected cases marked as expected failures until the fix fully resolves.

### DIFF
--- a/src/khora/engines/chronicle/engine.py
+++ b/src/khora/engines/chronicle/engine.py
@@ -2386,10 +2386,11 @@ class ChronicleEngine:
                     defensive_recheck=(filter_ast is not None and bool(filter_ast.children)),
                 )
             },
-            # Chronicle is a single "chunks" channel with no relationships; the
-            # entity surface is co-derived from the same chunk-side filter, so
-            # only "chunks" is covered and the rule is inert unless entities are
-            # ever emitted uncovered.
+            # Chronicle covers only the "chunks" surface: the entity channel
+            # (search_similar_entities) receives no filter_ast (#1458), so a HYBRID
+            # recall that surfaces entities emits an UNCOVERED entity surface and the
+            # builder honestly forces the filter's leaves into unenforced_keys. The
+            # #1458 fix filters the entity surface and adds it to covered_surfaces.
             surface_sizes={
                 "chunks": len(recall_chunks),
                 "entities": len(recall_entities),
--- a/src/khora/engines/chronicle/engine.py
+++ b/src/khora/engines/chronicle/engine.py
@@ -2386,6 +2386,16 @@ class ChronicleEngine:
                     defensive_recheck=(filter_ast is not None and bool(filter_ast.children)),
                 )
             },
+            # Chronicle is a single "chunks" channel with no relationships; the
+            # entity surface is co-derived from the same chunk-side filter, so
+            # only "chunks" is covered and the rule is inert unless entities are
+            # ever emitted uncovered.
+            surface_sizes={
+                "chunks": len(recall_chunks),
+                "entities": len(recall_entities),
+                "relationships": 0,
+            },
+            covered_surfaces={"chunks"},
         )
 
         return RecallResult(
--- a/src/khora/engines/skeleton/engine.py
+++ b/src/khora/engines/skeleton/engine.py
@@ -637,6 +637,14 @@ class SkeletonConstructionEngine:
         engine_info["filter"] = build_filter_report(
             filter_ast,
             {self._backend_type: backend_plan},
+            # Skeleton emits chunks only (no entities / relationships), and the
+            # single backend channel gates the chunk surface, so covered_surfaces
+            # defaults to {"chunks"} and the surface-coverage rule is inert here.
+            surface_sizes={
+                "chunks": len(recall_chunks),
+                "entities": 0,
+                "relationships": 0,
+            },
         ).model_dump(mode="json")
 
         return RecallResult(
--- a/src/khora/engines/vectorcypher/engine.py
+++ b/src/khora/engines/vectorcypher/engine.py
@@ -2658,6 +2658,18 @@ class VectorCypherEngine:
 
         filter_channel_plans = result.metadata.pop("_filter_channel_plans", {})
 
+        # Surface coverage for the honest filter report: the filter channels gate
+        # the chunk surface on every recall. Only "simple_" search modes (vector /
+        # vector+bm25, no graph path) return entities / relationships that the same
+        # chunk-side filter narrowed; a graph-path recall emits graph-derived
+        # entities / relationships the chunk filter never touched, so those
+        # surfaces stay uncovered and force the filter's leaves unenforced.
+        covered = {"chunks"} | (
+            {"entities", "relationships"}
+            if str(result.metadata.get("search_mode", "")).startswith("simple_")
+            else set()
+        )
+
         # Project materialized dream communities (#1276) onto the result (#1308):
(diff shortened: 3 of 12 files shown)


filippetrovic:
# Multi-perspective review — automated team + security audit

Six independent reviewers (security, performance, correctness, tests, docs, devil's advocate) plus a dedicated security audit.

**Overall: ✅ APPROVE.** No CRITICAL/HIGH blockers. The change is correct (partition invariant holds in every branch), secure (no findings), negligible performance cost, and well-tested (11 xfailed / **0 XPASS** in the affected suites). Everything below is non-blocking polish or a design note.

## Blocking findings
None.

## MEDIUM
- **M1 · Tests** — `tests/unit/filter/test_report.py`: no test exercises **two uncovered surfaces at once** (e.g. `surface_sizes={"chunks":2,"entities":3,"relationships":1}`, `covered={"chunks"}`). The `any(...)` rule at `report.py:246` handles it, but the multi-surface path deserves an explicit test. (Test reviewer flagged HIGH; downgraded — "code handles it correctly, not a blocker.")
- **M2 · Tests** — same file: no test for a would-be **post_filtered** leaf forced unenforced by an uncovered surface (only the would-be-*pushed* case is covered). `report.py:253-254` removes from both sets; an explicit test would lock that interaction.
- **M3 · Docs** — `src/khora/filter/report.py`: the `build_filter_report` / `FilterPushdownReport` / `unenforced_keys` docstrings could state explicitly that **all** filter leaves (not only surface-relevant ones) are forced unenforced, and why (the filter is a holistic constraint on the result set).

## LOW
- **L1 · Correctness (comment accuracy)** — `src/khora/engines/chronicle/engine.py` (~2389): the inline comment framing the entity surface as "co-derived from the chunk-side filter" is misleading — Chronicle's entity channel does **not** receive `filter_ast`, so entities are unfiltered, which is precisely why the surface-coverage rule (correctly) flags them. Comment-only; no behavioral impact. Reword to reflect that the rule is the honest-reporting mechanism here.
- **L2 · Docs** — `FilterPushdownReport` class docstring: the "single-channel engine always reports `[]`" framing is now secondary to the multi-surface case; a one-line note that single-channel engines emit only covered surfaces would help.
- **L3 · Design** — the `covered_surfaces` default `frozenset({"chunks"})` is implicit. It errs **conservative** (an omitted call site flags a *potential* leak rather than hiding one) and matches the intended contract; noted for a possible required-parameter API in a future refactor.

## Design notes (considered, intentionally deferred)
- **search_mode coupling** — coverage keys off `result.metadata["search_mode"].startswith("simple_")`, a string set only in `_simple_retrieve`; the graph path leaves it unset (→ not covered → honest leak signal, which is the intended behavior of this ship). A future retrieval path that emits chunk-derived entities should set `covered_surfaces` explicitly (or an explicit retriever mode) so it isn't mis-flagged. An enum-based mode is a reasonable later refactor, out of scope here.
- **"xfail leaves CI red" — clarification** — the pinned leak tests are `xfail(strict=True)`, i.e. **expected failures that keep CI green**; only an *XPASS* fails the build. The "red" this ship introduces is the leak becoming **visible in `unenforced_keys`**, not a failing pipeline. The sequencing (report-completeness first; the surface-filtering fix flips the xfails to xpass in the #1457/#1458 follow-ups) is deliberate.

## Security Audit
Scanned added/modified lines against all 16 categories. No dependency/lockfile changes; no injection/shell/exec, no HTTP/SSRF/file/deserialize sinks, no secrets or weak crypto, no new Pydantic fields (ADR-084 clean). Pure report arithmetic over integer surface counts with hardcoded surface names — no untrusted-input path.
**Verdict: PASS** — no findings.

## Verdict
**APPROVE.** Correct, secure, performant, and well-tested. MEDIUM/LOW items are optional polish; the design notes are deferred-with-rationale for this report-completeness step (the surface-filtering fix lands in the follow-ups, Refs #1457 · #1458).

filippetrovic:
Addressed the review feedback on the xfail pins (commit cc747934):

- `raises=AssertionError` added to every entity-surface-leak xfail marker; every precondition / anti-vacuity gate (`_filter_channel_plans` absence, `result.entities`, `graph_chunk_count > 0`, non-`simple_` search_mode) converted from `assert` to `pytest.fail` (raises `Failed`, not `AssertionError`). A broken seed/mock/graph channel now surfaces as a real failure instead of being absorbed as an expected XFAIL. Only the clean-report assertions the fix flips (`unenforced_keys == []`, `pushed_keys`, `pushed_down`) remain plain `assert`.
- The nine duplicated markers collapsed into a shared `_XFAIL_1457` constant.
- `test_five_concurrent_recalls_distinct_keys` restored to live coverage: xfail removed, the surface-rule-dependent top-level `pushed_keys` assertion dropped, and the per-call channel-level scoping assertion (unaffected by the surface rule) retained — the sink-isolation signal now runs green.
- Chronicle `build_filter_report` call-site comment corrected: the entity channel receives no `filter_ast` (#1458), so it emits an uncovered entity surface the builder honestly flags — it is not co-derived from the chunk-side filter.

Verification (no-Docker lane): `tests/unit/filter/test_report.py`, `tests/unit/engines/test_vectorcypher_filter_report.py`, `tests/recall/test_chronicle_filter_composition.py` → 77 passed, 10 xfailed, 0 XPASS; the concurrency test passes. The e2e cases are lane-gated and self-skip without live Postgres/Neo4j (CI covers them).

probablyrobot (approved):

PR #1487, opened 2026-07-10
Filter entities by provenance on Chronicle recall

### LINKED ISSUE #1458: Chronicle recall: entities bypass the caller filter (entity channel receives no filter_ast)

filippetrovic:
## Summary

Under a caller `filter_ast`, Chronicle's entity channel (`search_similar_entities`) was assembled from vector similarity the chunk post-filter never touched, so a filter that narrowed chunks could still leak entities it never constrained (the entity surface was reported uncovered, forcing every filter leaf into `unenforced_keys`). This re-applies the same `Chunk` predicate to each entity's provenance chunks (∃-over-provenance), reusing the engine-agnostic `filter_items_by_provenance` helper — mirroring the VectorCypher graph-path fix.

- An entity survives iff ≥1 of its `source_chunk_ids` passes the compiled filter predicate.
- Only runs under a filter with leaves and a non-empty entity set; unfiltered recalls take the zero-cost path (no extra fetch).
- Fail-closed: a provenance-fetch failure drops unverified entities and records one structured degradation (`component="chronicle.entity_filter"`) — every returned entity stays verified.
- The entity/relationship surfaces are marked covered in the honest filter report when the pass runs, so entity-bearing filtered recalls report `unenforced_keys == []`.

Fixes #1458

## Test plan

- [x] Unit — `tests/recall/test_chronicle_filter_composition.py`: flipped the previously-xfailed clean-report pin to green; added partial-match and fail-closed (degradation) cases.
- [x] Conformance oracle — `tests/unit/filter/test_recall_conformance_oracle.py`: added an implementation-blind Chronicle leg asserting the compiled predicate over returned chunks and ∃-over-provenance over returned entities.
- [x] E2E — `tests/e2e/test_filter_report_invariant.py`: re-enabled the entity-bearing chronicle report-invariant pin (previously skipped as vacuous); the entity channel now resolves real entities so the coverage rule is exercised.
- [x] Telemetry contract drift gate updated for the new `chronicle.entity_filter` component label.
- [x] CI: full conformance + e2e engine matrix.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved recall filtering to enforce entity constraints against provenance using an “∃ over provenance” strategy for date and source filters, while preserving original entity scores.
  * Entities are now omitted when provenance doesn’t satisfy the filter; filter reporting only marks entity constraints as enforced when the provenance pass actually ran.
  * If provenance fetching fails, the system fails closed by dropping affected entities and recording structured degradations.
* **Monitoring**
  * Updated telemetry contract for `khora.filter.provenance.degraded_total` to allow `chronicle.entity_filter` as a component value.
* **Tests**
  * Expanded/updated recall conformance and filter-report invariant coverage for Chronicle and HYBRID entity channels, including new provenance match and fail-closed scenarios.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,10 +6,6 @@ Format: versions match git tags (`git tag vX.Y.Z`). Versions before 0.5.1 were i
 
 ## [Unreleased]
 
-### Fixed
-
-- **Chronicle filtered recalls enforce `occurred_at` consistently across the chunk and entity surfaces** (#1458, #1487): the `∃`-over-provenance entity filter now evaluates each provenance chunk through the same `COALESCE(occurred_at, source_timestamp)` view chronicle's chunk post-filter uses, so an `occurred_at` predicate the chunk channel satisfies no longer false-drops the provenance entity whose chunk carries its event time only in `source_timestamp` (chronicle's shape — the raw `occurred_at` column is NULL). VectorCypher is unaffected (its chunks carry `occurred_at` natively; the shared `filter_items_by_provenance` helper defaults to the raw-chunk view).
-
 ## [0.22.2] - recall-quality correctness, HNSW index-backed search, durable reconcilers
 
 ### Added
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@ Format: versions match git tags (`git tag vX.Y.Z`). Versions before 0.5.1 were i
 
 ## [Unreleased]
 
+### Fixed
+
+- **Chronicle filtered recalls enforce `occurred_at` consistently across the chunk and entity surfaces** (#1458, #1487): the `∃`-over-provenance entity filter now evaluates each provenance chunk through the same `COALESCE(occurred_at, source_timestamp)` view chronicle's chunk post-filter uses, so an `occurred_at` predicate the chunk channel satisfies no longer false-drops the provenance entity whose chunk carries its event time only in `source_timestamp` (chronicle's shape — the raw `occurred_at` column is NULL). VectorCypher is unaffected (its chunks carry `occurred_at` natively; the shared `filter_items_by_provenance` helper defaults to the raw-chunk view).
+
 ## [0.22.2] - recall-quality correctness, HNSW index-backed search, durable reconcilers
 
 ### Added
--- a/docs/telemetry-contract.json
+++ b/docs/telemetry-contract.json
@@ -1677,7 +1677,7 @@
       "unit": "1",
       "owner": "filter",
       "labels": [
-        {"name": "component", "values": ["vectorcypher.entity_filter", "vectorcypher.relationship_filter"], "_comment": "Low-cardinality enum: the caller's per-surface value. Chronicle reuses the helper (#1458) and will add chronicle.* values."},
+        {"name": "component", "values": ["vectorcypher.entity_filter", "vectorcypher.relationship_filter", "chronicle.entity_filter"], "_comment": "Low-cardinality enum: the caller's per-surface value. Chronicle reuses the helper (#1458) via chronicle.entity_filter."},
         {"name": "reason", "values": ["provenance_fetch_failed"], "_comment": "Low-cardinality enum; extend when adding new provenance-filter fallback paths."}
       ],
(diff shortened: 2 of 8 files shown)


sale87 (approved):

filippetrovic:

Six independent reviewers examined the diff (correctness, test quality, security/leak-closure, observability & telemetry, performance, and docs/public-hygiene). **Verdict: APPROVE — no CRITICAL / HIGH / MEDIUM / LOW findings.**

### What was verified

**Correctness** — The `∃`-over-provenance wiring is sound: the gate (`filter_ast is not None and bool(filter_ast.children) and entity_hits`) is zero-cost when there is no filter or no entities; scores are re-paired via an id→score map with no orphan/off-by-one; `provenance_pass_ran` marks the entity/relationship surfaces covered only when the pass ran (including the fail-closed path, where every survivor is verified). The `chunk_record_adapter` addition is engine-agnostic and VC-safe (default `None` = raw-chunk identity → zero VectorCypher change), and makes chronicle enforce `occurred_at` with the same `COALESCE(occurred_at, source_timestamp)` view its chunk channel uses — closing a genuine cross-surface inconsistency that false-dropped entities whose event time lives only in `source_timestamp`.

**Security / leak-closure** — The filter leak is fully closed: an entity survives iff ≥1 provenance chunk passes; stub entities with empty `source_chunk_ids` are dropped; fail-closed (ADR-001) drops unverified items on a fetch exception (never returns them) and records exactly one `Degradation`. No injection, no PII on metrics, `namespace_id` correctly kept off the counter labels, no plain-`str` secret-named fields.

**Tests** — All new tests have teeth and anti-vacuity gates: the flipped composition pin, the partial-match narrowing case, the fail-closed degradation case, the implementation-blind conformance-oracle chronicle leg, and `TestChunkRecordAdapterCoalescesOccurredAt` (survives *with* the adapter, drops *without* it, and still drops when both fields fail the horizon). The re-enabled e2e pin keeps its `pytest.fail` anti-vacuity gates and now surfaces real entities. No leftover instrumentation or stray skip/xfail markers.

**Observability / telemetry** — `docs/telemetry-contract.json` correctly enumerates the new `chronicle.entity_filter` component value for `khora.filter.provenance.degraded_total`; the drift gate passes; the fail-closed path logs at WARNING with `exc_info` and records one structured degradation.

**Performance** — Bounded provenance fetch (500-id pages, unique-id dedup, single `get_chunks_batch` per page, no N+1); zero cost on unfiltered recalls; the `_chunk_to_record` adapter is O(1) per chunk; no redundant fetches (provenance chunks were never fetched before this path).

**Docs / hygiene** — No internal identifiers leaked (only GitHub issue/PR refs); docstrings and the corrected seed comment are accurate; the CHANGELOG `### Fixed` entry describes the behavioral fix.

### Non-blocking note
- The e2e test disables the query router via a private-field mutation (`_router_enabled = False`) to keep the entity channel live for a single-token query. It is safe/idempotent and mirrors the hermetic composition test, but promoting a "run all channels" knob to a first-class setting would be cleaner long-term. Not a blocker.

CI: all required checks green on the reviewed commit.

filippetrovic:
Reverted the CHANGELOG.md edit — khora authors the changelog at release time (sibling provenance-filter PR #1481 dropped its entry for the same reason). No functional change; the chronicle entity-provenance fix, telemetry-contract update, and all tests are unchanged. The reverted entry's content (COALESCE-view adapter rationale) remains available in the PR body for release-time authoring.

PR #1496, opened 2026-07-10
Make the provenance filter's record view document-key aware

### LINKED ISSUE #1494: ∃-over-provenance filter over-drops entities on document-key leaves (field-incomplete record view)
## Summary

Follow-up to #1457/#1458 (found by end-to-end testing through deyta-cli on main `88f66238`). The ∃-over-provenance entity filter **over-drops all entities when the caller filter contains a document-key leaf** (`source`, `source_name`, `title`, …) and the recall routes to the graph path. Fail-closed direction — no data leaks — but legitimate entities disappear, and the filter report credits the entity surface as enforced for a leaf that was evaluated against field-incomplete records (the ADR-092 field-completeness rule, added for #1223, prohibits exactly this crediting).

## Reproduction (embedded sqlite_lance, real pipeline — observed via deyta-cli)

```python
await kb.remember("The pagination bug was fixed by resetting the offset cursor",
                  namespace=ns, source="linear")   # doc-1 → entities: pagination bug, offset cursor
await kb.remember("Team agreed to ship the pagination fix on Friday",
                  namespace=ns, source="slack")    # doc-2 → entities: pagination fix, friday

# Compound doc-key + date filter (date key forces the graph path):
r = await kb.recall("pagination fix", namespace=ns,
                    filter={"source": "linear", "occurred_at": {"$gte": "2020-01-01T00:00:00Z"}})
# chunks   == [doc-1 chunk]   ← chunk channel enforces correctly
# entities == []              ← WRONG: doc-1's entities should survive (their provenance
#                               chunk is the very chunk returned above)
# engine_info.filter.unenforced_keys == []  ← report over-claims enforcement

# Controls:
#   {"occurred_at": {"$gte": "2020-…"}} (graph path)  → entities kept   ✓ (date leaves work)
#   {"source": "linear"}               (simple path)  → entities narrow ✓ (#857 projection)
```

## Mechanism (anchors on main `88f662389`)

1. The `Chunk` dataclass carries **none** of the eight Q8-denormalized document keys — only temporal keys + `metadata` (`core/models/document.py:158`).
2. `get_chunks_batch` reads the full `khora_chunks` row (`SELECT *`) but `_temporal_row_to_chunk` builds the `Chunk` without the doc-key columns — they are read and discarded (`storage/backends/sqlite_lance/vector.py:282,360`; the class-level structural gap applies to every backend).
3. `filter_items_by_provenance` with `chunk_record_adapter=None` (the VectorCypher call) evaluates the compiled predicate directly against those chunks (`filter/provenance.py:158`), so a `source == "linear"` leaf resolves against an absent field → fails for **every** provenance chunk → every entity (and hence every relationship) is dropped.
4. Chronicle is affected too, more subtly: its adapter `_chunk_to_record(chunk, doc)` resolves the seven doc keys from a hydrated `DocumentProjection` (`engines/chronicle/engine.py:476`, keys at `:465`), but the ∃ pass calls the adapter with `doc=None`, leaving only the `chunk.source_document` tertiary fallback — `None` on batch-fetched chunks.

## Scope

Only filters combining a doc-key leaf with graph-path routing (a date key or temporal wording) are affected. Date-only filters — including everything the Peras `start`/`end` fold emits — behave correctly, which is why the #1457/#1458 test suites (all date-keyed) stayed green.

## Fix (agreed direction)

Make the ∃ record view field-complete for doc keys — the data is already in the row/store:

- `filter_items_by_provenance` gains doc-key awareness: when `filter_leaf_keys(filter_ast)` intersects the doc-key set, batch-fetch `DocumentProjection`s for the fetched provenance chunks' unique `document_id`s via the existing `storage.get_document_projections_batch` (`storage/coordinator.py:2231`) and evaluate against a chunk+projection merged record. Zero extra fetch when the filter has no doc-key leaf.
- Adapter call convention becomes `(chunk, doc)`; Chronicle's `_chunk_to_record` already has exactly that signature — its `doc=None` gap closes for free.
- Projection-fetch failure follows the existing fail-closed contract: unverified items dropped + one `Degradation(reason="document_fetch_failed")`.

No ADR change needed — this is an implementation bug against ADR-092's existing INV-B + field-completeness rules.

filippetrovic:
## Summary
- The ∃-over-provenance filter that re-checks entities and relationships against the caller filter evaluated the compiled predicate against bare provenance chunks. Those chunks structurally lack the seven denormalized document keys (`source`, `source_name`, `source_type`, `source_url`, `external_id`, `content_type`, `title`), so any filter combining a document-key leaf with graph-path routing over-dropped **every** entity and relationship — even ones whose provenance chunk passed the filter (the chunk channel returned that very chunk).
- Hydrate the parent-document projection inside `filter_items_by_provenance` **only** when the filter references a document key, and fold those keys onto the record the predicate evaluates. Filters with no document-key leaf skip the projection fetch entirely (zero extra query, byte-identical behavior).
- `occurred_at` stays the raw chunk value on the no-adapter path, so the ∃ pass agrees exactly with the graph chunk channel (no coalescing). Chronicle's adapter now receives the hydrated projection, closing a latent gap where its entity filter always saw document keys as absent.
- A projection-fetch failure follows the existing fail-closed contract: unverified items are dropped and a single `document_fetch_failed` degradation is recorded (added to the telemetry contract).
- The document-key set is now a single source of truth shared between the filter helper and the Chronicle engine.

## Test plan
- [x] Unit tests pass — provenance/conformance-oracle suites, full `recall` suite, telemetry contract drift gate
- [x] Repro verified: a compound `{source, occurred_at}` filter on the graph path keeps the matching document's entities and drops the rest; a filter with no document-key leaf performs zero projection fetches (spied)
- [ ] Integration tests pass (CI — require Postgres/Neo4j)
- [ ] E2E `vc_full` filter-report-invariant case (CI — requires the live matrix)

Fixes #1494

## Summary by CodeRabbit

* **New Features**
  * Provenance filters can now evaluate denormalized document-level fields (document-key hydration) when the filter references document keys.
  * Compound document-key and date filters behave consistently across supported recall paths.
* **Bug Fixes**
  * Fail-closed provenance filtering now covers both provenance-chunk fetch failures and document-projection hydration failures, excluding only items that were already verified.
  * Chronicle deterministic recall filtering now uses consistent document-key handling for doc-key predicates.
* **Monitoring**
  * Telemetry metric `khora.filter.provenance.degraded_total` now reports `document_fetch_failed` separately from provenance retrieval failures.


filippetrovic:

Eight independent reviewers (security, performance, correctness, tests, database, observability, docs, devil's advocate) + a dedicated security-audit pass. Overall assessment: **APPROVE** — the one HIGH finding was investigated and refuted; the actionable MEDIUM/LOW items were addressed in `96f5edbe`.

### HIGH
- **H1 · [Correctness] — REFUTED (not a bug).** `filter/provenance.py` — the concern that a multi-page recall could append *multiple* `document_fetch_failed` degradations. Investigated: the `break` lives in the `except` clause paired with the `try` that wraps the *inner* doc-page loop, so at that point the only enclosing loop is the **outer** chunk-page loop — `break` terminates it. Confirmed by code reading and an empirical replication of the exact control structure (outer iterations = 1, degradations = 1). Exactly **one** degradation is emitted regardless of page count, matching the docstring/ADR-001 contract. No change required. (A dedicated multi-page-doc-failure guard test is a reasonable low-risk follow-up.)

### MEDIUM
- **M1 · [Observability] — FIXED (`96f5edbe`).** The `khora.filter.provenance.degraded_total` counter *description string* in source still named only `provenance_fetch_failed`. Updated to name both `provenance_fetch_failed` and `document_fetch_failed` (the telemetry-contract JSON was already correct; the CI drift gate passes).
- **M2 · [Devil's Advocate] — FIXED (`96f5edbe`).** The degrade-**closed** behavior of the entity/relationship ∃ pass vs Chronicle's chunk channel degrading **open** on a hydration failure was undocumented. Added a docstring note: the ∃ pass drops unverified items because an existence claim over provenance must be *verified, not guessed*, and a transient projection-fetch failure can therefore return chunks while the entity surface narrows (surfaced via the `document_fetch_failed` degradation).
- **M3 · [Devil's Advocate] — post-merge discussion.** Whether doc keys should be denormalized onto the `Chunk` DTO at fetch time (storage layer) to remove the filter-layer hydration seam entirely. Legitimate architectural question; out of scope for this bug fix. Noted for a follow-up ADR/discussion.
- **M4 · [Performance] — acknowledged, acceptable.** Hydration runs per chunk-page rather than once up-front. Sound trade-off: fails closed early, dedups doc ids across pages via the shared `projections` cache, and avoids blocking on docs before chunk filtering. No change.

### LOW
- **L1 · [Devil's Advocate] — FIXED (`96f5edbe`).** Documented the no-adapter path's assumption that `Chunk` carries `occurred_at` natively; an engine whose chunk channel coalesces `occurred_at` MUST pass a `chunk_record_adapter`.
- **L2 · [Tests] — follow-up.** Coverage islands worth closing later: a relationship-surface (not just entity) doc-key case, a `$ne` doc-key filter *with* hydration, a multi-page doc-hydration-failure guard, and a present-but-`None` projection value. All low-risk given the ∃-pass logic is shared and already covered for the entity surface.

### PASS / clean
- **Security (reviewer + audit): PASS.** No injection/auth/crypto/SSRF/file/deserialization surface; no new dependencies. Negation operators fail safe (absent key → `$ne`/`$nin` match, over-keep only where semantically correct); the new `get_document_projections_batch` fetch is namespace-scoped (tenant isolation enforced at the SQL layer); fetch failures drop unverified items (fail-closed).
- **Database: PASS.** Protocol signature matches the coordinator; IN-list bounded by `_PAGE_SIZE`; deduped; no schema/migration/ORM/pool changes.
- **Docs: CLEAN.** Module docstring + `_PAGE_SIZE` comment accurately describe the hydration pass with no early-exit claim; no architecture-doc drift; CHANGELOG correctly untouched.
- **Tests: strong.** Repro asserts both survivor-present and non-match-absent; the needs_docs=False path spies the fetcher and asserts zero calls; the fail-closed test pins exactly one `document_fetch_failed` degradation; the conformance-oracle falsifiability case genuinely fails on a field-incomplete record.

**Verdict: APPROVE.** No CRITICAL findings; the HIGH was refuted; MEDIUM/LOW actionable items addressed in `96f5edbe`.

probablyrobot (approved):

Issue #1490, opened 2026-07-10
khora_chunks writers stamp chunk bookkeeping into user metadata and drop chunker_info (all temporal-path writers)


filippetrovic:
## Summary

`khora_chunks.metadata` should carry user/document metadata, but **every
temporal-path writer merges chunker bookkeeping into it**: `chunk_index`,
`start_char`, `end_char`, `token_count`. Consequences:

- Whole-blob `$eq` metadata filters can never match — the stored blob always
  contains internal keys the caller didn't write.
- `metadata.<path>` filters can match internal keys.
- The blobs are writer-inconsistent: the pipeline ingest writer lets user keys
  win collisions, while the engine writers stamp bookkeeping **on top of**
  user/document keys.

Simultaneously, two of the three writer families **drop
`TemporalChunk.chunker_info`** — the column migration 038 added precisely to
carry chunker output — so the chunker's self-identification and per-chunk
output (e.g. the conversation chunker's message map) never reach
`khora_chunks`, and `RecallChunk.chunker_info` returns `{}` despite its
documented minimum contract `{"chunker": "<name>"}`.

The main `chunks` table already does this correctly: typed bookkeeping columns
plus `chunker_info` (migration 037). `khora_chunks` should follow the same
ownership rule: **`metadata` is user-owned; chunker output lives in
`chunker_info`.**

## Environment

- **Khora**: `v0.22.2` (line references: `main` @ `ebbb7e4e`)
- **Storage backends**: all (the defect is at the writer level, storage-agnostic)
- **Engines**: `vectorcypher`, `skeleton`

## Mechanism

Six writer sites construct persisted `TemporalChunk`s:

| Site | Metadata pollution | `chunker_info` |
| --- | --- | --- |
| `pipelines/flows/ingest.py:1188` | 4 keys, user keys win collisions | dropped → `'{}'` |
| `engines/skeleton/engine.py:425`, `:1033` | 3 keys stamped **over** doc metadata (no `token_count`) | dropped; `raw_chunk.metadata` (chunker output) discarded |
| `engines/vectorcypher/engine.py:1272`, `:2030`, `:3443` | 3 keys stamped over doc metadata | kept correctly |

The shared reader `core/temporal.py:137-141` (`temporal_chunk_to_chunk`)
restores the four `Chunk` fields from the metadata blob and passes the
polluted blob through to filtering and recall.

(The `TemporalChunk` constructed at `engines/vectorcypher/engine.py:1953` is
[...]

PR #1498, opened 2026-07-10
Backfill khora_chunks bookkeeping keys from metadata into chunker_info

### LINKED ISSUE #1491: Backfill: move chunk bookkeeping from khora_chunks.metadata to chunker_info (Postgres + SQLite)
## Summary

Companion data migration to #1490. After #1490, newly written `khora_chunks`
rows are clean (`metadata` user-owned, bookkeeping + chunker output in
`chunker_info`). This migration repairs **existing** rows: it strips the four
bookkeeping keys (`chunk_index`, `start_char`, `end_char`, `token_count`) from
`metadata` and materializes `chunker_info`, recovering the chunker output
`khora_chunks` never stored where it is recoverable.

**Must ship in the same release as #1490**: the fixed reader has no metadata
fallback, and the backfill assumes writers no longer pollute (rows written by
pre-fix writers after the migration ran would stay dirty).

## Key insight — join-based recovery, the blob is never trusted

`khora_chunks.id == chunks.id`: both tables are written from the same `Chunk`
object at ingest (`pipelines/flows/ingest.py` writes `chunks` via
`storage.create_chunks_batch` and `khora_chunks` via the temporal store, same
IDs). The main `chunks` table holds the four values in **typed integer
columns** and a **populated `chunker_info`** (since migration 037). So for
every row with a main-table twin, recovery is authoritative — including the
chunker name and per-chunk output that `khora_chunks` never persisted.

## Design

**Tier 1 — main-table twin exists** (the majority):

```sql
UPDATE khora_chunks kc
SET chunker_info = kc.chunker_info || c.chunker_info || jsonb_build_object(
        'chunk_index', c.chunk_index,
        'start_char',  c.start_char,
        'end_char',    c.end_char,
        'token_count', c.token_count),
    metadata = kc.metadata - ARRAY(
        SELECT t.k FROM (VALUES
            ('chunk_index', to_jsonb(c.chunk_index)),
            ('start_char',  to_jsonb(c.start_char)),
            ('end_char',    to_jsonb(c.end_char)),
            ('token_count', to_jsonb(c.token_count))) AS t(k, v)
        WHERE kc.metadata -> t.k = t.v)
FROM chunks c
WHERE c.id = kc.id
  AND kc.namespace_id = :namespace_id
  AND kc.metadata ?| ARRAY['chunk_index','start_char','end_char','token_count']
  AND NOT (kc.chunker_info ? 'chunk_index')
```

- **Merge, not replace**: `kc.chunker_info || c.chunker_info || bookkeeping`
  preserves `chunker_info` already written by the engine-path writers.
- **Equality-gated strip** — exact, not heuristic: the old pipeline writer let
  user keys win collisions, so a blob value that *differs* from the typed
  column is user data — it stays in `metadata`, while `chunker_info` still
  receives the authoritative column value.

**Tier 2 — no twin** (e.g. skeleton-written rows): type-gated blob move —
move a key only if its value is `number`-typed (the writers only ever wrote
ints; a user's string-valued `chunk_index` survives untouched). No chunker
name is fabricated for these rows.

**Guard / idempotency**: `metadata ?| ARRAY[four keys] AND NOT
(chunker_info ? 'chunk_index')` — post-fix rows always carry `chunk_index` in
`chunker_info`, so a new row whose *user* metadata legitimately contains such
a key is never touched; re-running converges to zero rows.

## Dialects and ops

- **Postgres** — migration-043 ops pattern: `has_table` guards
  (`khora_chunks` is runtime-created, not Alembic-managed; fresh deploys
  no-op), `SET lock_timeout = '5s'`, `khora_chunks_content_tsv_update`
  trigger disabled in try/finally (only `metadata`/`chunker_info` change,
  never `content`), namespace-batched UPDATEs, structured
  `khora.migration.applied` completion log.
- **SQLite** (sqlite_lance stack — `metadata`/`chunker_info` live on the
  SQLite side; LanceDB holds embeddings only, and the main `chunks` table is
  in the same database file, so the tier-1 join applies): same statements via
  `json_patch` / `json_remove` / `json_type` + `UPDATE ... FROM`, following
  migration 038's cross-dialect pattern.
- **Downgrade**: no-op — once moved, the origin of a key (user vs. writer)
  cannot be disambiguated in reverse (migration-043 precedent).

## Out of scope — stated plainly

- **SurrealDB, weaviate, turbopuffer rows**: no migration infrastructure for
  these stores. Their pre-fix rows keep the bookkeeping keys in `metadata` as
  inert data — no code path reads them; only a whole-blob metadata `$eq`
  filter can observe them. No reader fallback exists (by design, see #1490).
- **Twins predating migration 037**: the four keys are recovered from the
  typed columns, but the chunker name is unrecoverable — it was never
  persisted anywhere before 037.
- Every pre-fix row is rewritten once; on large Postgres tables plan a
  `VACUUM` afterward.

## Acceptance

Post-migration on Postgres/SQLite:

- No row retains any of the four bookkeeping keys in `metadata`, except
  user-owned values preserved by the equality gate.
- Every tier-1 row's `chunker_info` ⊇ its main-table `chunker_info` plus the
  four keys.
- Re-running the migration statements updates zero rows.

filippetrovic:
## Summary
- Adds Alembic data migration `053_khora_chunks_bookkeeping_to_chunker_info` that relocates the four chunk-bookkeeping keys (`chunk_index`, `start_char`, `end_char`, `token_count`) out of `khora_chunks.metadata` and into `khora_chunks.chunker_info` on every existing row. Backfill companion to the writer/reader change in #1490: the temporal-chunk reader now sources those fields from `chunker_info` with **no metadata fallback**, so legacy rows must be migrated before release (they otherwise read back as zeros).
- Runs on **both** Postgres (JSONB `||` / `-` operators) and SQLite (JSON1 `json_patch` / `json_remove`), since `khora_chunks` exists on both the pgvector store and the embedded sqlite_lance stack — the SQLite branch is load-bearing, not a green-keeper.
- Two tiers per row, keyed on whether the row has a twin in the main `chunks` table:
  - **Tier 1 (twin exists):** copy the twin's typed column values into `chunker_info`, merged as `kc.chunker_info || c.chunker_info || bookkeeping` (bookkeeping stamped last so it wins), and strip a metadata key **only where its value equals the typed column** — a differing value is a user key that merely collides on name and stays.
  - **Tier 2 (no twin):** read from `metadata` itself, moving a key only when its value is number-typed (`jsonb_typeof='number'` / `json_type='integer'`); string-typed collisions stay.
- Idempotent / convergent: guarded by `metadata ?| [four keys] AND NOT (chunker_info ? 'chunk_index')`, so a re-run touches zero rows. Postgres path bounds `lock_timeout` to 5s and suppresses the `content_tsv` trigger during the relocation (restored in a `finally`); no-op downgrade (a relocated key's origin is not reversibly disambiguable). Both `has_table` guards early-return `(0,0)` on fresh deploys.
- Two SQLite JSON1 traps handled explicitly: `json_remove(x, NULL)` nulling the `NOT NULL` column (guarded with an absent-path sentinel) and `json_patch` deleting keys whose patch value is JSON `null` per RFC 7386 (guarded via nested `json_set` with a NULL *path* no-op).
- Bumps the head-revision assertions in the sibling migration tests to the new head and the bundled-migration count.

## Test plan
- [x] Unit tests pass — SQLite lane: six archetypes (incl. a three-way base→twin→bookkeeping merge-precedence case) + idempotent re-run; full Alembic chain reaches the new head on a fresh SQLite DB; migration-bundling + sqlite-migration guards
- [ ] Integration tests pass — Postgres lane runs in CI: the same six archetypes against real JSONB plus `content_tsv` trigger restoration (requires Docker Postgres, unavailable in the authoring sandbox)
- [ ] Manual verification: `uv run alembic upgrade head` on a populated store; confirm `khora_chunks.metadata` no longer carries the four keys (except equality-gate-preserved user values) and `chunker_info` carries them

Fixes #1491

## Summary by CodeRabbit

- **Data Migration**
  - Backfilled chunk bookkeeping values into the dedicated chunk information store from existing per-row metadata.
  - Preserves user-supplied metadata when keys collide, moving only the numeric bookkeeping fields.
  - Runs safely on PostgreSQL and SQLite with idempotent behavior; PostgreSQL includes operational safeguards and restores related refresh behavior after the move.

- **Tests**
  - Added comprehensive coverage for relocation rules, collision handling, and idempotent reruns (including SQLite and PostgreSQL lanes).
  - Updated existing integration/unit expectations for the new migration “head” version and adjusted migration bundling file count.

### DIFF
--- a/src/khora/db/migrations/versions/053_khora_chunks_bookkeeping_to_chunker_info.py
+++ b/src/khora/db/migrations/versions/053_khora_chunks_bookkeeping_to_chunker_info.py
@@ -1,0 +1,398 @@
+"""Move chunk bookkeeping keys from khora_chunks.metadata into chunker_info.
+
+Revision ID: 053_khora_chunks_bookkeeping_to_chunker_info
+Revises: 052_entities_source_chunk_ids_gin
+Create Date: 2026-07-10
+
+Data change only — NO schema change. The four chunk-bookkeeping keys
+``chunk_index`` / ``start_char`` / ``end_char`` / ``token_count`` used to
+live in ``khora_chunks.metadata`` (the temporal store's user/document
+metadata blob). The writer/reader refactor (khora#1491) moved them into
+``khora_chunks.chunker_info`` at every chunk-writer site and the
+temporal-chunk reader now sources the four fields from ``chunker_info``
+**exclusively, with no metadata fallback** — a legacy row that still
+carries them only in ``metadata`` reads back as zeros. This migration is
+the backfill companion: it relocates the four keys on every existing row
+so those rows keep reading correctly after the reader change lands. It
+must run before release for the same reason.
+
+Cross-dialect: unlike 043/044 this migration runs on **both** dialects.
+``khora_chunks`` exists on Postgres (``PgVectorTemporalStore``) and on the
+sqlite_lance embedded stack, so the SQLite branch performs the same
+relocation with the JSON1 functions (``json_patch`` / ``json_remove`` /
+``json_extract``) instead of the Postgres JSONB operators. The
+sqlite_lance fixture stack runs the full Alembic chain, so a SQLite no-op
+would leave embedded rows unmigrated — the SQLite branch is load-bearing,
+not a green-keeper.
+
+Two tiers per row, keyed on whether the row has a twin in the main
+``chunks`` table (``chunks.id = khora_chunks.id``, the same UUID):
+
+* **Tier 1 (twin exists).** The main ``chunks`` row carries the four
+  values in *typed* columns (``chunk_index`` etc., ``NOT NULL`` integers)
+  plus its own ``chunker_info``. The relocation copies the typed column
+  values into ``khora_chunks.chunker_info`` (merged after the twin's
+  ``chunker_info``) and strips a metadata key **only where its metadata
+  value equals the typed column value**. A metadata key whose value
+  differs from the column is a user key that merely collides on name — it
+  is left in ``metadata`` untouched, while ``chunker_info`` still gets the
+  authoritative column value.
+
+* **Tier 2 (no twin).** No typed columns to trust, so the values are read
+  from ``metadata`` itself. A key is moved only when its metadata value is
+  number-typed (``jsonb_typeof = 'number'`` on Postgres, ``json_type =
+  'integer'`` on SQLite); a string-typed collision is a user key and stays
+  in ``metadata``. Exactly the moved keys are stripped. No synthetic
+  ``chunker`` name is fabricated — only the four numeric keys move.
+
+Merge precedence (both tiers, both dialects): the row's own existing
+``chunker_info`` is the base and always wins over nothing, the twin's
+``chunker_info`` (Tier 1) layers on top, and the four bookkeeping keys are
+stamped **last** so they win any collision — mirroring the writer, which
+stamps the bookkeeping keys last. On Postgres this is
+``kc.chunker_info || c.chunker_info || bookkeeping`` (right operand wins in
+JSONB concat). On SQLite ``json_patch(a, b)`` lets ``b`` win, so the calls
+nest ``json_patch(json_patch(kc.chunker_info, c.chunker_info),
+bookkeeping)`` to reproduce the same base → twin → bookkeeping order.
+
+Idempotency / convergence (both tiers, both dialects): the UPDATE is
+guarded by ``metadata ?| ARRAY[four keys] AND NOT (chunker_info ?
+'chunk_index')`` — a row is visited only while it still carries any of the
+four keys in ``metadata`` *and* has not yet had ``chunk_index`` written to
+``chunker_info``. Once relocated, ``chunker_info`` carries ``chunk_index``
+so the row is skipped, and re-running the migration touches zero rows. A
+row where a differing-value user key legitimately remains in ``metadata``
+is still skipped on re-run because ``chunker_info`` now has ``chunk_index``.
+
+Scale (~1M chunk rows): the relocation runs namespace-batched — one Tier-1
+UPDATE and one Tier-2 UPDATE per distinct ``khora_chunks.namespace_id`` —
+so no single statement spans the whole table. The whole Alembic chain runs
+inside ONE transaction (``env.py:do_run_migrations``), so batching bounds
+per-statement cost and planning, not lock-hold duration.
+
+Trigger suppression (Postgres): ``khora_chunks`` carries a
+``BEFORE INSERT OR UPDATE`` trigger (``khora_chunks_content_tsv_update``)
+that recomputes ``to_tsvector('english', content)`` on every updated row.
+This migration touches only ``metadata`` / ``chunker_info`` and never
+``content``, so re-deriving the tsvector is wasted CPU on a ~1M-row table.
+The trigger is disabled for the relocation and re-enabled in a ``finally``
+so it is restored on every exit path. ``DISABLE TRIGGER`` / ``ENABLE
+TRIGGER`` is transactional, so a rolled-back migration leaves the trigger
+enabled. SQLite has no such trigger and skips the toggle.
+
+Lock-timeout safety (Postgres): a ``SET lock_timeout = '5s'`` bounds the
+brief ``ACCESS EXCLUSIVE`` lock ``DISABLE TRIGGER`` takes and the row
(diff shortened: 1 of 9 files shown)


filippetrovic:
## 🔍 Multi-perspective review (automated review team)

Seven independent reviewers (security, performance, correctness, tests, database, docs, devil's advocate) plus a security audit pass reviewed this migration. **Overall assessment: APPROVE** — no blocking findings. Two follow-up recommendations below; both are non-blocking for this PR.

### CRITICAL
_None._

### HIGH
_None._ (Two findings were initially flagged as blockers; both were investigated and downgraded — see Notes.)

### MEDIUM
- **M1 — [Database / Devil's Advocate] `chunks.{chunk_index,start_char,end_char,token_count}` are nominally nullable** (`db/models.py:229-232` declare `default=0` with no `nullable=False`; `000_initial_schema` created them with `server_default="0"` only). Tier-1 stamps `jsonb_build_object('chunk_index', c.chunk_index, …)` into `chunker_info`, so a genuinely-NULL column would write a JSON `null`. **Why this is not a blocker for this PR:** (a) the reader coerces `int(ci.get("chunk_index", 0) or 0)` (`core/temporal.py:138-141`), so a `null` degrades to `0` — no crash; (b) `server_default=0` plus integer-writing writers mean these columns are never NULL in practice; (c) there is **no regression** vs. current post-#1490 behavior — a legacy row already reads `0`, and the equality-strip does not fire on a `null` mismatch so the real value stays preserved in `metadata`; (d) adding `NOT NULL` is a schema change outside this backfill's scope. **Recommendation:** open a separate ticket to add `NOT NULL` constraints (+ ORM `nullable=False`) to these four `chunks` columns.
- **M2 — [Performance] Lock-hold duration.** The relocation runs inside the single Alembic chain transaction (`env.py` wraps the chain), so row locks on `khora_chunks` are held until the chain commits; on a ~1M-row table the cumulative hold can reach tens of seconds. This matches the established `043` backfill precedent (namespace-batching bounds per-statement cost, not lock duration), and `SET lock_timeout='5s'` bounds each statement. **No code change required.** **Recommendation:** run during a low-write window and note the expected duration in release notes.

### LOW
- **L1 — [Performance]** The `metadata ?|` / `NOT (chunker_info ? 'chunk_index')` JSONB guards and the Tier-2 `NOT EXISTS` are not index-backed, but the working set is bounded by the per-namespace batching and the PK-indexed twin join. Informational.

### Security Audit
No security-relevant surface: the migration is static SQL with a single bound `:namespace_id` parameter (no string interpolation of data → no injection), touches no dependencies, crypto, secrets, endpoints, or auth, and performs no cross-namespace access. No findings; nothing to auto-fix.

### Notes on refuted findings
- **Migration-file count** was flagged as "should be 55" — this is a miscount. The directory holds **54** migration files and `test_versions_dir_has_54_files` passes; the count bump (53→54) is correct.
- A suggested "redundant `c.id = kc.id AND NOT EXISTS`" in Tier-1 is based on a misread — Tier-1 uses only the twin join (`c.id = kc.id`); the `NOT EXISTS` is Tier-2's disjoint guard.

### What's strong
- Faithful to the pinned two-tier spec on **both** dialects; merge precedence (`kc.chunker_info || c.chunker_info || bookkeeping`) correct in Postgres `||` and the SQLite `json_patch(json_patch(kc, c), bookkeeping)` nesting.
- Idempotent/convergent guard (`metadata ?| [keys] AND NOT (chunker_info ? 'chunk_index')`); re-run touches zero rows.
- Two real SQLite JSON1 data-loss traps handled explicitly: `json_remove(x, NULL)` nulling the `NOT NULL` column (absent-path sentinel) and RFC-7386 `json_patch` null-key deletion (`json_set` NULL-path no-op).
- `has_table` guards for both tables, trigger disable/enable in `finally`, `55P03` lock-timeout classification, no-op downgrade — parity with the `043`/`038` precedent.
- Test coverage: six archetypes (incl. a three-way merge-precedence regression guard) + idempotent re-run, on both the SQLite unit lane and the Postgres integration lane.

_This is an automated consolidated review; the two MEDIUM items are follow-up recommendations, not merge blockers._

filippetrovic:
CodeRabbit nitpick applied in 2a8a379c: the precedence archetype now asserts the metadata strip side in both lanes — `md == {"start_char": 10, "author": "bob"}` (chunk_index equals the twin column → stripped; start_char 10 ≠ 333 → preserved as a user key). SQLite lane verified locally (2 passed).

msrdic (approved):

PR #1513, opened 2026-07-13
Read recall temporal fields from first-class Chunk columns

filippetrovic:
## Summary

Flip the recall-path temporal **readers** from the chunk `metadata` blob to the first-class `Chunk.occurred_at` / `Chunk.source_timestamp` columns. Stage 1 already made every recall-path chunk constructor populate those columns; this change makes the readers consume them.

Reader sites migrated (all in `src/khora/engines/vectorcypher/retriever.py`):
- `_extract_occurred_at` — now returns `datetime | None` off the first-class column (no `source_timestamp` fallback in the boost path); its consumer `_calculate_recency_scores` drops the now-redundant `fromisoformat` parse and keeps naive→UTC normalization.
- Temporal-sort key `_ts` — sorts on `chunk.occurred_at` with the same naive→UTC coercion and None/ordering semantics.
- Rerank date-prefix builders (cross-encoder + LLM) — read `item.occurred_at or item.source_timestamp` off the object, preserving the existing fallback order and rendered date format. The date prefix is built independently of the metadata blob; the adjacent `session_id` / `conversation_id` reads stay gated on the blob (user-space keys).
- Top-1-age telemetry — reads the first-class `occurred_at`; metric name and semantics unchanged.

### Behavior changes (all corrections, confined to temporally-ranked recalls)
- PPR-sourced and BM25-sourced chunks gain temporal visibility across **recency scoring, temporal sort, and the rerank date prefix** — they were previously blind because their metadata blob carried no usable `occurred_at`.
- A stale user-stored blob `occurred_at` can no longer silently override the authoritative column value during ranking.
- Vector / graph / recency / skeleton-sourced chunks are value-identical (their injections copied the column), so no ordering change there.

### Scope notes
- The metadata blob injections are intentionally left in place as dead writes; a follow-up removes them. Ungating the rerank date prefix from the blob (this PR) is what keeps that follow-up from silently dropping the prefix for empty-metadata chunks.

## Test plan
- [x] Unit tests pass (engines unit suite green; `retriever.py` reader lines executed)
- [x] Rewrote the tz-contract sort test to set the first-class column; added a blob-override-death test (stale blob + fresh column → column wins) and a naive-normalization test
- [x] Rerank date-prefix tests (cross-encoder + LLM) diverge a stale metadata-blob `occurred_at` from a fresh column so they fail if the reader regresses to the blob; added an empty-blob test per site (`metadata={}` + populated column → `Date:` present, no `Session:`) guarding the blob-gate removal
[...]

PR #1510, opened 2026-07-13
Populate first-class temporal fields in recall producers

filippetrovic:
## Summary

Recall-path `Chunk` constructors now populate first-class `occurred_at` / `source_timestamp` from their authoritative source (the chunk column or graph-node property) instead of leaving them null or mirroring one field onto the other. Two sites that overwrote user metadata stop doing so.

- **Recency channel** — sets `occurred_at` (previously omitted).
- **PPR channel** — sets `occurred_at` and `source_timestamp` (previously only `created_at`).
- **Typed-entity fast path** — preserves the stored user-metadata blob (previously replaced wholesale with a single injected key) and reads the real `source_timestamp` node property instead of mirroring `occurred_at`. No key is injected on this path; it returns directly from `retrieve()` and is gated on "no constraining filter", so preserving the real blob does not change filter results here.
- **Graph channel (SurrealDB/embedded fallback)** — carries `metadata` plus both temporal fields into the record builder (metadata was previously dropped in memory, and `occurred_at` was mirrored from `source_timestamp`).
- **Graph channel (Neo4j)** — `get_chunks_by_entities` now returns `source_timestamp`, and the reader uses it instead of mirroring `occurred_at`.

### Scope / behavior notes

- **Temporal-ranking readers are deliberately untouched in this stage.** `_extract_occurred_at`, the `_ts` sort key, the cross-encoder / LLM rerank prefix builders, and the top-1-age telemetry all still consume the `metadata` blob, and every existing blob injection is preserved — so ranking, sort, and rerank **ordering** is unchanged here. Those readers flip to the first-class fields in a follow-up stage.
- **Intended output refinement.** Chunks that have `occurred_at` set but `source_timestamp` NULL — or a real `source_timestamp` with NULL `occurred_at` — now surface the real value in the recall output projection instead of `null`. This is the data-correctness fix this change exists for.
- **Metadata completeness.** The fast path and the SurrealDB/embedded graph fallback stop discarding the user-metadata blob in memory. This is additive (nothing that was present is removed).

This is a staged change (1 of 3): stage 1 makes producers correct, stage 2 flips the readers to the first-class fields (with a ranking benchmark), stage 3 removes the now-redundant blob injections. Keeping the injections in place this stage is what makes stage 1 ordering-invisible.

Fixes #1507 (stage 1). Part of #1492 — the umbrella issue stays open for the final stage.

## Test plan

- [x] Unit tests pass (`make test-unit`: 9959 passed, 0 failed)
[...]

PR #1516, opened 2026-07-13
Remove recall-time metadata injections; return stored blob verbatim on every recall channel

### LINKED ISSUE #1492: Recall channels inject engine-internal keys into in-memory chunk metadata; filter and ranking semantics diverge per channel
## Summary

Five recall-time code paths construct in-memory `Chunk` objects whose
`metadata` dict is **not** the stored metadata: they inject engine-internal
keys (`occurred_at`, `connected_entities`, `ppr_score`) on top of it — and one
path replaces the user's metadata entirely. The injected blobs are never
persisted; they exist only during recall. But recall's own machinery consumes
`chunk.metadata` in two load-bearing places — the in-memory filter post-pass
and the temporal ranking readers — so the divergence between stored metadata
and in-memory metadata produces **channel-dependent filter semantics** and a
**silently temporally-blind PPR channel**.

This is the read-time half of the metadata-ownership problem. #1490 / #1491
fix the *persisted* pollution (bookkeeping keys written into
`khora_chunks.metadata`); even after they land, these injections keep the
in-memory metadata of five channels diverging from storage.

This issue documents the problem surface only. **Fix direction is
deliberately left open** — it will be refined separately.

## Environment

- **Khora**: `v0.22.2` (line references: `main` @ `ebbb7e4e`)
- **Engines**: `vectorcypher` (four sites, incl. the PPR channel), `skeleton`
  (one site). **`chronicle` is unaffected** — it models event time as a
  top-level record key and keeps `metadata` pure
  (`engines/chronicle/engine.py:500-512`), demonstrating that the injection
  pattern is not structurally required.
- **Storage backends**: all — the defect is engine-level and in-memory;
  nothing is persisted.

### Injection sites (in-memory `Chunk.metadata`, never persisted)

| # | Site | What it does |
| --- | --- | --- |
| 1 | `engines/vectorcypher/retriever.py:1355` (typed-entity fast path) | `metadata={"occurred_at": c_data.get("occurred_at")}` — **replaces** user metadata wholesale; the graph record carries no metadata blob on this path, so user keys are absent entirely |
| 2 | `engines/vectorcypher/retriever.py:4040-4042` (graph chunk channel) | injects `"occurred_at"` + `"connected_entities"` beneath the stored blob (stored keys win) |
| 3 | `engines/vectorcypher/retriever.py:4281-4286` (recency channel) | injects `"occurred_at"` beneath the stored blob; the adjacent comment states it exists for the downstream recency boost |
| 4 | `engines/vectorcypher/ppr_retrieval.py:482-488` (PPR channel) | injects `"occurred_at"` — read from `chunk.metadata.get("occurred_at")`, **not** from the chunk's first-class `occurred_at` attribute — plus `"ppr_score"`; the constructed `Chunk` also omits the first-class `occurred_at`/`source_timestamp` fields entirely |
| 5 | `engines/skeleton/engine.py:567-575` (skeleton recall) | re-wraps every store result into a fresh `Chunk` with `"occurred_at"` injected beneath the stored blob |

### The readers that make the injected keys load-bearing

- `_extract_occurred_at` (`engines/vectorcypher/retriever.py:504-510`) reads
  `(chunk.metadata or {}).get("occurred_at")` — feeds temporal sort and
  recency scoring.
- Cross-encoder rerank prefix builder (`retriever.py:2821-2826`) reads
  `metadata["occurred_at"]` / `["source_timestamp"]` /
  `["session_id"]` / `["conversation_id"]`.
- LLM rerank prefix builder (`retriever.py:3015-3023`) — same pattern.
- `_extract_source_system` (`retriever.py:513+`) reads
  `metadata["source_system"]` — same key-in-user-space pattern family.

### The seam where divergence becomes caller-visible

The graph chunk channel applies a full-AST **in-memory** post-filter
(`retriever.py:2124+`, via `filter/compilers/python.py`) against the
reconstructed chunks — i.e. against the **injected** blobs — while the
vector/BM25/recency channels evaluate the same filter as SQL pushdown against
the **stored** blobs. Whole-blob `$eq` semantics (`filter/model.py:841`,
Mongo-style embedded-document equality) make the difference observable.

## Observable bug surface

**(a) Channel-dependent filter semantics.** The same filter on the same chunk
gives different results depending on which channel retrieved it:

- `{"metadata.occurred_at": {"$exists": false}}` — stored blob has no
  `occurred_at` → SQL pushdown matches; the graph channel's post-filter sees
  the injected `occurred_at` → rejects. The predicate flips per channel.
- `{"metadata": {"team": "alpha"}}` (whole-blob equality) — matches on the
  vector channel once stored blobs are clean (#1490), but fails on the graph
  channel (injected `occurred_at` + `connected_entities` make the blob
  unequal) and always fails on the typed-entity fast path (blob replaced with
  a single-key dict).
- Any `metadata.<user key>` predicate is unsatisfiable on the typed-entity
  fast path — user metadata is absent there by construction.

**(b) PPR channel is silently temporally blind.** Site 4 sources chunks from
the chunk store, where event time lives in the first-class `occurred_at`
attribute — but it injects `chunk.metadata.get("occurred_at")` (normally
absent from the blob) and omits the first-class fields on the `Chunk` it
builds. Result: PPR-sourced chunks carry `occurred_at=None` on both surfaces,
so `_extract_occurred_at` returns `None` and recency boosting / temporal
sorting silently degrade for everything the PPR channel contributes. No
degradation is recorded.

**(c) User metadata destroyed in-memory on the fast path.** Site 1 discards
the caller's metadata for every chunk it returns — the rerank prefix builders
(which look for `session_id`/`conversation_id`) and any other in-memory
metadata consumer see none of it.

**(d) Dead-weight keys.** `connected_entities` is written and never read
(the public `RecallChunk.connected_entity_ids` is derived after the engine
returns, by inverting entity→chunk links). No reader of the injected
`ppr_score` blob copy was found either (the score travels separately in the
result tuple).

**(e) Inconsistent read-side override.** On sites 2/3/4/5 the stored blob
spreads last, so a user-stored `metadata.occurred_at` value overrides the
injected one for the blob readers — user metadata influences temporal ranking
at read time on those channels, but not on channels without injections, and
not via the documented ingest-time contract (`metadata.custom["occurred_at"]`
feeding the column).

## Reproduction

1. Ingest a document with `metadata={"team": "alpha"}` into a namespace with
   entities/graph content (so recall routes through both the vector and graph
   channels).
2. `recall(..., filter={"metadata.occurred_at": {"$exists": false}})`.
3. Chunks arriving via the vector channel match (stored blob has no
   `occurred_at`); the same chunks arriving via the graph channel are
   rejected by the in-memory post-filter (injected `occurred_at` present).
   Which copy wins depends on fusion — the filter outcome is a function of
   retrieval channel, not of the chunk's data.

For (b): any recall routed through the PPR channel with recency boosting
enabled — compare the effective `occurred_at` seen by
`_extract_occurred_at` (always `None`) against the chunk's stored
`occurred_at` column.

## Expected behavior

- A given (chunk, filter) pair evaluates identically regardless of which
  retrieval channel produced the chunk.
- `chunk.metadata` visible to recall internals contains exactly the keys the
  caller stored — engine-derived values do not share a namespace with user
  data.
- A chunk whose store carries an event time is never invisible to temporal
  ranking merely because of the channel that retrieved it.

## Actual behavior

As per Mechanism / Observable bug surface: five channels operate on synthetic
metadata blobs; filter predicates flip per channel; the PPR channel is
temporally blind; the typed-entity fast path erases user metadata in-memory.

### LINKED ISSUE #1509: Remove all recall-time metadata injections; chunk.metadata = stored blob on every channel (stage 3/3, fixes #1492)
## Context

Stage 3 of 3 — **fixes #1492**. **Depends on #1508 (stage 2)**: the injected blob keys now have zero readers. Delete every recall-time metadata injection so `chunk.metadata` is the stored blob, verbatim, on every channel.

This is the stage where #1492's caller-visible invariants are restored:

- The graph channel's full-AST in-memory post-filter (`retriever.py:2229` region) evaluates the same stored blob the vector/BM25 SQL pushdown sees → **a given (chunk, filter) pair evaluates identically regardless of retrieval channel**.
- Response `metadata` contains exactly the caller's stored keys on every channel.

Zero ranking change — after stage 2, nothing reads these keys.

Line references: `main` @ `828e1eb`.

## Changes

Replace the constructor `metadata` expression with stored-blob pass-through at all six injecting sites:

1. `src/khora/engines/vectorcypher/retriever.py:3286-3302` (`_simple_retrieve`): `metadata=r.chunk.metadata or {}` — drop injected `occurred_at`.
2. `retriever.py:4237-4250` (`_vector_search_chunks`): same.
3. `retriever.py:4145-4164` (`_fetch_chunks_from_entities`, covers both the Neo4j and SurrealDB-fallback branches): `metadata=record.get("metadata") or {}` — drop injected `occurred_at` AND `connected_entities`.
4. `retriever.py:4386-4404` (`_recency_channel_chunks`): stored blob only — drop injected `occurred_at`.
5. `src/khora/engines/vectorcypher/ppr_retrieval.py:477-490`: `metadata=chunk.metadata if isinstance(chunk.metadata, dict) else {}` — drop injected `occurred_at` AND `ppr_score`.
6. `src/khora/engines/skeleton/engine.py:569-581` (skeleton recall re-wrap): `metadata=result.chunk.metadata or {}` — drop injected `occurred_at`.

(The typed-entity fast path became injection-free in stage 1 / #1507.)

Nothing else moves:

- `RecallChunk.connected_entity_ids` is derived post-engine by inverting entity→chunk links — unchanged.
- The PPR score travels in the result tuples — unchanged.
- A caller-stored blob `occurred_at` is user data and is returned verbatim — never scrubbed.

## Tests

- Update the two assertions that pin injected keys: `tests/unit/engines/test_retriever_coverage_push.py:1421-1422` (graph channel: expect metadata `{"author": "alice"}` exactly) and `:1495-1496` (vector channel: expect `{"source": "x"}` exactly).
- Add filter-invariant cases (filter-fuzzer + channel-matrix infra):
  - `{"metadata.occurred_at": {"$exists": false}}` on a chunk with a clean stored blob and a populated `occurred_at` column → matches via vector, BM25, AND graph channels.
  - Whole-blob `$eq` (`{"metadata": {<exact stored blob>}}`) → matches via the graph channel.
- Add metadata-shape assertions: for every channel (vector simple, vector full, BM25, graph Neo4j, graph SurrealDB fallback, recency, PPR, skeleton, typed-entity fast path), the returned `chunk.metadata` equals the stored blob; no `occurred_at`/`connected_entities`/`ppr_score` key appears unless the caller stored it.

## Verification

- `make format && make test`; filter-fuzzer suite green.
- Downstream audit (2026-07-13): peras / poros / console-monorepo / deyta-cli / typescript-sdk / deyta-core / omphalos / atlas-monorepo / connections-registry contain zero reads of `metadata["occurred_at"]` / `ppr_score` / `connected_entities` from recall responses — no consumer coordination required.

Include `Fixes #1492` in the PR body — this stage closes the umbrella issue.

filippetrovic:
## Summary
- Drop the engine-internal keys (`occurred_at`, `connected_entities`, `ppr_score`) that were injected into the in-memory `Chunk.metadata` at all six recall-path `Chunk` constructors (vector simple, vector full, graph entity-fetch, recency, PPR, skeleton). `Chunk.metadata` now carries the caller's stored blob, verbatim, on every retrieval channel.
- Temporal ranking and the two **production** rerank date-prefix builders (cross-encoder, LLM) already read the first-class `occurred_at`/`source_timestamp` columns (preserved on every constructor), so this is a **zero-ranking-change dead-write removal**. (The opt-in, non-production `CrossEncoderReranker(include_date_prefix=True)` experiment still reads a top-level blob key — out of scope here.)
- Restores cross-channel **filter invariance**: a given `(chunk, filter)` pair now evaluates identically whether the chunk came from the vector/BM25 SQL pushdown or the graph channel's in-memory post-filter, because the post-filter now sees the same stored blob. This is the internal `Chunk.metadata` the filter pass and rerank candidates consume — the caller-facing `RecallChunk` carries no `metadata` field, so there is no response-shape change.
- The dropped keys had zero readers (audited repo-wide): `connected_entity_ids` is a first-class `RecallChunk` field derived post-engine by inverting entity→chunk links, and the PPR score travels in the result tuple. A caller-stored `occurred_at`/`connected_entities`/`ppr_score` blob key is user data and still passes through verbatim.

## Test plan
- [x] `make format` clean, `make lint` green
- [x] Affected + new unit tests pass; full unit suite 9976 passed, 84% coverage (≥65%)
- [x] Embedded integration lane (sqlite_lance matrix): 154 passed, incl. a new cross-channel filter-invariance matrix test and a graph-channel post-filter agreement test
- [x] Postgres/Neo4j integration + e2e legs — green in CI (conformance, skeleton, vc_full)
- Added metadata-shape assertions across every channel, filter-invariant cases (`metadata.occurred_at $exists:false`, whole-blob `$eq`) proving identical evaluation via the vector, BM25, and graph channels, and a positive caller-stored-key pass-through case; tightened the two pinned-key assertions to exact stored-blob equality plus the first-class column.

Fixes #1492
Closes #1509

## Summary by CodeRabbit

* **Bug Fixes**
  * Preserved stored metadata exactly as provided during recall across vector, keyword, hybrid, graph, recency, and PPR retrieval.
  * Prevented system-generated fields such as event time, connection data, and ranking scores from being added to metadata.
  * Kept event timestamps available through their dedicated fields.
  * Ensured user-provided metadata keys are returned unchanged.
* **Tests**
  * Added coverage confirming consistent metadata filtering and whole-blob matching across retrieval modes.

### DIFF
--- a/src/khora/engines/skeleton/engine.py
+++ b/src/khora/engines/skeleton/engine.py
@@ -571,10 +571,7 @@ class SkeletonConstructionEngine:
                 namespace_id=result.chunk.namespace_id,
                 document_id=result.chunk.document_id,
                 content=result.chunk.content,
-                metadata={
-                    "occurred_at": result.chunk.occurred_at.isoformat() if result.chunk.occurred_at else None,
-                    **(result.chunk.metadata or {}),
-                },
+                metadata=result.chunk.metadata or {},
                 created_at=result.chunk.created_at or result.chunk.occurred_at or datetime.now(UTC),
                 occurred_at=result.chunk.occurred_at,
                 source_timestamp=result.chunk.source_timestamp,
--- a/src/khora/engines/vectorcypher/ppr_retrieval.py
+++ b/src/khora/engines/vectorcypher/ppr_retrieval.py
@@ -479,13 +479,7 @@ async def ppr_retrieve_chunks(
                         namespace_id=chunk.namespace_id,
                         document_id=chunk.document_id,
                         content=chunk.content,
-                        metadata={
-                            "occurred_at": (
-                                chunk.metadata.get("occurred_at") if isinstance(chunk.metadata, dict) else None
-                            ),
-                            "ppr_score": score_map[cid],
-                            **(chunk.metadata if isinstance(chunk.metadata, dict) else {}),
-                        },
+                        metadata=chunk.metadata if isinstance(chunk.metadata, dict) else {},
                         created_at=getattr(chunk, "created_at", None),
                         occurred_at=chunk.occurred_at,
                         source_timestamp=chunk.source_timestamp,
--- a/src/khora/engines/vectorcypher/retriever.py
+++ b/src/khora/engines/vectorcypher/retriever.py
@@ -3340,10 +3340,7 @@ class VectorCypherRetriever:
                     namespace_id=r.chunk.namespace_id,
                     document_id=r.chunk.document_id,
                     content=r.chunk.content,
-                    metadata={
-                        "occurred_at": r.chunk.occurred_at.isoformat() if r.chunk.occurred_at else None,
-                        **(r.chunk.metadata or {}),
-                    },
+                    metadata=r.chunk.metadata or {},
                     chunker_info=r.chunk.chunker_info or {},
                     created_at=r.chunk.created_at or r.chunk.occurred_at,
                     # Carry the chunk event-time and the producer's verbatim
@@ -4197,11 +4194,7 @@ class VectorCypherRetriever:
                     namespace_id=namespace_id,
                     document_id=UUID(record["document_id"]),
                     content=record["content"],
-                    metadata={
-                        "occurred_at": record.get("occurred_at"),
-                        "connected_entities": record.get("entity_ids", []),
-                        **(record.get("metadata") or {}),
-                    },
+                    metadata=record.get("metadata") or {},
                     chunker_info=_decode_chunker_info(record.get("chunker_info")),
                     # The graph store carries the chunk event-time (occurred_at)
                     # and the producer time (source_timestamp) as separate
@@ -4288,10 +4281,7 @@ class VectorCypherRetriever:
                         namespace_id=r.chunk.namespace_id,
                         document_id=r.chunk.document_id,
                         content=r.chunk.content,
-                        metadata={
-                            "occurred_at": r.chunk.occurred_at.isoformat() if r.chunk.occurred_at else None,
-                            **(r.chunk.metadata or {}),
-                        },
+                        metadata=r.chunk.metadata or {},
                         chunker_info=r.chunk.chunker_info or {},
                         created_at=r.chunk.created_at or r.chunk.occurred_at,
                         occurred_at=r.chunk.occurred_at,
@@ -4424,8 +4414,8 @@ class VectorCypherRetriever:
             for idx, sim in sim_pairs:
                 chunk, _emb = chunks_with_embedding[idx]
                 # Re-shape into ``(chunk_id, score, Chunk)`` matching
-                # ``_vector_search_chunks``. The Chunk's metadata dict
-                # carries ``occurred_at`` for the downstream recency
+                # ``_vector_search_chunks``. The Chunk's first-class
+                # ``occurred_at`` column feeds the downstream recency
                 # boost (RRF only uses rank position, so the score is
                 # informational).
                 filtered.append(
@@ -4437,19 +4427,14 @@ class VectorCypherRetriever:
                             namespace_id=chunk.namespace_id,
                             document_id=chunk.document_id,
                             content=chunk.content,
-                            metadata={
-                                "occurred_at": (
-                                    chunk.occurred_at.isoformat() if getattr(chunk, "occurred_at", None) else None
-                                ),
-                                **(getattr(chunk, "metadata", None) or {}),
-                            },
+                            metadata=getattr(chunk, "metadata", None) or {},
                             chunker_info=getattr(chunk, "chunker_info", None) or {},
                             created_at=getattr(chunk, "created_at", None) or getattr(chunk, "occurred_at", None),
                             occurred_at=getattr(chunk, "occurred_at", None),
                             # The temporal store carries the chunk event-time
-                            # (occurred_at, surfaced into metadata above) and the
-                            # producer time (source_timestamp); the projection
-                            # uses event-time first, then producer time.
+                            # (occurred_at) and the producer time
+                            # (source_timestamp); the projection uses
+                            # event-time first, then producer time.
                             source_timestamp=getattr(chunk, "source_timestamp", None),
                         ),
                     )
--- a/tests/integration/matrix/test_stored_blob_filter_invariant.py
+++ b/tests/integration/matrix/test_stored_blob_filter_invariant.py
@@ -1,0 +1,305 @@
+"""End-to-end cross-channel invariant: the stored metadata blob is what recall
+filters see, on every retrieval channel.
+
+The recall read paths carry the STORED chunk metadata blob verbatim — the
+first-class ``occurred_at`` column is never folded back into the blob at read
+time. This suite pins the user-observable consequence through a REAL recall:
+
+1. A chunk ingested with a producer ``source_timestamp`` (so its ``occurred_at``
+   column is populated) but a CLEAN metadata blob (no ``occurred_at`` key) is
+   selected identically by a ``metadata.occurred_at $exists false`` filter across
+   the vector, keyword/BM25, and hybrid (graph-bearing) recall channels. Before
+   the read paths stopped injecting the column into the blob, the graph channel's
+   in-memory post-filter saw an ``occurred_at`` key the SQL-pushdown channels
+   never saw — so the same filter kept the chunk on one channel and dropped it on
+   another. This is that cross-channel divergence, guarded end-to-end.
+
+2. The mirror ``metadata.occurred_at $exists true`` selects NOTHING — the blob
+   has no such key on any channel.
+
+3. A whole-blob ``$eq`` (the exact stored blob) selects the chunk — the in-memory
+   full-AST post-filter compares against the same stored blob the SQL pushdown
+   does.
+
+4. The raw ``khora_chunks`` blob is exactly what was written: user keys only, no
+   ``occurred_at`` / ``connected_entities`` / ``ppr_score`` injected. Reading the
+   stored row (rather than a recalled projection) also covers the skeleton engine
+   channel, whose rebuilt chunk is internal to the engine.
+
+Scope: sqlite_lance only — lightweight, no Docker. The embedded ``khora_chunks``
+table matches the production schema, so the write/read round-trip exercised here
+is faithful. VectorCypher's HYBRID recall fires the graph channel (an entity is
+extracted from the content); Skeleton is graph-less, so its HYBRID recall is the
+vector + BM25 legs.
+"""
+
+from __future__ import annotations
+
+import hashlib
+import json
+import sqlite3
+from collections.abc import AsyncIterator
+from datetime import UTC, datetime, timedelta
+from pathlib import Path
+from typing import Any
+from uuid import UUID
+
+import pytest
+
+try:
+    import aiosqlite  # noqa: F401
+    import lancedb  # noqa: F401
+
+    _HAS_EMBEDDED = True
+except ImportError:
+    _HAS_EMBEDDED = False
+
+from khora.config import KhoraConfig
+from khora.config.schema import SQLiteLanceConfig
+from khora.extraction.extractors.base import ExtractedEntity, ExtractionResult
+from khora.extraction.skills import ExpertiseConfig
+from khora.khora import Khora
+from khora.query import SearchMode
+
+EMBED_DIM = 32
+
+pytestmark = [
+    pytest.mark.embedded,
+    pytest.mark.integration,
+    pytest.mark.skipif(not _HAS_EMBEDDED, reason="aiosqlite/lancedb not installed"),
+]
+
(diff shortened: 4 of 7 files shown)


filippetrovic:

Six independent reviewers (security, performance, correctness, tests, docs, devil's advocate) plus a dedicated security audit. **Overall assessment: APPROVE / merge.** No CRITICAL or HIGH findings. The three MEDIUM findings (all test-quality) were addressed in the follow-up commit; LOW findings are informational or already handled.

### CRITICAL
_None._

### HIGH
_None._

### MEDIUM — all addressed in `8e257a5a`
- **M1 [Tests]** `tests/unit/engines/test_retriever_stored_blob_metadata.py` — the contract "a caller-stored `occurred_at`/`connected_entities`/`ppr_score` blob key is user data and passes through verbatim" had no positive coverage: every fixture omitted the keys and asserted absence, so a future over-correction that *scrubbed* those names from a user blob would have passed the whole suite. **Fixed:** added `TestCallerStoredKeysReturnedVerbatim` (graph + vector constructor paths) asserting caller-owned keys survive verbatim while the first-class column value stays distinct.
- **M2 [Tests]** `tests/unit/filter/test_filter_fuzz.py` — `TestStoredBlobColumnInvariant` docstring claimed it would catch a reintroduced recall-constructor injection; it actually exercises the SQL compiler + oracle over the raw row. **Fixed:** docstring reworded to its true scope (compiler/schema `$exists` + whole-blob `$eq`), with an explicit note it does not exercise the recall constructors.
- **M3 [Tests]** `tests/integration/matrix/test_stored_blob_filter_invariant.py` — module docstring overstated skeleton coverage (the raw-row read is the write path; skeleton's rebuilt `Chunk.metadata` is internal and discarded). **Fixed:** docstring corrected to describe the write-path precondition and the indirect/unobservable nature of the skeleton dead-write.

### LOW
- **L1 [Performance / Correctness]** — the returned `Chunk.metadata` is now the store-result dict by reference (`slots=True` dataclass, no copy) rather than a fresh `{**}` dict. Benign today (audited: no in-place mutation of a recalled chunk's metadata on any read path; stores already materialize a fresh dict per read; callers never receive the dict — `RecallChunk` has no metadata field). Invariant to preserve going forward; do **not** "fix" with `dict(...)` — that reintroduces the allocation this change removes. Informational, no action.
- **L2 [Correctness / Devil's Advocate]** — PR description imprecision ("response metadata" wording; blanket "rerank builders read first-class columns" vs. the opt-in `include_date_prefix` experiment). **Addressed:** PR body tightened to scope the change to the internal `Chunk.metadata` and note the non-production experiment.
- **L3 [Tests / Devil's Advocate]** — HYBRID integration cases don't assert the graph channel actually fired, so they could silently degrade to vector-only. Mitigated by the unit graph-post-filter agreement test (the faithful regression guard). Optional future hardening.
- **L4 [Devil's Advocate]** — test proportionality: some redundancy (graph-fallback exercised at three layers). Accepted as-is; the cross-channel invariant is subtle and regression-prone.

### Security Audit
**Verdict: PASS.** Pure in-memory metadata dead-write removal; no new trust boundary, sink, endpoint, secret, deserialization, or dependency change. Added-line scan surfaced only test-side static SQL over a fixture table and test method names — no findings. The change slightly *reduces* internal-state disclosure by dropping `connected_entities` (graph adjacency) and `ppr_score` (ranking mass) from response chunks.

### Why this is safe (reviewer consensus)
The injected keys lived only on the engine-internal `Chunk.metadata` and never reached the caller-facing `RecallChunk` (which has no metadata field), so no downstream consumer could have read them — repo-wide grep confirms zero readers. Stages #1510 (populate first-class columns) and #1513 (flip readers to columns) already landed, so temporal ranking reads columns, not the blob → genuinely zero ranking change. The graph channel's in-memory post-filter now evaluates the same stored blob the vector/BM25 SQL pushdown sees, restoring cross-channel filter invariance (the intended #1492(a) fix).

filippetrovic:
Reviewed against DYT-5461 / #1509. Spec conformance verified:

- All six injection sites deleted with stored-blob pass-through exactly as specified (`_simple_retrieve`, `_vector_search_chunks`, `_fetch_chunks_from_entities` covering both graph branches, `_recency_channel_chunks`, PPR re-wrap, skeleton re-wrap). First-class `occurred_at`/`source_timestamp` kwargs preserved at every site; no `connected_entities`/`ppr_score` writes remain anywhere in src.
- The two pinned-key assertions tightened to exact stored-blob equality + first-class column (`coverage_push` :1579/:1658 regions).
- Shape coverage is complete across all nine channels: new per-channel unit module (simple vector, BM25, storage graph fallback, recency, PPR, typed-entity fast path) + coverage_push (vector full, graph Neo4j) + matrix lane (skeleton, end-to-end). The adversarial `TestCallerStoredKeysReturnedVerbatim` pins the never-scrub direction with deliberately un-column-like values — good.
- Filter invariants pinned at three layers: matrix e2e (`$exists` false/true + whole-blob `$eq` across vector/keyword/hybrid on both engines), compiler/schema fuzzer seam (oracle-vs-lance agreement, column-never-coalesces-into-blob), and raw `khora_chunks` write-path cleanliness. Notably the `$exists: true` mirror is the test that would have caught the original #1492 divergence end-to-end (old code: graph post-filter accepts the injected key → chunk surfaces where SQL channels dropped it); the direct graph-channel constructor pins cover the direction fusion could mask.
- Zero-ranking-change claim holds: stage 2 + the prefix ungating left no blob readers on the production path; the opt-in `CrossEncoderReranker(include_date_prefix=True)` experiment reads user-space ingest keys (`metadata.custom.occurred_at`/`sent_at`) — sanctioned reads of user data, correctly out of scope.
- Body carries `Fixes #1492` + `Closes #1509`. CI green.

No changes required. Merge-ready. Merging closes the #1492 umbrella — stage chain complete.

probablyrobot (approved):

PR #1525, opened 2026-07-16
Render prompt templates in a Jinja sandboxed environment

filippetrovic:
## Summary
- Extraction/expertise prompts (`ExpertiseComposer.render_prompt`) and persona chat prompts (`PromptGenerator`) previously rendered through a raw, non-sandboxed `jinja2.Template`. Any caller able to influence a prompt string could execute arbitrary code in-process on the host (server-side template injection → RCE).
- Both sinks now render through a shared module-level `jinja2.sandbox.ImmutableSandboxedEnvironment`. Loops, filters, conditionals, and normal attribute access on data render identically; only dunder-traversal / code-execution gadgets are refused.
- **Behavior change (fail closed):** a blocked construct now raises `jinja2.exceptions.SecurityError` instead of the previous fail-open path that logged a warning and returned the raw, un-rendered template. The composer and all three extractor render call sites re-raise `SecurityError` so a malicious or malformed prompt fails loudly rather than silently degrading to a default/un-rendered prompt. Non-security render errors keep their existing fallback behavior.
- Dropped a now-dead `ImportError` fallback in the composer (jinja2 is a hard dependency).

## Test plan
- [x] Unit tests pass (`tests/unit/test_prompt_sandbox.py`, 11 tests): SSTI proof payload raises `SecurityError` and returns no interpolated output; legit `{% for %}` + `| join` prompt renders byte-identically; persona `{% if history_summary %}` renders with and without history; `SecurityError` propagates and is not downgraded to the raw template (composer- and extractor-caller level); all three builtin expertise skills render without error and idempotently.
- [x] Regression suites pass (`test_chat_prompt`, `test_expertise`, `test_extraction_llm`, persona/passthrough): 124 passed combined.
- [x] `ruff check` / `ruff format` clean on touched files.
- [ ] CI full suite

## Summary by CodeRabbit

- **Security**
  - Prompt templates are now rendered in a hardened Jinja sandbox to block unsafe expressions and mitigate SSTI-to-RCE risks.
  - When unsafe constructs are detected, processing fails closed by raising `SecurityError` instead of falling back to a default/less-restricted prompt.
  - Safe, legitimate prompt features and prior handling for non-security rendering errors remain supported.
- **Tests**
  - Added unit coverage to verify `SecurityError` propagation and correct rendering for persona, expertise, and extractor prompt flows.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@ Format: versions match git tags (`git tag vX.Y.Z`). Versions before 0.5.1 were i
 
 ## [Unreleased]
 
+### Security
+
+- **Prompt templates now render in a Jinja `ImmutableSandboxedEnvironment`**, closing an SSTI -> RCE vector: extraction/expertise prompts (`ExpertiseComposer.render_prompt`) and persona chat prompts (`PromptGenerator`) previously rendered through a raw, non-sandboxed `jinja2.Template`. Templates that attempt unsafe constructs now raise `jinja2.exceptions.SecurityError` instead of failing open and returning the raw template string. The LLM extractor prompt-render call sites fail closed on `SecurityError` (propagate rather than silently degrading to the default prompt), while genuine non-security render errors keep their existing fallback behavior.
+
 ## [0.22.2] - recall-quality correctness, HNSW index-backed search, durable reconcilers
 
 ### Added
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,10 +6,6 @@ Format: versions match git tags (`git tag vX.Y.Z`). Versions before 0.5.1 were i
 
 ## [Unreleased]
 
-### Security
-
-- **Prompt templates now render in a Jinja `ImmutableSandboxedEnvironment`**, closing an SSTI -> RCE vector: extraction/expertise prompts (`ExpertiseComposer.render_prompt`) and persona chat prompts (`PromptGenerator`) previously rendered through a raw, non-sandboxed `jinja2.Template`. Templates that attempt unsafe constructs now raise `jinja2.exceptions.SecurityError` instead of failing open and returning the raw template string. The LLM extractor prompt-render call sites fail closed on `SecurityError` (propagate rather than silently degrading to the default prompt), while genuine non-security render errors keep their existing fallback behavior.
-
 ## [0.22.2] - recall-quality correctness, HNSW index-backed search, durable reconcilers
 
 ### Added
--- a/docs/extraction/expertise-system.md
+++ b/docs/extraction/expertise-system.md
@@ -371,6 +371,8 @@ inference_rules:
 
 Prompts support Jinja2 templating:
 
+Prompts render in a Jinja `ImmutableSandboxedEnvironment`; unsafe constructs (dunder/private attribute access, mutating methods) are rejected and raise `SecurityError`.
+
 ```yaml
 system_prompt: |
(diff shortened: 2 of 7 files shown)


filippetrovic:
## Multi-perspective review — PR #1525 (render prompt templates in a Jinja sandboxed environment)

Reviewers: Security, Correctness+Performance, Test Quality, AI/Prompt, Docs, Devil's Advocate. All ran independently against the PR diff.

**Overall assessment: APPROVE.** No CRITICAL or HIGH findings. The fix correctly closes the SSTI→RCE surface at both raw-`jinja2.Template` sinks by rendering through a shared `ImmutableSandboxedEnvironment`, and makes blocked templates fail closed (`SecurityError` propagates at the composer and all three `extraction/extractors/llm.py` call sites) instead of failing open. Fail-closed propagation was traced end-to-end to the `extract()` and batch entry points. Sandbox effectiveness was verified empirically against the PoC gadget (jinja2 3.1.6). No legitimate prompt (3 builtin skills, persona default, documented `{{ }}` / `| join` / `{% if %}` / attribute-read constructs) changes output; autoescape stays off, so no prompt corruption.

### MEDIUM

- **M1 (Test) — batch-path / extraction-prompt caller not covered.** AC#4 names the batch path explicitly, but only `_render_system_prompt` has a caller-level fail-closed test. The `_render_extraction_prompt` and `_extract_multi_batch` `except SecurityError: raise` blocks are untested; a regression in either would go uncaught. Addressed in a follow-up commit (caller-level tests for both sites).

### LOW

- **L1 (Devil's Advocate) — chat sink logs nothing on block.** The composer + 3 extractor sinks log a distinct warning before re-raising `SecurityError`; the persona chat sink (`chat/prompt.py`) propagates it silently. Added warn+re-raise parity so a blocked SSTI on the chat path is detectable.
- **L2 (Docs) — sandbox constraint not documented for prompt authors.** `docs/extraction/expertise-system.md` and `docs/extraction/extractors.md` describe Jinja support but not the new sandbox / `SecurityError` fail-closed behavior. Added a one-line note to each.
- **L3 (Correctness) — no compiled-template cache in `render_prompt`.** `from_string` recompiles per call on the extraction hot path — identical cost to the prior per-call `Template(...)`, so not a regression. Optional future optimization; out of scope.

### Follow-up (out of this ticket's scope)

- **F1 (Devil's Advocate) — template DoS amplification not closed.** `ImmutableSandboxedEnvironment` stops code execution but not allocation amplification (`{{ "a" * 10**9 }}`, nested loops) — same template-author trust boundary. Explicitly deferred by the ticket; recommend a separate tracking issue rather than expanding this PR.

### Confirmed non-issues (raised and dismissed)

- Batch abort on a poisoned expertise is the correct blast radius — untrusted document text and chat history are passed as Jinja *variables*, never as the template, so they can never trigger `SecurityError`; only the operator-authored template can. "Isolate one bad doc" is not meaningful.
- Dropping the composer's `except ImportError` branch is safe — jinja2 is a hard core dependency (`pyproject.toml`), and `chat/prompt.py` already imported it unconditionally.
- Shared module-level `_SANDBOX_ENV` is concurrency-safe — Jinja environments are designed for concurrent rendering; no globals/filters are mutated.
- No telemetry-contract update owed (only `logger.warning` lines added; no new span/metric); re-raising branches are not ADR-001 silent degradations.

_Consolidated from an independent multi-perspective agent review._

sale87 (approved):

PR #1547, opened 2026-07-21
Make entity attribute validation additive-safe

### LINKED ISSUE #1546: Models: make validate_attributes additive-safe so it never drops ontology attribute keys
Part of #1541.

## Summary

`validate_attributes` (`src/khora/core/models/schemas.py:151–173`) validates against a small hardcoded registry of Pydantic models (`ATTRIBUTE_SCHEMAS`: PERSON, ORGANIZATION, LOCATION, CONCEPT, EVENT, TECHNOLOGY, PRODUCT, DATE, STATE_CHANGE) and returns `model_dump(exclude_none=True)` — which **silently drops any attribute not declared on that hardcoded model**. For a registered type like PERSON, ontology attributes such as `slack_user_id` / `timezone` / `linear_user_id` would be discarded. It is currently harmless only because `Entity.validate()` (`src/khora/core/models/entity.py:74–78`) is dead code (never called on the ingest path) — a latent hazard: if anyone wires it, ontology attributes vanish. See #1541.

## Fix — make validation additive-safe

Make `validate_attributes` **coerce known fields without ever dropping unknown ones**, so it is safe to run regardless of whether it is wired:

```python
schema = ATTRIBUTE_SCHEMAS.get(entity_type.upper())
if not schema:
    return attributes
try:
    validated = schema.model_validate(attributes).model_dump(exclude_none=True)
    # preserve ontology/extra keys the hardcoded schema doesn't declare
    return {**{k: v for k, v in attributes.items() if v is not None}, **validated}
except ValidationError:
    return attributes  # unchanged: graceful passthrough
```

Do **not** wire `Entity.validate()` into the write path (out of scope) — attributes continue to flow verbatim through ingest. This ticket only removes the drop-hazard from the helper and documents the intent.

Add a one-line note at `Entity.validate()` that it is intentionally off the ingest path and that `validate_attributes` is additive-safe.

## Acceptance criteria

- [ ] `validate_attributes("PERSON", {"name": "X", "slack_user_id": "U1", "timezone": "CET"})` returns a dict still containing `slack_user_id` and `timezone`.
- [ ] `validate_attributes("LINEAR_ISSUE", {...})` (type not in the registry) returns the input unchanged.
- [ ] Known fields are still coerced; `ValidationError` still falls back to unchanged passthrough.
- [ ] `Entity.validate()` remains uncalled on the ingest path (behavior unchanged); note added.

filippetrovic:
## Summary
- `validate_attributes` previously returned `model_dump(exclude_none=True)` from the matched attribute schema, which silently dropped any attribute key not declared on the hardcoded model (e.g. a PERSON's `slack_user_id` / `timezone`).
- It now merges the schema-coerced known fields *over* the caller's non-None attributes, so known keys are still coerced/normalized while unknown (e.g. ontology-defined) keys are preserved.
- Unregistered entity types and `ValidationError` cases still pass the original attributes through unchanged (graceful degradation preserved).
- Added a note on `Entity.validate()` clarifying it is intentionally not wired into the ingest path and that `validate_attributes` is additive-safe. This is latent-hazard hardening — no ingest-path behavior changes today.

## Test plan
- [x] Unit tests pass (`tests/unit/test_attribute_schemas.py` — 17 passed): unknown keys preserved, unregistered type unchanged, known-field coercion still applies, `ValidationError` returns original attributes, and the None-known-field-stripped-while-unknown-key-preserved intersection case.
- [ ] Integration tests pass (CI — Docker not available in local sandbox)
- [x] Manual verification: additive merge confirmed via pydantic coercion probes; no new call sites added, `Entity.validate()` remains uncalled in `src/`.

Fixes #1546

## Summary by CodeRabbit

* **Bug Fixes**
  * Preserved additional entity attributes during validation instead of silently removing them.
  * Continued converting recognized fields to their expected formats.
  * Retained original attributes when validation fails or no schema is registered.
  * Removed fields with empty (`None`) values while preserving other attributes.

* **Tests**
  * Added coverage for attribute preservation, coercion, and validation-failure behavior.

### DIFF
--- a/src/khora/core/models/entity.py
+++ b/src/khora/core/models/entity.py
@@ -75,8 +75,8 @@ class Entity:
         """Validate and clean attributes using the registered schema for this entity type.
 
         Note: intentionally NOT called on the ingest path. ``validate_attributes``
-        is additive-safe (it never drops ontology attribute keys), so calling this
-        is optional cleanup rather than a required write-path step.
+        is additive-safe (it never drops non-None ontology attribute keys), so
+        calling this is optional cleanup rather than a required write-path step.
         """
         from khora.core.models.schemas import validate_attributes
 
--- a/src/khora/core/models/entity.py
+++ b/src/khora/core/models/entity.py
@@ -72,7 +72,12 @@ class Entity:
             self.source_tool = ""
 
     def validate(self) -> None:
-        """Validate and clean attributes using the registered schema for this entity type."""
+        """Validate and clean attributes using the registered schema for this entity type.
+
+        Note: intentionally NOT called on the ingest path. ``validate_attributes``
+        is additive-safe (it never drops ontology attribute keys), so calling this
+        is optional cleanup rather than a required write-path step.
+        """
         from khora.core.models.schemas import validate_attributes
 
         self.attributes = validate_attributes(self.entity_type, self.attributes)
--- a/src/khora/core/models/schemas.py
+++ b/src/khora/core/models/schemas.py
@@ -149,25 +149,31 @@ def register_attribute_schema(
 
 
 def validate_attributes(entity_type: str, attributes: dict[str, Any]) -> dict[str, Any]:
-    """Validate and coerce attributes using the registered schema.
+    """Validate and coerce attributes using the registered schema (additive-safe).
 
-    Falls back to passthrough if no schema is registered for the entity type
-    or if validation fails (graceful degradation).
+    Coerces the fields the registered schema knows about while preserving any
+    unknown keys with non-None values (e.g. ontology-specific attributes)
+    rather than dropping them. Falls back to passthrough if no schema is
+    registered for the entity type or if validation fails (graceful
+    degradation).
 
     Args:
         entity_type: The entity type name (e.g. "PERSON", "TICKET")
         attributes: Raw attributes dict to validate
 
     Returns:
-        Cleaned attributes dict with None values excluded
+        On the validated path: schema-known fields coerced and all non-None
+        unknown keys preserved, with None-valued keys excluded (unchanged from
+        prior behavior). On the degraded paths (unregistered type or
+        ValidationError) the original attributes are returned unchanged.
     """
     schema = ATTRIBUTE_SCHEMAS.get(entity_type.upper())
     if not schema:
         return attributes
 
     try:
-        validated = schema.model_validate(attributes)
-        return validated.model_dump(exclude_none=True)
+        validated = schema.model_validate(attributes).model_dump(exclude_none=True)
+        return {**{k: v for k, v in attributes.items() if v is not None}, **validated}
     except ValidationError:
         # Graceful degradation: return original attributes if validation fails
         return attributes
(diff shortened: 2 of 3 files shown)


filippetrovic:

Reviewed across six lenses: **Correctness, Security, Performance, Test, Docs, Devil's Advocate**.

**Overall assessment: ✅ APPROVE.** No CRITICAL / HIGH / MEDIUM defects. The change is small (3 files, ~50 net lines), surgical, and well-tested; it faithfully preserves the graceful-degradation contract while making attribute validation additive-safe.

### Highlights
- **Merge order is correct.** `{**caller_non_none, **validated}` spreads the schema-coerced values last, so known-field coercion always wins on collision (verified via the `CONCEPT.related_concepts` tuple→list test), while unknown/ontology keys survive.
- **Degraded paths unchanged.** Unregistered types and `ValidationError` still return the original attributes verbatim; `model_dump` stays inside the `try`, so exception coverage is identical to before.
- **No new runtime surface.** The only caller is `Entity.validate()`, which is uncalled anywhere in `src/` — this is latent-hazard hardening, exactly as intended; no ingest-path behavior changes today.
- **Not over-engineered.** The `if v is not None` base-dict filter is load-bearing (the naive `{**attributes, **validated}` would leak None-valued unknown keys); the complexity is justified.

### Findings

| ID | Severity | Lens | Finding | Status |
|----|----------|------|---------|--------|
| L1 | LOW | Docs | `Entity.validate()` note said "never drops ontology attribute keys" — too strong, since None-valued keys are excluded. | ✅ Fixed — tightened to "non-None". |
| L2 | LOW | Test | The load-bearing `if v is not None` filter for a None-valued **unknown** key on the validated path was untested. | ✅ Fixed — added `test_none_unknown_key_dropped_on_validated_path` (suite now 18 tests). |
| L3 | LOW | Security | Semantics shift from implicit allowlist → coercion-with-passthrough would widen keys reaching storage *if* `validate()` were ever wired to ingest. Safe under the standard parameter-map binding pattern; not exercised by this diff. | Noted — no action. |
| L4 | LOW | Correctness / DA | Downstream schemas using `extra="forbid"` fall to the passthrough branch (skipping coercion), and non-None schema defaults get injected. Both are **pre-existing** behaviors, unchanged by this PR; no built-in schema is affected. | Noted — no action. |
| L5 | LOW | Correctness | Shallow-merge means the success path shares value references with the caller's dict — standard Python dict-merge semantics; the sole caller reassigns, so no impact. | Noted — no action. |
| M1 | MEDIUM (judgment) | Docs | Consider a CHANGELOG entry since `register_attribute_schema` is a documented downstream extension point. | **Deliberately out of scope** — `validate_attributes` is in no formal public `__all__` export, there is no in-repo behavior change (dead path), and the change was scoped to the two source files + tests. Recording the decision here rather than silently dropping it. |

The two actionable LOW nits (L1, L2) were addressed in `98018110`. Everything else is informational / pre-existing.

_Automated multi-perspective review._

probablyrobot (approved):

PR #1548, opened 2026-07-21
Merge entity attributes across chunks in pipeline dedup

### LINKED ISSUE #1544: Ingestion: merge entity attributes across chunks in pipeline dedup instead of discarding them
Part of #1541.

## Summary

When the same entity (`name`, `entity_type`) recurs across chunks during staging, the in-pipeline dedup merges `mention_count` and source ids but **discards later chunks' `attributes`** — the first chunk's attributes win and any attributes found in subsequent chunks are lost. Independent of the extraction fix (#1541): valuable for any config that produces attributes (non-allowlisted models, custom prompts) and directly compounds the loss once A1 lands.

## Where

The existing-key merge branches that update counts/sources but not attributes:

- `src/khora/pipelines/tasks/extract.py` (~lines 307–328)
- `src/khora/pipelines/flows/ingest.py` (~lines 361–377)

A correct attribute merge already exists in `Entity.merge_with` (`src/khora/core/models/entity.py:80–107`, "prefer existing, add new"), but it is **not** on this hot path.

## Fix

In **both** merge branches, after the existing count/source updates, union attributes into the existing entity — **prefer the existing value, fill only missing keys, skip `None`/empty**:

```python
existing_attrs = existing.attributes if isinstance(existing.attributes, dict) else {}
incoming_attrs = new.attributes if isinstance(new.attributes, dict) else {}
for k, v in incoming_attrs.items():
    if v in (None, "") :
        continue
    existing_attrs.setdefault(k, v)
existing.attributes = existing_attrs
```

Keep it a surgical inline merge (do **not** swap in `Entity.merge_with` here — it merges aliases/other fields too and would widen behavior beyond attributes).

## Acceptance criteria

- [ ] Both merge branches union attributes with existing-preferred / add-missing semantics.
- [ ] Unit test: two `ExtractedEntity` with the same `(name, type)` across chunks — first `{email: ...}`, second `{title: ...}` → merged entity has **both**; on a key present in both, the **existing** value wins; empty/`None` incoming values are skipped.

filippetrovic:
## Summary
- When the same `(name, entity_type)` entity recurs across chunks, the in-pipeline dedup updated mention counts and source ids but **discarded later chunks' `attributes`** — the first chunk's attributes won and any attributes found in later chunks were lost.
- Union attributes inline at both dedup call sites (`extract_entities` in `pipelines/tasks/extract.py` and `stream_extract_and_embed_entities` in `pipelines/flows/ingest.py`) with **existing-preferred / add-missing / skip-None-or-empty-string** semantics: a key already present keeps its existing value, missing keys are filled from later chunks, and `None`/empty-string incoming values are skipped.
- Kept the merge surgical and inline — deliberately did not reuse the broader `Entity.merge_with` (it also merges aliases and other fields, out of scope here).

## Test plan
- [x] Unit tests pass — new `tests/unit/pipelines/test_attribute_merge_dedup_1544.py` drives the real dedup path at **both** call sites and asserts: union (both keys present), existing value wins on a shared key, and `None`/empty-string incoming values are skipped.
- [x] `make format` clean; `make lint` passes (ruff clean, no new type findings in changed files).
- [ ] Full `make test` (unit + integration + e2e, coverage gate) — validated in CI (requires `lancedb` + Docker Postgres/Neo4j not available in the dev sandbox).

Fixes #1544

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved entity deduplication by merging attributes across extraction chunks when multiple mentions resolve to the same entity.
  * Preserves already-known attribute values and fills only missing keys from later chunks.
  * Skips attributes that are `None` or empty strings while keeping other falsy-but-valid values (e.g., `0`).

* **Tests**
  * Added unit coverage to verify attribute-union behavior in both extraction and ingest pipelines.

### DIFF
--- a/src/khora/pipelines/flows/ingest.py
+++ b/src/khora/pipelines/flows/ingest.py
@@ -375,6 +375,14 @@ async def stream_extract_and_embed_entities(
                         _vu = _parse_temporal_date(_t.valid_until) if _t else None
                         if _vu and (not existing.valid_until or _vu > existing.valid_until):
                             existing.valid_until = _vu
+                        # Union attributes across chunks: prefer existing, fill missing, skip empty (#1544)
+                        existing_attrs = existing.attributes if isinstance(existing.attributes, dict) else {}
+                        incoming_attrs = extracted.attributes if isinstance(extracted.attributes, dict) else {}
+                        for _k, _v in incoming_attrs.items():
+                            if _v in (None, ""):
+                                continue
+                            existing_attrs.setdefault(_k, _v)
+                        existing.attributes = existing_attrs
                     else:
                         entity_type = extracted.entity_type or "CONCEPT"
 
--- a/src/khora/pipelines/flows/ingest.py
+++ b/src/khora/pipelines/flows/ingest.py
@@ -375,13 +375,15 @@ async def stream_extract_and_embed_entities(
                         _vu = _parse_temporal_date(_t.valid_until) if _t else None
                         if _vu and (not existing.valid_until or _vu > existing.valid_until):
                             existing.valid_until = _vu
-                        # Union attributes across chunks: prefer existing, fill missing, skip empty (#1544)
+                        # Union attributes across chunks: prefer existing non-empty, fill
+                        # missing or empty from later chunks, skip None/empty incoming (#1544)
                         existing_attrs = existing.attributes if isinstance(existing.attributes, dict) else {}
                         incoming_attrs = extracted.attributes if isinstance(extracted.attributes, dict) else {}
                         for _k, _v in incoming_attrs.items():
                             if _v in (None, ""):
                                 continue
-                            existing_attrs.setdefault(_k, _v)
+                            if existing_attrs.get(_k) in (None, ""):
+                                existing_attrs[_k] = _v
                         existing.attributes = existing_attrs
                     else:
                         entity_type = extracted.entity_type or "CONCEPT"
--- a/src/khora/pipelines/tasks/extract.py
+++ b/src/khora/pipelines/tasks/extract.py
@@ -326,6 +326,14 @@ async def extract_entities(
                     merge_valid_from = chunk.source_timestamp
                 if existing.valid_from and merge_valid_from is not None and merge_valid_from < existing.valid_from:
                     existing.valid_from = merge_valid_from
+                # Union attributes across chunks: prefer existing, fill missing, skip empty (#1544)
+                existing_attrs = existing.attributes if isinstance(existing.attributes, dict) else {}
+                incoming_attrs = extracted.attributes if isinstance(extracted.attributes, dict) else {}
+                for _k, _v in incoming_attrs.items():
+                    if _v in (None, ""):
+                        continue
+                    existing_attrs.setdefault(_k, _v)
+                existing.attributes = existing_attrs
             else:
                 # Create new entity — preserve original type string from LLM
                 entity_type = extracted.entity_type or "CONCEPT"
--- a/src/khora/pipelines/tasks/extract.py
+++ b/src/khora/pipelines/tasks/extract.py
@@ -326,13 +326,15 @@ async def extract_entities(
                     merge_valid_from = chunk.source_timestamp
                 if existing.valid_from and merge_valid_from is not None and merge_valid_from < existing.valid_from:
                     existing.valid_from = merge_valid_from
-                # Union attributes across chunks: prefer existing, fill missing, skip empty (#1544)
+                # Union attributes across chunks: prefer existing non-empty, fill
+                # missing or empty from later chunks, skip None/empty incoming (#1544)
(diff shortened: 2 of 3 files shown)


filippetrovic:
## 🔍 Multi-perspective PR review

Consolidated review from a 6-reviewer team (security, performance, correctness, tests, docs, devil's advocate) plus a dedicated security audit. Reviewed at commit `6608612`.

**Overall assessment: ✅ APPROVE.** No CRITICAL or HIGH findings. The change is correct, surgical, and matches the resolved design (union entity attributes across chunks in the in-pipeline dedup at both call sites, existing-preferred / add-missing / skip-None-or-empty). Items below are one MEDIUM design note (an accepted consequence of the "prefer existing" decision) and LOW/follow-up polish.

### MEDIUM
- **M1 — [correctness / devil's advocate] An *existing empty* value shadows a later non-empty one.** `existing_attrs.setdefault(k, v)` is a no-op when the key already exists, even if the existing value is `""`/`None`. If the first chunk seeds `{"role": ""}` and a later chunk has `{"role": "engineer"}`, the real value is not adopted. This is the deliberate, resolved semantic ("prefer existing value; skip empty *incoming*") — the skip guard only applies to incoming values, not existing ones. Flagging it explicitly so it's a conscious tradeoff rather than an accidental one. **No change in this PR** (it would contradict the agreed semantics); reasonable follow-up if "fill empty existing from a later non-empty chunk" is ever desired — either overwrite-when-existing-is-empty in the merge, or strip empties on the create branch too.

### LOW
- **L1 — [devil's advocate] Divergent attribute-merge semantics across call sites.** These two new inline merges skip empty incoming values; the pre-existing `Entity.merge_with` (`core/models/entity.py:93`) and the cross-tool merge in `entity_resolution.py` do a key-fill *without* the empty skip. Whether an empty attribute survives now depends on which dedup path touched the entity. Reusing `merge_with` here isn't clean (the loop holds an `ExtractedEntity`, not an `Entity`), so the inline copy is justified — but a future refactor unifying all attribute-union logic behind one helper would remove the divergence. Follow-up candidate, not a blocker.
- **L2 — [security] No per-entity attribute-key cap.** For an entity recurring across N chunks, the merge now accumulates the union of distinct keys (previously only the first chunk's). Growth is per-distinct-key and bounded by document size + chunking, so it's a modest linear amplification, not a blowup — and there was no cap before either. Optional defensive hardening for untrusted multi-tenant ingest (cap distinct keys + record an ADR-001 `Degradation`); not required to merge.
- **L3 — [tests] ✅ Addressed in `6608612`.** The skip predicate is `_v in (None, "")` (not `if not _v`), so legitimate falsy values (`0`, `0.0`, `False`) are correctly preserved — but that subtle, refactor-fragile behavior was untested. Added a regression guard: chunk 2 now carries `"count": 0` and both call-site tests assert it survives the merge.

### Security audit
**Verdict: PASS.** No dangerous-sink patterns in the added lines across all 16 categories (no injection, crypto, secret handling, network/file/redirect, deserialization, mass-assignment, or regex surface); no dependency/manifest changes. The merged values are LLM-extracted attributes already written verbatim by the pre-existing create branch (and NUL-stripped upstream), so no new trust boundary or sink is introduced.

### Notes (INFO)
- **Performance:** negligible — O(attributes-per-mention) on the hot path, strictly cheaper than the existing per-mention source-id list scans in the same branch; `Entity` is a `slots=True` dataclass with no validate-on-assign.
- **Docs:** clean — the only docs describing attribute merge cover the separate, unchanged `Entity.merge_with` path and already state "prefer existing." No CHANGELOG entry expected in this PR (this repo adds `### Fixed` lines in the release-bump PR, not per-feature). No telemetry-contract / public-`__all__` impact.
- **Tests:** both call sites (`extract_entities` and `stream_extract_and_embed_entities`) are driven through their real dedup paths and fail if the merge block is removed — not a false pass.
- **ADR-084 secret-typing:** N/A — pure dict-merge logic, no new Pydantic/settings fields.

_Reviewers found no blocking issues; M1 is a documented, accepted consequence of the resolved design._

probablyrobot (approved):

filippetrovic:
### Update — M1 addressed (commit `7bd7d13`)

Resolved the MEDIUM finding from the review: an *existing empty* (`""`/`None`) attribute value no longer shadows a later chunk's non-empty value.

The merge now fills a key when the existing value is **missing or empty**, while still preferring existing **non-empty** values and skipping incoming empties:

```python
for _k, _v in incoming_attrs.items():
    if _v in (None, ""):
        continue
    if existing_attrs.get(_k) in (None, ""):
        existing_attrs[_k] = _v
```

Applied at both call sites (`extract.py`, `ingest.py`). The unit test now covers it: chunk 1 seeds `phone=""` / `fax=None`, chunk 2 provides non-empty values, and both call-site tests assert the non-empty values are adopted — alongside the existing checks (existing non-empty wins, missing key filled, incoming empty/None skipped, falsy `0` preserved).

Issue #1543, opened 2026-07-21
Extraction: feed the ontology's per-type attribute schema into the extraction prompt (ungated)


filippetrovic:
Part of #1541. Depends on the emission channel added in the A1 sub-issue (attributes must be emittable before naming which keys to emit is useful).

## Summary

Once the extractor can emit attributes (A1), it should be told **which** attributes each entity type expects, so emitted keys converge on the ontology's `required`/`optional` instead of whatever the model invents. The per-type attribute schema already lives on `EntityTypeConfig.attributes`, but the only code that surfaces it into the prompt (`_build_tool_context`) is gated behind `tool_schemas` + `context["source_tool"]` — unreliable, because adaptive batches mix sources and a per-batch `source_tool` is often absent or wrong. See #1541.

## Where

`src/khora/extraction/extractors/llm.py`:

- `_build_tool_context` (~lines 1779–1791) — builds an `EXPECTED ENTITY ATTRIBUTES:` block from `et.attributes`, but only after the early-return guard `if not expertise or not expertise.tool_schemas or not context: return ""` and a `context["source_tool"]` match.
- `_render_extraction_prompt` (single) and the batch prompt assembly (~lines 2340/2359, `tool_prefix`).

## Fix

1. Add `_build_attribute_schema_block(expertise, entity_types) -> str`: for each type **in the resolved `entity_types` list** that has attributes, emit one line `TYPE: required=[...]; optional=[...]` under a header `ATTRIBUTE SCHEMA (emit these keys in attributes when present):`. Return `""` when `expertise` is `None` or no extracted type declares attributes. **Not** gated on `tool_schemas` / `source_tool`.
2. Inject this block into the extraction prompt on **both** paths (single `_render_extraction_prompt` and the multi/batch assembly), alongside the existing tool/document context.
3. Remove the attribute-listing lines from `_build_tool_context` (the `EXPECTED ENTITY ATTRIBUTES` block, ~1779–1791) so attributes are not listed twice; leave the tool-field portion of `_build_tool_context` intact.

Keys are hints (v1) — the model is asked to prefer them, not constrained to them. Extraction still uses the A1 key/value-pair channel.

## Acceptance criteria

- [ ] With an `expertise` that declares attributes, the rendered prompt (single and batch) contains the `ATTRIBUTE SCHEMA` block listing `required`/`optional` for the extracted types, **regardless of `context`/`source_tool`** (present even when `context is None`).
[...]

PR #1549, opened 2026-07-21
Emit entity attributes on the strict structured-output path

### LINKED ISSUE #1542: Extraction: let the LLM extractor emit entity attributes (strict structured-output schema + prompt + parser)
Part of #1541.

## Summary

On the strict structured-output path (every model in `LLMEntityExtractor.MODELS_REQUIRING_JSON_SCHEMA`, including the default `gpt-4o-mini`), the extractor can never return entity `attributes`: the response JSON schema omits the field and the prompt never asks for it. This is the keystone fix — with it, directly-extracted entities can carry attributes at all (today only synthesized `EVENT` entities do). See #1541 for the full root-cause writeup.

## Where

`src/khora/extraction/extractors/llm.py`:

- `_get_response_format` (single-doc) and `_get_multi_response_format` (batch) — entity item schema: `required: [name, entity_type, description, aliases, temporal]`, `additionalProperties: false`, `strict: true`. **No `attributes` property.**
- `EXTRACTION_PROMPT_STRUCTURED` — used for allowlisted models; does not mention attributes.
- `_parse_response` — already reads `e.get("attributes", {})`.

## Fix

OpenAI strict mode cannot express an open-ended map (`additionalProperties` must be `false` and every key declared), so represent attributes as an **array of `{key, value}` string pairs** and fold it back to a dict on parse.

1. **Response schema (both `_get_response_format` and `_get_multi_response_format`):** add to each entity item's `properties`:
   ```json
   "attributes": {
     "type": "array",
     "items": {
       "type": "object",
       "properties": { "key": {"type": "string"}, "value": {"type": "string"} },
       "required": ["key", "value"],
       "additionalProperties": false
     }
   }
   ```
   and add `"attributes"` to that item's `required` array (strict mode requires every declared property to be required; the array may be empty).
2. **Prompt (`EXTRACTION_PROMPT_STRUCTURED`):** add one instruction to populate `attributes` as `{key, value}` pairs from the entity's salient fields (identifiers, emails, state, urls, dates, …). Keep it generic — the per-type key list is A2 (#1541).
3. **Parser (`_parse_response`):** accept **both** shapes — a list of `{key, value}` dicts → fold to `{k: v}`; an already-dict `attributes` (non-allowlisted / `json_object` models and the legacy `EXTRACTION_PROMPT`) → keep as-is; anything else → `{}`. Preserve the existing non-dict defensive coercion.

Leave `EXTRACTION_PROMPT` (non-structured, dict example) unchanged — the parser handles both shapes.

Non-goals: per-entity-type typed attribute schemas (infeasible under strict mode and would balloon token cost); typed/list attribute values (a list value is emitted as a string in v1).

## Acceptance criteria

- [ ] Both `_get_response_format` and `_get_multi_response_format` include an `attributes` key/value-pair array on the entity item, `strict`-valid (all properties `required`, `additionalProperties: false`).
- [ ] `_parse_response` folds the pairs array into a dict **and** still accepts the legacy dict form; a mocked-LLM extraction returning pairs yields a populated `entity.attributes`.
- [ ] Unit tests cover: strict-schema validity, pairs→dict fold, dict passthrough, empty/absent → `{}`.

filippetrovic:
## Summary
- On the strict `json_schema` extraction path (used by models like the default `gpt-4o-mini`), the response schema omitted `attributes` and the structured prompt never asked for them, so every directly-extracted entity was persisted with empty attributes.
- Add an `attributes` field to the entity item in **both** the single-doc (`_get_response_format`) and multi-section (`_get_multi_response_format`) strict response schemas. Because open maps are illegal under `strict`, attributes are modeled as an array of `{key, value}` string pairs; the field is added to the item's `required` list so the schema stays strict-valid (every property required, `additionalProperties: false`).
- Add one line to the structured extraction prompt instructing the model to emit `attributes` as `{key, value}` pairs from an entity's salient fields (identifiers, emails, state, urls, dates).
- Fold the pairs back into a dict on parse, while still accepting the legacy dict shape and collapsing anything else to `{}`.
- The non-structured prompt (dict-example) is unchanged.

Fixes #1542

## Behavior change (reviewer awareness)
Before this change, strict-schema models always produced `attributes == {}`, which left the downstream `STATE_CHANGE` handler (`pipelines/tasks/extract.py`) effectively dormant for those models. Now that entities can carry attributes (e.g. `entity_affected` / `new_state` / `transition_date`), that handler activates on the strict path and can drive bi-temporal `SUPERSEDES` version creation. This is the intended benefit of populating attributes, but it is a behavior change beyond "just emit attributes." The `attributes` public type (`dict[str, Any]`) is unchanged, so no public API contract is affected.

## Follow-ups (out of scope here)
- Attribute-value **grounding**: the prompt currently nudges the model to fill attributes from salient fields but does not constrain values to text explicitly present. Tightening this to reduce fabrication (and feeding the ontology's per-type attribute keys into the prompt) is tracked separately as khora#1543.

## Test plan
- [x] Unit tests pass (`tests/unit/test_extraction_llm.py`): both strict response formats carry the `attributes` pairs array in `properties` + `required` and stay strict-valid; parser folds pairs → dict, passes an already-dict through, collapses empty/absent/scalar → `{}`, coerces non-string/None pair values, drops malformed items, and last-wins on duplicate keys; async round-trip yields populated `.attributes`.
- [x] Adjacent extractor suites green (`test_extraction_llm_coverage.py`, `test_extraction_second_pass_batch.py`, `test_parameterized_extraction.py`).
- [x] `ruff format` / `ruff check` / typecheck clean.
- [ ] Full integration + e2e suite — validated in CI (requires Postgres/Neo4j).

## Summary by CodeRabbit

* **Enhancements**
  * Updated entity extraction to emit `attributes` as an array of `{key, value}` pairs for structured outputs.
  * Strengthened strict JSON schema support to require `attributes` in `{key, value}` pair form across single and multi-part extraction.
  * Added normalization so extracted `attributes` are consistently returned as a key/value dictionary, with safe handling of missing data.
* **Bug Fixes**
  * Improved robustness for empty, malformed, or non-standard `attributes` inputs (invalid entries are ignored; duplicate keys keep the last value).

### DIFF
--- a/src/khora/extraction/extractors/llm.py
+++ b/src/khora/extraction/extractors/llm.py
@@ -209,6 +209,7 @@ Extract entities, relationships, and temporal information from the following tex
 {document_context}
 Entity types to extract: {entity_types}
 Relationship types to use: {relationship_types}
+For each entity, emit "attributes" as an array of {{"key": ..., "value": ...}} string pairs drawn from its salient fields (identifiers, emails, state, urls, dates, etc.).
 
 Text:
 {text}"""
@@ -745,8 +746,27 @@ class LLMEntityExtractor(EntityExtractor):
                                                 {"type": "null"},
                                             ],
                                         },
+                                        "attributes": {
+                                            "type": "array",
+                                            "items": {
+                                                "type": "object",
+                                                "properties": {
+                                                    "key": {"type": "string"},
+                                                    "value": {"type": "string"},
+                                                },
+                                                "required": ["key", "value"],
+                                                "additionalProperties": False,
+                                            },
+                                        },
                                     },
-                                    "required": ["name", "entity_type", "description", "aliases", "temporal"],
+                                    "required": [
+                                        "name",
+                                        "entity_type",
+                                        "description",
+                                        "aliases",
+                                        "temporal",
+                                        "attributes",
+                                    ],
                                     "additionalProperties": False,
                                 },
                             },
@@ -843,8 +863,20 @@ class LLMEntityExtractor(EntityExtractor):
                                         {"type": "null"},
                                     ],
                                 },
+                                "attributes": {
+                                    "type": "array",
+                                    "items": {
+                                        "type": "object",
+                                        "properties": {
+                                            "key": {"type": "string"},
+                                            "value": {"type": "string"},
+                                        },
+                                        "required": ["key", "value"],
+                                        "additionalProperties": False,
+                                    },
+                                },
                             },
-                            "required": ["name", "entity_type", "description", "aliases", "temporal"],
+                            "required": ["name", "entity_type", "description", "aliases", "temporal", "attributes"],
                             "additionalProperties": False,
                         },
                     },
@@ -2770,9 +2802,23 @@ Return ONLY valid JSON, no other text."""
                             valid_until=t.get("valid_until"),
                         )
 
-                # Ensure attributes is a dict (LLM sometimes returns a list)
-                attrs = e.get("attributes", {})
-                if not isinstance(attrs, dict):
+                # Normalize attributes to a dict. The strict schema emits a list
+                # of {"key": ..., "value": ...} pairs; other paths may return a
+                # dict directly. Anything else collapses to an empty dict.
+                raw_attrs = e.get("attributes", {})
+                if isinstance(raw_attrs, dict):
+                    attrs = raw_attrs
+                elif isinstance(raw_attrs, list):
+                    attrs = {}
+                    for pair in raw_attrs:
+                        if not isinstance(pair, dict):
+                            continue
+                        key = pair.get("key")
+                        if not isinstance(key, str):
+                            continue
+                        value = pair.get("value")
+                        attrs[key] = str(value) if value is not None else ""
+                else:
                     attrs = {}
 
                 # QUALITY FIX: Use heuristic confidence instead of hardcoded 0.9
--- a/src/khora/extraction/extractors/llm.py
+++ b/src/khora/extraction/extractors/llm.py
@@ -190,7 +190,7 @@ Guidelines:
 - Use canonical entity names (e.g., "Jennifer Walsh" not "Jenny", "Acme Corporation" not "Acme Corp")
 - Include aliases for entities that have multiple names/abbreviations
 - Extract temporal information when dates, times, or relative time references appear
-- For STATE_CHANGE detection: when text indicates transitions ("switched from X to Y", "no longer X", "used to X", "previously X but now Y"), extract a STATE_CHANGE entity with these required attributes: {"entity_affected": "name of entity whose state changed", "previous_state": "old value", "new_state": "new value", "attribute_changed": "what changed (e.g. job_title, location, instrument)", "transition_date": "ISO date or null"}. Set valid_from to the transition date. Use INVOLVES to link it to the affected entity
+- For STATE_CHANGE detection: when text indicates transitions ("switched from X to Y", "no longer X", "used to X", "previously X but now Y"), extract a STATE_CHANGE entity whose attributes carry these keys as {"key", "value"} pairs: entity_affected (name of entity whose state changed), previous_state (old value), new_state (new value), attribute_changed (what changed, e.g. job_title, location, instrument), transition_date (ISO date or null). Set valid_from to the transition date. Use INVOLVES to link it to the affected entity
 - For EVENT detection: when text describes specific occurrences, extract the event with date, participants, and location when available
 - Use temporal relationships (PRECEDES, FOLLOWS, INVOLVES) to connect events and state changes to other entities
 - Ensure relationship source/target names match extracted entity names exactly
@@ -2361,7 +2361,8 @@ Return a JSON object with a "sections" array, one object per input section:
     {"entities": [...], "relationships": [...], "events": [...]},
     ...
 ]}
-Each section follows the entity/relationship format from the instructions above."""
+Each section follows the entity/relationship format from the instructions above.
+For each entity, emit "attributes" as an array of {"key": ..., "value": ...} string pairs drawn from its salient fields (identifiers, emails, state, urls, dates, etc.)."""
             )
 
             prompt_context = {
@@ -2403,6 +2404,7 @@ Return a JSON object with a "sections" array, one object per section:
 ]}}
 
 Each section follows the same entity/relationship/event format.
+For each entity, emit "attributes" as an array of {{"key": ..., "value": ...}} string pairs drawn from its salient fields (identifiers, emails, state, urls, dates, etc.).
 Return ONLY valid JSON, no other text."""
 
         try:
--- a/tests/unit/test_extraction_llm.py
+++ b/tests/unit/test_extraction_llm.py
@@ -1182,3 +1182,164 @@ class TestBisectionOnTruncation:
 
         assert results == []
         mock_multi.assert_not_called()
+
+
+# ---------------------------------------------------------------------------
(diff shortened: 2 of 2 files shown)


filippetrovic:
## Consolidated Review — multi-perspective team

**Overall assessment: APPROVE / merge.** No CRITICAL or HIGH findings across security, performance, correctness, tests, docs, AI, and devil's-advocate perspectives. One non-blocking MEDIUM and three LOWs, summarized below.

### MEDIUM
- **[AI] Attribute-value fabrication / grounding risk** — Adding `attributes` to the strict schema's `required` list means the model is asked to emit an attributes array for every entity. This invites fabricated attribute values, and attribute *quality* is not asserted by the new tests (they validate the emission channel and parsing, not correctness of values). This is inherent to enabling attribute emission and is intentionally out of scope for this keystone change — grounding attribute keys to the ontology's per-type schema is the tracked follow-up (khora#1543). Non-blocking; recommend the follow-up add grounding + quality guardrails. The empty-array escape (mirroring `aliases`) means the model is never forced to invent content, which mitigates the pressure.

### LOW
- **[Correctness]** `src/khora/extraction/extractors/llm.py` (`_parse_response`) — pair-form values are `str()`-coerced while the already-dict passthrough is not. Benign inconsistency, only reachable on the loose (non-strict) path, and consistent with prior dict behavior.
- **[Security]** `src/khora/extraction/extractors/llm.py` (`_parse_response`) — no explicit cap on attribute pair count/length. Bounded by LLM output and identical to the pre-existing dict path; informational hardening only.
- **[Docs]** `src/khora/extraction/extractors/llm.py` (structured prompt) — the added attributes line is clear and correctly brace-escaped for `.format()`; no change needed.

### Perspective notes
- **Security** — No blocking concerns. Untrusted LLM output is folded defensively (`isinstance` gating on pair items and string keys; `str()` coercion of inert JSON types; no eval/exec path). Attributes flow downstream as bound parameters, not string-interpolated into Cypher/SQL. No secrets or new dependencies.
- **Performance** — No concerns. The fold is O(total pairs) with no unbounded growth; the only cost is the intended output tokens for the required attributes array.
- **Correctness** — All edge cases handled (non-dict items skipped, non-string keys skipped, missing/None value → `""`, duplicate-key last-wins, dict passthrough, non-list/non-dict → `{}`). Both strict entity-item schemas are strict-valid and mutually consistent (`attributes` in `properties` + `required`, `additionalProperties: false`, `strict: true`); these are the only two entity-item strict schemas, so nothing is left inconsistent. No placeholders.
- **Tests** — Adequate: both response formats' strict-validity, pair→dict fold, dict passthrough, empty/absent/scalar → `{}`, and an async round-trip.
- **Devil's Advocate** — No blocking objection. Array-of-pairs is the correct, idiomatic representation for strict structured outputs (open maps are illegal under `strict`); `attributes` in `required` is API-mandated with an empty-array escape; scope is correctly minimal with per-type key hints deferred.

## Security Audit
**Verdict: PASS.** Diff scanned against the 16 categories (injection, authn, crypto/secrets, access control, misconfig, XSS, SSRF, CSRF, deserialization, file/path, open redirect, mass assignment, race conditions, ReDoS, supply chain, boundary validation). No CRITICAL/HIGH/MEDIUM findings. The only trust boundary is untrusted LLM output folded into a dict, handled defensively with no dangerous sink. No dependency/manifest changes.

_Reviewers: security, performance, correctness, test, docs, AI, devil's-advocate + dedicated security audit._

probablyrobot (approved):

filippetrovic:
## Independent second-pass review — batch-path gap

Verified the keystone plumbing end-to-end: both strict schemas carry `attributes` in `properties` **and** `required`; the pairs→dict fold runs on the batch path too (via the shared `_parse_response`, so folded dicts — not raw pair lists — are stored); brace-escaping is safe for `.format()`; `attributes` is required-but-*emptyable* (no forced fabrication). One must-fix before this delivers value in production, plus two nits.

### M1 — MUST FIX: the batch (production) prompt never asks for attributes
Production ingest runs `extract_multi` → `_extract_multi_batch`, which builds its **own** prompt. The single-doc `EXTRACTION_PROMPT_STRUCTURED` — the only prompt this PR touched — is never used on that path. So in batch mode the model is told nothing about attributes, and because the schema permits `attributes: []`, it will emit empty arrays the vast majority of the time → entities keep landing with `{}`, the exact outcome #1542 is meant to fix.

Add the same one-line instruction — *emit `attributes` as `{key, value}` pairs from the entity's salient fields (identifiers, emails, state, urls, dates)* — to **both** batch prompt builders in `_extract_multi_batch` (brace-escaped as `{{ }}`, since these are `.format()`/f-strings):
- the fallback hardcoded prompt (~line 2373, near *"Each section follows the same entity/relationship/event format."*), and
- the custom-`extraction_prompt` multi-section block (~line 2332).

Don't route this through `DEFAULT_SYSTEM_PROMPT` — an expertise that supplies its own `system_prompt` bypasses it, so the batch **user** prompt is the correct place.

### T1 — add a batch-path round-trip test
Current tests drive `_parse_response` directly and single-doc `extract()` only. Add one `extract_multi` round-trip test (mock a `{"sections": [{"entities": [{…, "attributes": [{"key": …, "value": …}]}]}]}` response) asserting `result.entities[0].attributes` is the folded dict. The batch path folds via the shared `_parse_response` today, so this is a regression guard — without it, a future refactor giving the batch path its own parse would silently drop attributes in production with no failing test.

### N1 — nit (optional): stale STATE_CHANGE example
`DEFAULT_SYSTEM_PROMPT` (~line 193) still shows STATE_CHANGE attributes as a dict literal, now inconsistent with the pairs schema. Harmless (the strict schema wins), optional cleanup for consistency.

---

Scope note: M1 is partly an omission in the originating issue — #1542 named only `EXTRACTION_PROMPT_STRUCTURED`. #1542 has been updated to include the batch prompt in scope; per-type attribute **keys** remain #1543's job, layered on top of this nudge on both paths.

filippetrovic:
## Addressed the second-pass review (commit a0905059)

**M1 (must-fix) — done.** The batch prompt now asks for attributes on **both** builders in `_extract_multi_batch`:
- Custom-`extraction_prompt` multi-section block — added to the `## MULTI-SECTION RESPONSE FORMAT` text with **single** braces (it's substituted as a Jinja `text` value, not `.format()`-processed, so it renders literally — matching the existing `{"sections": [...]}` in that same block).
- Hardcoded fallback prompt — added with `{{ }}` escaping since it's an f-string.

Not routed through `DEFAULT_SYSTEM_PROMPT` (an expertise `system_prompt` would bypass it), per your note. Verified the rendered fallback prompt contains the literal line `...an array of {"key": ..., "value": ...} string pairs...` with no stray `{{`.

**T1 — done.** Added `test_extract_multi_round_trip_folds_pair_attributes`: mocks a `{"sections":[{"entities":[{…,"attributes":[{key,value}]}]}]}` response and asserts `results[0].entities[0].attributes` is the folded dict — a regression guard on the batch parse path.

**N1 — done (cheap).** Reworded the `STATE_CHANGE` example in `DEFAULT_SYSTEM_PROMPT` from a dict literal to `{"key", "value"}` pairs, preserving every key (`entity_affected`, `previous_state`, `new_state`, `attribute_changed`, `transition_date`) and its meaning — now consistent with the pairs schema.

Schema and `_parse_response` fold left as-is per your note. 168 extractor tests pass (`test_extraction_llm`, `_second_pass_batch`, `_parameterized_extraction`, `_llm_coverage`); ruff format/check clean. CI re-running.

PR #1551, opened 2026-07-21
Guard entity upsert against empty attributes overwriting populated ones

### LINKED ISSUE #1545: Storage: stop empty attributes overwriting populated ones on entity upsert (Neo4j + pgvector)
Part of #1541.

## Summary

On entity upsert, an incoming **empty** `attributes` overwrites a previously-stored populated one. Because the same entity recurs across batches within a single ingest, a later batch that extracted no attributes for it wipes attributes an earlier batch captured — attribute loss even in a clean namespace. See #1541.

## Where

- Neo4j `_UPSERT_CYPHER` (`src/khora/storage/backends/neo4j.py`, ~line 1944): `ON MATCH SET ... e.attributes = row.attributes` (unconditional overwrite). Also the single-update path (~1780–1794).
- pgvector (`src/khora/storage/backends/pgvector.py`, ~line 1090): `on_conflict_do_update(... set_={"attributes": stmt.excluded.attributes})` (unconditional overwrite).

`attributes` is stored as a JSON **string** in Neo4j and as **JSONB** in Postgres.

## Fix — empty-guard (keep existing when incoming is empty)

Scope is a guard, **not** a deep key-union merge (see non-goal). Overwrite only when the incoming value is non-empty:

- **Neo4j** `_UPSERT_CYPHER` `ON MATCH`:
  ```cypher
  e.attributes = CASE
    WHEN row.attributes IS NULL OR row.attributes = '{}' OR row.attributes = ''
    THEN e.attributes ELSE row.attributes END
  ```
  Apply the same guard to the single-update path.
- **pgvector** `on_conflict_do_update`, `attributes` value becomes a SQLAlchemy `case`:
  ```python
  attributes=sa.case(
      (sa.cast(stmt.excluded.attributes, JSONB) == sa.cast("{}", JSONB), <table>.c.attributes),
      else_=stmt.excluded.attributes,
  )
  ```
  (handle `NULL` incoming as "keep existing" as well).

**Non-goal (explicit):** deep key-union merge of old+new attributes. That needs JSON parsing on the graph side (APOC / server-side JSON), is a larger change, and is not required to stop the observed loss. File separately if wanted.

## Acceptance criteria

- [ ] Upsert `{a: 1}` then upsert same key with `{}` (or `NULL`) → stored attributes remain `{a: 1}` on **both** Neo4j and pgvector.
- [ ] Upsert same key with a non-empty `{b: 2}` → stored attributes become `{b: 2}` (plain overwrite; key-union is out of scope — assert `{b: 2}`, not `{a: 1, b: 2}`, and note this intent in the test).
- [ ] Integration tests on both the neo4j and pgvector lanes; `make format && make test` green (coverage ≥65%).

filippetrovic:
## Summary
- On entity upsert, an incoming **empty** (`{}`) or `NULL` `attributes` no longer overwrites a stored, populated `attributes`. The same entity recurs across batches within a single ingest, so a later empty-attribute batch previously wiped earlier-captured attributes — this stops that data loss, even in a clean namespace.
- **Neo4j** (`attributes` stored as a JSON string): the `ON MATCH SET` clause of the batch upsert (`_UPSERT_CYPHER`) and the single-entity `update_entity` path now keep the existing value when the incoming string is `NULL`, `'{}'`, or empty, via `CASE WHEN ... THEN e.attributes ELSE row.attributes END`.
- **pgvector** (`attributes` stored as JSONB): the `ON CONFLICT DO UPDATE` set-clause keeps the existing column when the incoming value is SQL `NULL` or `{}`::jsonb, via a `CASE`. The `NULL` branch uses an explicit `IS NULL` check (since `NULL = '{}'::jsonb` is `NULL`, not false).
- A **non-empty** incoming value still overwrites the stored dict wholesale. Deep key-union merge of old + new keys is intentionally **out of scope** (filed separately if wanted).

## Test plan
- [x] Unit tests: no regressions introduced by this change
- [ ] Integration tests (both Neo4j and pgvector lanes) — added under `tests/integration/`, exercised in CI (require live Postgres/Neo4j)
- [x] Manual verification: confirmed the guard matches the serializer — an empty dict serializes to `'{}'` and `None` to a `NULL`, so every empty representation is fenced; a non-empty dict overwrites and is asserted to replace (not union) the stored value

Fixes #1545

## Summary by CodeRabbit

- **Bug Fixes**
  - Updated entity upsert behavior across Neo4j and PostgreSQL/pgvector to preserve existing `attributes` when the incoming payload is `null` or empty; non-empty payloads still fully overwrite.
  - Prevented creation of unintended `EntityVersion` snapshots and `SUPERSEDES` edges during no-op (empty/NULL) re-upserts.
- **Tests**
  - Expanded Neo4j integration tests to verify overwrite vs no-overwrite behavior and assert correct version-chain/edge creation.
  - Added pgvector integration tests covering the same empty/NULL/non-empty semantics for both single and batch upserts.

### DIFF
--- a/src/khora/storage/backends/neo4j.py
+++ b/src/khora/storage/backends/neo4j.py
@@ -1781,7 +1781,7 @@ class Neo4jBackend(GraphBackendBase):
             MATCH (e:Entity {id: $id, namespace_id: $namespace_id})
             SET e.name = $name,
                 e.description = $description,
-                e.attributes = $attributes,
+                e.attributes = CASE WHEN $attributes IS NULL OR $attributes = '{}' OR $attributes = '' THEN e.attributes ELSE $attributes END,
                 e.source_document_ids = $source_document_ids,
                 e.source_chunk_ids = $source_chunk_ids,
                 e.mention_count = $mention_count,
@@ -1941,7 +1941,7 @@ class Neo4jBackend(GraphBackendBase):
                 e.confidence = CASE WHEN row.confidence > e.confidence THEN row.confidence ELSE e.confidence END,
                 e.updated_at = row.updated_at,
                 e.version_valid_from = coalesce(e.version_valid_from, row.version_valid_from),
-                e.attributes = row.attributes
+                e.attributes = CASE WHEN row.attributes IS NULL OR row.attributes = '{}' OR row.attributes = '' THEN e.attributes ELSE row.attributes END
             RETURN e.id AS id, e.name AS name, row.id AS input_id,
                    CASE WHEN e.id = row.id THEN true ELSE false END AS is_new
         """
--- a/src/khora/storage/backends/neo4j.py
+++ b/src/khora/storage/backends/neo4j.py
@@ -2145,8 +2145,13 @@ class Neo4jBackend(GraphBackendBase):
                     if not pre:
                         continue
 
-                    # Compare serialized attributes to detect real changes
-                    if pre["attributes"] == input_row["attributes"]:
+                    # Compare serialized attributes to detect real changes.
+                    # An empty/NULL incoming attributes is treated as no change,
+                    # mirroring the ON MATCH SET guard above so an empty re-upsert
+                    # of a populated entity does not append a phantom version_row.
+                    incoming_attrs = input_row["attributes"]
+                    incoming_is_empty = incoming_attrs is None or incoming_attrs == "{}" or incoming_attrs == ""
+                    if incoming_is_empty or pre["attributes"] == incoming_attrs:
                         continue
 
                     version_rows.append(
--- a/src/khora/storage/backends/pgvector.py
+++ b/src/khora/storage/backends/pgvector.py
@@ -1087,13 +1087,17 @@ class PgVectorBackend(AsyncSessionMixin):
                 constraint="uq_entities_namespace_name_type",
                 set_={
                     "description": stmt.excluded.description,
-                    # An incoming empty ({}) or NULL attributes must not clobber a
-                    # stored populated attributes; keep the existing value in that
-                    # case, otherwise overwrite with the incoming value.
+                    # An incoming empty ({}), SQL NULL, or json ``null`` attributes
+                    # must not clobber a stored populated attributes; keep the
+                    # existing value in that case, otherwise overwrite with the
+                    # incoming value. JSONB has ``none_as_null=False``, so a Python
+                    # None binds as ``'null'::jsonb`` (not SQL NULL) — the
+                    # jsonb_typeof check catches that path.
                     "attributes": sa.case(
                         (
                             sa.or_(
                                 stmt.excluded.attributes.is_(None),
+                                sa.func.jsonb_typeof(stmt.excluded.attributes) == "null",
                                 stmt.excluded.attributes == sa.cast("{}", JSONB),
                             ),
                             EntityModel.attributes,
--- a/src/khora/storage/backends/pgvector.py
+++ b/src/khora/storage/backends/pgvector.py
@@ -1054,7 +1054,7 @@ class PgVectorBackend(AsyncSessionMixin):
         Acquires a namespace-scoped advisory lock to prevent deadlocks
         when concurrent coroutines upsert entities in the same namespace.
         """
-        from sqlalchemy.dialects.postgresql import insert
+        from sqlalchemy.dialects.postgresql import JSONB, insert
 
         async with self._get_session() as session:
             # Advisory lock prevents deadlocks with concurrent upserts
@@ -1087,7 +1087,19 @@ class PgVectorBackend(AsyncSessionMixin):
                 constraint="uq_entities_namespace_name_type",
                 set_={
                     "description": stmt.excluded.description,
-                    "attributes": stmt.excluded.attributes,
+                    # An incoming empty ({}) or NULL attributes must not clobber a
+                    # stored populated attributes; keep the existing value in that
+                    # case, otherwise overwrite with the incoming value.
+                    "attributes": sa.case(
+                        (
+                            sa.or_(
+                                stmt.excluded.attributes.is_(None),
+                                stmt.excluded.attributes == sa.cast("{}", JSONB),
+                            ),
+                            EntityModel.attributes,
+                        ),
+                        else_=stmt.excluded.attributes,
+                    ),
                     "source_document_ids": _accumulate_source_ids_sql("source_document_ids"),
                     "source_chunk_ids": _accumulate_source_ids_sql("source_chunk_ids"),
                     "mention_count": stmt.excluded.mention_count,
--- a/src/khora/storage/backends/pgvector.py
+++ b/src/khora/storage/backends/pgvector.py
@@ -1092,13 +1092,16 @@ class PgVectorBackend(AsyncSessionMixin):
                     # existing value in that case, otherwise overwrite with the
                     # incoming value. JSONB has ``none_as_null=False``, so a Python
                     # None binds as ``'null'::jsonb`` (not SQL NULL) — the
-                    # jsonb_typeof check catches that path.
+                    # jsonb_typeof check catches that path. The empty-object check
+                    # uses ``type_coerce({}, JSONB)`` (binds the dict -> ``'{}'``)
+                    # rather than ``cast("{}", JSONB)``, which would JSON-encode the
+                    # string to ``'"{}"'`` and never match a real empty object.
                     "attributes": sa.case(
                         (
                             sa.or_(
                                 stmt.excluded.attributes.is_(None),
                                 sa.func.jsonb_typeof(stmt.excluded.attributes) == "null",
-                                stmt.excluded.attributes == sa.cast("{}", JSONB),
+                                stmt.excluded.attributes == sa.type_coerce({}, JSONB),
                             ),
                             EntityModel.attributes,
                         ),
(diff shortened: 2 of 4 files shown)


probablyrobot (approved):

filippetrovic:

Reviewers: Security, Performance, Correctness, Test, Database, Docs, Devil's Advocate (parallel, independent).

**Overall: REQUEST CHANGES.** The pgvector and Neo4j current-state guards are sound, but two defects need fixing before merge — one is CI-blocking.

### HIGH
- **H1 [Correctness / Database]** `src/khora/storage/backends/pgvector.py:1096` — the NULL branch is inert. `attributes` is `JSONB` with SQLAlchemy's default `none_as_null=False`, so a Python `None` binds as `'null'::jsonb`, **not** SQL `NULL`. Both `excluded.attributes.is_(None)` and `excluded.attributes == '{}'::jsonb` are therefore false, the `CASE` falls to `else_`, and a populated row is clobbered with json `null`. The `None`-incoming integration test will **fail against real Postgres** (CI-blocking). Fix: also match jsonb `'null'`, e.g. add `func.jsonb_typeof(excluded.attributes) == 'null'` (or `== cast('null', JSONB)`) to the OR.
- **H2 [Correctness / Database / Performance / Devil's Advocate]** `src/khora/storage/backends/neo4j.py:2149` — the Phase-2 bi-temporal change-detector still compares the stored value against the **raw** incoming attributes (`pre["attributes"] == input_row["attributes"]`). On an empty/NULL re-upsert of a populated entity the `ON MATCH SET` guard correctly keeps the stored value, but this comparison sees a "change" and creates a phantom `:EntityVersion` + `SUPERSEDES` snapshot and bumps `version_valid_from`. This pollutes point-in-time recall (an active read path) and grows the version chain unboundedly on the common re-mention-without-attributes ingest pattern. Fix: mirror the empty/NULL guard at `:2149` (treat incoming in `{None, "{}", ""}` as no-change).

### MEDIUM
- **M1 [Devil's Advocate]** The same clobber pattern exists on other backends (SurrealDB, Memgraph, AGE/Neptune) which are **not** guarded here. This is **intentionally out of scope** for this change (scoped to Neo4j + pgvector); recommend a follow-up issue to extend the guard rather than leaving a silently backend-dependent guarantee.
- **M2 [Devil's Advocate]** Guarding the single-entity `update_entity` path removes the only way to intentionally clear attributes to `{}` on pgvector (it shares `_upsert_entity` with `create_entity`). Verified no current caller relies on clearing — flag, not block.

### LOW
- **L1 [Devil's Advocate / Correctness / Database]** The `... = ''` branch in the Neo4j `CASE` (`neo4j.py:1784`, `:1944`) is unreachable — `serialize_dict` only yields `None` or `json.dumps(dict)`, never `''`. Harmless belt-and-suspenders (matches the agreed guard spec); drop or comment as intentional.
- **L2 [Devil's Advocate / Test]** `tests/integration/test_neo4j_entity_attributes_guard.py` docstring claims CI does not provision Neo4j — this is inaccurate; CI provisions Neo4j and sets the opt-in env var, so both new tests **do** run in CI. Correct the docstring, and add a version-chain assertion so H2 is covered by a test.
- **L3 [Security]** The new integration tests hardcode localhost `make dev` default credentials — env-overridable, gated behind reachability checks, byte-identical to the existing sibling test convention. Not a real secret; no action.

### No-finding confirmations
- **Security:** injection CLEAR (static SQL/Cypher, all values bound as parameters; no user-data string interpolation); namespace scoping / IDOR guards untouched; no data-exposure; no internal-identifier leak in the public diff.
- **Performance:** guard per-row cost negligible on both backends (`'{}'::jsonb` cast is constant-folded at plan time; Neo4j does ≤3 inline string-equality checks; no index defeat, no added round-trips). Note the H2 fix also removes now-wasted graph writes.
- **Correctness:** no placeholders (TODO/FIXME/NotImplementedError/stub) in the diff; Neo4j guard `CASE` is exhaustive vs. the serializer.
- **Docs:** no docs drift; no CHANGELOG entry required (PR titles are the changelog of record; no version/Rust/pyproject files touched).

### Security Audit
Scope-appropriate: this change is a JSONB/Cypher write-path guard with no authentication, authorization, input-to-query interpolation, cryptographic, or new-dependency surface. The Security Reviewer's OWASP-relevant pass (injection, authz, secrets, data exposure) returned CLEAR; no additional findings. Only the informational L3 (localhost test credentials) applies.

**Action:** H1 and H2 will be fixed on this branch before merge.

filippetrovic:
### Review findings addressed — CI green

Both blocking findings from the consolidated review are fixed on this branch, validated by the live Postgres + Neo4j integration lane:

- **H1** — pgvector now keeps existing attributes for an incoming SQL `NULL`, json `null`, and empty `{}`. Two subtleties were corrected: a Python `None` binds as `'null'::jsonb` (not SQL `NULL`) so it's matched via `jsonb_typeof(...) = 'null'`; and the empty-object check now uses `type_coerce({}, JSONB)` — `cast("{}", JSONB)` was JSON-encoding the string to `'"{}"'::jsonb` and never matching a real empty object.
- **H2** — the Neo4j Phase-2 change-detector now mirrors the `ON MATCH SET` guard, so an empty/NULL re-upsert of a populated entity no longer creates a phantom `:EntityVersion` snapshot. Added regression tests: zero snapshots on empty/NULL re-upsert, exactly one on a genuine change.
- Test docstring corrected (CI does provision Neo4j), and the pgvector test comments now describe the `'null'::jsonb` binding accurately.

All required checks pass.

filippetrovic:
### Follow-up: pgvector batch-ingest path also guarded

A further pass found the earlier pgvector fix only covered the single-entity `_upsert_entity` path; the **batch** path `upsert_entities_batch` (the primary bulk-ingest route used by the coordinator / ingest pipeline / vectorcypher engine) still did a plain `attributes` overwrite on conflict. That is precisely the "empty re-upsert across batches wipes earlier attributes" scenario in the problem statement, and no test covered it.

Fixed by applying the identical keep-existing guard to the batch `on_conflict_do_update`, plus 3 batch-path integration tests (empty / NULL / plain-overwrite). Both pgvector conflict-update sites for `attributes` are now guarded; the Neo4j single + batch + Phase-2 paths were already handled.

CI green — the live Postgres lane runs all 6 pgvector tests (3 single + 3 batch) and 5 Neo4j tests.

filippetrovic:
## Independent review — entity-attribute upsert guard

Verified against head `7f5bb6e3`. Strong PR — all four iterated fixes are present and correct, and the main risk I checked (a pgvector phantom-version bug symmetric to the Neo4j H2) is **not present**: pgvector entities have no version machinery at all (`EntityModel` carries only `valid_from`/`valid_until`; `version_valid_from`/`:EntityVersion` are Neo4j-only). No third entity-attribute overwrite site is missed; the Neo4j guard's empty-definition matches the serializer (`serialize_dict({}) == '{}'`, `None → NULL`); both pgvector conflict sites and both Neo4j paths are guarded; the pgvector `type_coerce({}, JSONB)` / `jsonb_typeof = 'null'` fix is correct.

One requested change before merge, plus two explicitly out-of-scope follow-ups.

### M1 — add a Neo4j `update_entity` single-path test (MEDIUM, test gap)
The guard on the single `update_entity` Cypher (`neo4j.py:1784`) is correct but **untested** — every Neo4j test drives `upsert_entities_batch` only. `update_entity` is a live production path (`pipelines/flows/expansion.py`, `storage/coordinator.py`). CodeRabbit's request to cover the single path (its comment on `tests/integration/test_neo4j_entity_attributes_guard.py:149`) was addressed for pgvector — whose `create_entity`/`update_entity` both route through the guarded `_upsert_entity` — but not for Neo4j, where `create_entity` is insert-only `CREATE` and so never exercises the update guard.

Add one test: seed a populated `{"a": 1}` via batch, then `await backend.update_entity(<entity with attributes={}>, namespace_id=ns)` and assert `attributes == {"a": 1}`; add a `None` variant. Guards against a future edit diverging line 1784 from the batch guard at line 1944.

### F1 / F2 — out of scope for this PR (track separately; do NOT add here)
#1545 scopes strictly to entity `attributes`, so these are explicitly not in scope — noting them only so they can become a follow-up ticket:
- **F1** — pgvector still overwrites `valid_from`/`valid_until` from `excluded` unconditionally on an empty re-upsert, and relationship `properties` (pgvector + the Neo4j relationship MERGE) are not empty-guarded. Same class of "empty payload clobbers stored value," different columns.
- **F2** (pre-existing, not introduced here) — the Neo4j Phase-2 change-detector compares JSON *strings* (`pre["attributes"] == incoming`), which is key-order sensitive; a re-extraction that reorders keys could snapshot a spurious `:EntityVersion`. Predates this PR and is unaffected by the guard.

filippetrovic:
### M1 addressed — single `update_entity` path now covered

Added three integration tests to `tests/integration/test_neo4j_entity_attributes_guard.py` that drive the single-entity `update_entity` path directly (the existing Neo4j tests only exercised `upsert_entities_batch`): seed a populated `{"a": 1}` via the batch path, then `update_entity(...)` the same node and assert —
- incoming `{}` → stays `{"a": 1}`
- incoming `None` → stays `{"a": 1}` (the `$attributes IS NULL` branch)
- incoming `{"b": 2}` → replaces to exactly `{"b": 2}` (no key-union)

Guard code is unchanged (it was already verified correct on all paths). The live Postgres + Neo4j lane runs the full file (now 8 Neo4j tests).

**F1 / F2 are intentionally left out of scope here** — they concern `valid_from`/`valid_until`, relationship `properties`, and a pre-existing key-order-sensitive comparison, unrelated to #1545. Tracking those separately.

PR #1552, opened 2026-07-21
Add ungated per-type attribute-schema hints to extraction prompts

### LINKED ISSUE #1543: Extraction: feed the ontology's per-type attribute schema into the extraction prompt (ungated)

filippetrovic:
## Summary
- Add `_build_attribute_schema_block`, which lists each extracted entity type's declared `required`/`optional` attribute keys so the model knows which attribute keys to populate on extracted entities.
- The block is built from the **resolved** entity types and is **ungated** — no longer tied to tool schemas or source-tool context — so it reaches both the single-document (`_render_extraction_prompt`) and batch (`_extract_multi_batch`) prompt paths, including when no source-tool context is present.
- Default/structured and hardcoded-fallback templates inject the block directly (adjacent to the existing general "emit attributes" nudge, without duplicating it); custom `extraction_prompt` templates receive it as a template variable on both single and batch paths, mirroring how tool-context is delivered.
- Remove the now-redundant per-type attribute listing from the tool-context builder to avoid double-listing; its source/tool-field output is unchanged.

## Test plan
- [x] Unit tests pass (extraction suite: 287 passed)
- [ ] Integration tests pass (run in CI)
- [x] Block present on single + batch prompts incl. no-source-tool-context case; absent when no expertise / no declaring type; tool-context builder no longer double-lists (its field output intact)

Fixes #1543

## Summary by CodeRabbit

* **New Features**
  * Added an ontology-driven **ATTRIBUTE SCHEMA** prompt block for LLM entity extraction, listing required/optional attribute keys per entity type.
  * Included the attribute schema in single-document and batch/multi-section extraction prompts, including custom templates via `{{ attribute_schema }}`.

* **Bug Fixes**
  * Removed inline “expected entity attributes” hinting from the tool context so extraction guidance uses the standalone schema block.
  * Omitted schema details when expertise or attribute declarations are missing.

* **Documentation**
  * Updated extraction docs to explain `{{ tool_context }}` and `{{ attribute_schema }}` usage in custom prompts.

* **Tests**
  * Expanded unit coverage for attribute-schema rendering and prompt injection.

### DIFF
--- a/docs/extraction/expertise-system.md
+++ b/docs/extraction/expertise-system.md
@@ -373,6 +373,8 @@ Prompts support Jinja2 templating:
 
 Prompts render in a Jinja `ImmutableSandboxedEnvironment`; unsafe constructs (dunder/private attribute access, mutating methods) are rejected and raise `SecurityError`.
 
+Besides any keys you pass via `context`, templates can reference `{{ tool_context }}` and `{{ attribute_schema }}`. `{{ attribute_schema }}` renders the per-type attribute-key hints — each entity type's `required`/`optional` keys. The built-in prompts inject it automatically; a custom `extraction_prompt` must interpolate `{{ attribute_schema }}` explicitly to include those per-type keys.
+
 ```yaml
 system_prompt: |
   You are an expert at extracting information about {{ domain }} companies.
--- a/docs/extraction/extractors.md
+++ b/docs/extraction/extractors.md
@@ -198,6 +198,8 @@ expertise = ExpertiseConfig(
 
 Custom prompts render in a Jinja `ImmutableSandboxedEnvironment`; unsafe constructs (dunder/private attribute access, mutating methods) are rejected and raise `SecurityError`.
 
+Available variables include any keys you pass via `context` plus `{{ tool_context }}` and `{{ attribute_schema }}`. Interpolate `{{ attribute_schema }}` to include the per-type attribute-key hints (each entity type's `required`/`optional` keys); the built-in prompts add it automatically, but a custom template must reference it explicitly.
+
 ```python
 expertise = ExpertiseConfig(
     name="custom",
--- a/src/khora/extraction/extractors/llm.py
+++ b/src/khora/extraction/extractors/llm.py
@@ -1808,21 +1808,48 @@ class LLMEntityExtractor(EntityExtractor):
                 if key != "fields" and isinstance(values, list):
                     lines.append(f"  {key}: {', '.join(str(v) for v in values)}")
 
-        # Add attribute schema hints from entity types
-        if expertise.entity_types:
-            lines.append("\nEXPECTED ENTITY ATTRIBUTES:")
-            for et in expertise.entity_types:
-                required = et.attributes.get("required", [])
-                optional = et.attributes.get("optional", [])
-                if required or optional:
-                    parts = []
-                    if required:
-                        parts.append(f"required: {', '.join(required)}")
-                    if optional:
-                        parts.append(f"optional: {', '.join(optional)}")
-                    lines.append(f"  {et.name}: {'; '.join(parts)}")
-
         return "\n".join(lines)
+
+    def _build_attribute_schema_block(
+        self,
+        expertise: ExpertiseConfig | None,
+        entity_types: list[str] | None,
+    ) -> str:
+        """Build the per-type ATTRIBUTE SCHEMA hint block from the ontology.
+
+        For each extracted entity type that declares attributes in the
+        ExpertiseConfig, list its required/optional attribute key names so the
+        model knows which keys to populate. The keys are hints only — the model
+        prefers them but is not constrained to them, and emission still rides the
+        general ``attributes`` {"key", "value"}-pair channel.
+
+        Args:
+            expertise: Optional ExpertiseConfig carrying per-type attribute schemas.
+            entity_types: Resolved entity-type names being extracted.
+
+        Returns:
+            The schema block, or "" when there is no expertise or when no
+            extracted type declares any attributes.
+        """
+        if expertise is None or not entity_types:
+            return ""
+
+        by_name = {et.name: et for et in expertise.entity_types}
+        lines: list[str] = []
+        for type_name in entity_types:
+            et = by_name.get(type_name)
+            if et is None:
+                continue
+            required = et.attributes.get("required", [])
+            optional = et.attributes.get("optional", [])
+            if not required and not optional:
+                continue
+            lines.append(f"{et.name}: required=[{', '.join(required)}]; optional=[{', '.join(optional)}]")
+
+        if not lines:
+            return ""
+
+        return "ATTRIBUTE SCHEMA (emit these keys in attributes when present):\n" + "\n".join(lines)
 
     @staticmethod
     def _build_document_context(context: dict[str, Any] | None) -> str:
@@ -1906,6 +1933,7 @@ class LLMEntityExtractor(EntityExtractor):
                     "entity_types": entity_types,
                     "relationship_types": relationship_types,
                     "tool_context": tool_context,
+                    "attribute_schema": self._build_attribute_schema_block(expertise, entity_types),
                 }
                 return composer.render_prompt(
                     expertise.extraction_prompt,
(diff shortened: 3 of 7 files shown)


filippetrovic:

Reviewed by an independent 7-role team (correctness, AI/prompt, tests, security, performance, docs, devil's advocate) plus a dedicated security audit. **Overall: APPROVE** — one real correctness bug was found and has been fixed in this branch; everything else is non-blocking polish or follow-up.

No CRITICAL or HIGH findings.

### MEDIUM
- **M1 · Correctness · ✅ Fixed (`3e7795de`)** — `_build_attribute_schema_block` raised `TypeError` when an entity type's `required`/`optional` attribute side was present-but-`None` (the natural result of an empty YAML scalar, e.g. `optional:` with no value). The prior gated code guarded each side before joining; ungating widened the blast radius to every expertise-driven extraction, and CI wouldn't catch it (built-in ontologies are well-formed). Fixed by coercing `None`/missing → `[]`, with a regression test driven through the real `ExpertiseConfig.from_dict` YAML path.
- **M2 · AI/Prompt · By design (surfaced for awareness)** — the per-type line renders `TYPE: required=[…]; optional=[…]`. The word "required" is an extraction hint only; it has no downstream enforcement (attribute validation is additive-safe and drops absent keys), so in principle it could nudge a model to over-emit. The `required`/`optional` wording and header are the intended, fixed format for this change; the header already hedges "emit these keys **when present**". Flagged for future consideration, not changed here.

### LOW
- **L1 · Custom-prompt delivery** — for a user-authored `extraction_prompt`, the block only appears if the template references `{{ attribute_schema }}` (default and fallback templates inject it unconditionally). This mirrors the existing `{{ tool_context }}` contract, so it's consistent — but authors migrating a default prompt to a custom one could silently lose the hint. Recommend documenting available template variables.
- **L2 · Docs** — `ExpertiseComposer.render_prompt`'s "supported template variables" docstring lists neither `attribute_schema` nor the pre-existing `tool_context`. Small doc follow-up.
- **L3 · Placement** — the single-doc default path prepends the block at the top of the prompt; the batch fallback appends it right after the general "emit attributes" nudge (better co-located). Both are valid and non-duplicating; a comment slightly overstates adjacency. Cosmetic.
- **L4 · Duplicate/unknown types** — duplicate names in the resolved `entity_types` list would emit a duplicate line (last-wins on the config); an unknown name is skipped. Matches prior behavior; cosmetic.
- **L5 · Malformed non-list value** — a non-list string value (`required: email` instead of `[email]`) char-splits into `[e, m, a, i, l]`. Does **not** crash; left as-is rather than adding handling for a clearly-malformed config.
- **L6 · Test polish** — the unknown-type-skip branch, duplicate-type behavior, and a "custom template omits the variable → block absent" negative case are unpinned. Core behavior is well covered (18 tests inspecting real assembled prompts / captured LLM messages).

### Security Audit
**PASS** — no CRITICAL/HIGH/MEDIUM. The change is a pure string builder over operator-supplied config: no injection sinks (`.format()` is never applied to attribute-derived content; the block is concatenated/f-string-interpolated as inert text), no deserialization, no secrets, no network/file/redirect surface. Custom templates still render through the sandboxed composer with `SecurityError` fail-closed. Attribute **key names** reaching the prompt are operator config — the same trust level as the entity-type names and the full custom prompt template that already flow in — so no new untrusted→prompt path.

### Performance
No concerns — `O(types)`, computed once per LLM call (before retry loops), negligible added prompt tokens (one compact line per declaring type).

### Assessment
**APPROVE.** The single correctness bug (M1) is fixed with a regression test; the remaining items are documentation/polish that can land as follow-ups without blocking.

filippetrovic:
## Independent review — per-type attribute-schema hints

Verified against the API patch at head `3e7795de` (local clone is on `main`, so all checks used the API head). **APPROVE-WITH-NITS — nothing blocking; good to merge.** CI 26/26, codecov full, CodeRabbit clean.

**Spec conformance — all confirmed:**
- `_build_attribute_schema_block` is **ungated** (built from the resolved `entity_types` + `expertise`; never reads `context`/`source_tool`/`tool_schemas`); returns `""` when no expertise / no declaring type; correct `TYPE: required=[…]; optional=[…]` format.
- **Both batch builders get the block** (the A1 lesson): the hardcoded fallback prompt — the production/benchmark-relevant path, since the ontology has no custom `extraction_prompt` — appends it via direct f-string injection right after A1's general nudge, and `test_block_in_batch_prompt_ungated` drives the real `_extract_multi_batch` with `context=None` to prove it. The custom multi-section path delivers it via the `{{ attribute_schema }}` template var. Single path covered on both templates. Placed adjacent to A1's nudge, no duplication (distinct content: nudge = pairs format, block = per-type keys).
- `_build_tool_context` per-type block removed cleanly; its source/field output intact.
- **M1 fix confirmed** (`required`/`optional` present-but-`None` → `[]`), with a regression test that genuinely routes an empty-YAML-scalar shape through `ExpertiseConfig.from_dict` (not a hand-built dict).
- The two rippled tests (`test_saas_extraction.py`, `test_extraction_llm_coverage.py`) are **legitimate** — the attribute assertions were inverted but paired with strengthened positive checks (surviving SOURCE-CONTEXT field output), not gutted to pass.

**Nits — all non-blocking:**

### N1 — optional: document the `{{ attribute_schema }}` placeholder for custom templates
A custom `extraction_prompt` template receives the per-type block only if it interpolates `{{ attribute_schema }}`. Asymmetry worth noting: A1's general nudge is baked into `{{ text }}` (always present), but this block rides a separate `{{ attribute_schema }}` var — so a custom template that references `{{ text }}` but not `{{ attribute_schema }}` keeps the nudge and silently loses the per-type keys. **Not a production risk here** (the benchmark ontology has no custom `extraction_prompt`; ingest uses the direct-injection paths that always get the block), and it mirrors the existing `{{ tool_context }}` contract. Optional: a one-line doc note for custom-template authors, or bake the block into the composed prompt like the batch nudge.

### N2 — design note (no action): "required=" wording
The per-line `required=[…]` label is an extraction hint with no downstream enforcement; in principle it could nudge over-emission. Mitigated by the header's "…when present" and the additive-safe validator (khora#1546). Surfaced for awareness — matches the author's M2.

### N3 — nit: per-call token cost
The block re-lists every declaring resolved type on every extraction call (single + batch). Bounded/small for this ontology; note only.

Good to merge as-is; **N1** is the only item I'd optionally fold in.

probablyrobot (approved):

PR #1555, opened 2026-07-21
Compare parsed attribute dicts in Neo4j entity-version change-detector

### LINKED ISSUE #1553: Entity version change-detector compares serialized JSON strings, spawning spurious :EntityVersion snapshots on key-order-only differences
## Summary

The Neo4j Phase-2 bi-temporal change-detector decides whether to snapshot a prior `:EntityVersion` by comparing the stored vs incoming attributes **as serialized JSON strings**. Because serialization is insertion-order (`json.dumps`, not `sort_keys`), two extractions of the **same attribute set in a different key order** compare unequal → a spurious `:EntityVersion` snapshot + `SUPERSEDES` edge is created and `version_valid_from` advances, even though nothing semantically changed.

## Where

`src/khora/storage/backends/neo4j.py`, in the batch-upsert Phase-2 versioning pass:

```python
# Compare serialized attributes to detect real changes
if pre["attributes"] == input_row["attributes"]:
    continue
version_rows.append({ ... "old_attributes": pre["attributes"], "new_version_valid_from": input_row["version_valid_from"], ... })
```

Both sides are `serialize_dict(...)` output (JSON strings). Any key-order difference (or whitespace difference) between two equal attribute sets is read as a change.

## Why this surfaces now

Previously, entities extracted on the strict structured-output path always had `attributes == {}` (see the attribute-population work), so this comparison was always `'{}' == '{}'` → it never fired on attributes → no version churn. Now that attributes are populated — as LLM-ordered `{key, value}` pairs folded into a dict, and progressively unioned across chunks during a single ingest — this branch is live. It can emit **false-positive version chains within one ingest**: a re-upsert that merely reorders keys (or a batch that adds one key) snapshots a new `:EntityVersion`. That pollutes point-in-time, `entity_evolution`, and temporal-ordering reads.

## Fix

Compare **order-insensitively** rather than by raw string:

```python
if json.loads(pre["attributes"] or "{}") == json.loads(input_row["attributes"] or "{}"):
    continue
```

(or serialize with `sort_keys=True` everywhere the value is written and compared). Keep any existing empty/NULL guard in front of the comparison. Optionally normalize on the write side too so stored strings are canonical.

## Acceptance criteria

- [ ] A re-upsert whose attributes are the **same set in a different key order** creates **no** new `:EntityVersion` and does not advance `version_valid_from`.
- [ ] A genuine attribute change still creates exactly one snapshot (existing behavior preserved).
- [ ] Regression test covering the key-order-equal case and the genuine-change case.
- [ ] `make format && make test` green.

Context: latent pre-existing behavior, activated by the entity-attribute population work.

filippetrovic:
## Summary
- The Neo4j Phase-2 bi-temporal change-detector compared stored vs incoming entity `attributes` as **serialized JSON strings**. `json.dumps` is insertion-order sensitive (no `sort_keys`), so re-upserting the *same* attribute set in a *different key order* compared unequal — creating a spurious `:EntityVersion` snapshot + `SUPERSEDES` edge and advancing `version_valid_from` with no real change. Once entity attributes became populated, this false-positive version churn polluted point-in-time / entity-evolution / temporal-ordering reads.
- Fix: parse both sides to dicts and compare **order-insensitively** (dict equality) via the existing `deserialize_dict` helper, which also guards against malformed / `NULL` values (defaults to `{}`). The pre-existing empty/`NULL` short-circuit guard and the write-side serialization format are **unchanged** — normalization happens at compare time only.
- Single-site change in `src/khora/storage/backends/neo4j.py`; no new imports.

## Test plan
- [x] Unit + recall + security suites pass locally (`make test-unit`: 10146 passed, 35 skipped, 84% coverage)
- [x] Lint / format / typecheck clean (`make lint`, `make format`)
- [ ] Integration (Neo4j lane) — runs in CI against a real Neo4j service. New `@pytest.mark.integration` regression tests:
  - key-order-only re-upsert → **zero** new `:EntityVersion`, `version_valid_from` unchanged
  - genuine attribute change → **exactly one** snapshot, `version_valid_from` advances
  - empty/`NULL` re-upsert of a populated entity → **zero** (guard preserved)

Fixes #1553

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented unnecessary entity version updates when attributes are unchanged but serialized differently.
  * Numeric values with equivalent meaning are now recognized as unchanged.
  * Genuine attribute changes continue to create new versions, while list reordering remains treated as a change.
* **Tests**
  * Added coverage for attribute comparison, version advancement, and supersession behavior.

### DIFF
--- a/src/khora/storage/backends/neo4j.py
+++ b/src/khora/storage/backends/neo4j.py
@@ -2145,13 +2145,16 @@ class Neo4jBackend(GraphBackendBase):
                     if not pre:
                         continue
 
-                    # Compare serialized attributes to detect real changes.
+                    # Compare parsed attributes order-insensitively to detect real
+                    # changes. Serialization is insertion-order, so the same attribute
+                    # set in a different key order would compare unequal as raw strings;
+                    # parse both sides to dicts before comparing.
                     # An empty/NULL incoming attributes is treated as no change,
                     # mirroring the ON MATCH SET guard above so an empty re-upsert
                     # of a populated entity does not append a phantom version_row.
                     incoming_attrs = input_row["attributes"]
                     incoming_is_empty = incoming_attrs is None or incoming_attrs == "{}" or incoming_attrs == ""
-                    if incoming_is_empty or pre["attributes"] == incoming_attrs:
+                    if incoming_is_empty or _deserialize_dict(pre["attributes"]) == _deserialize_dict(incoming_attrs):
                         continue
 
                     version_rows.append(
--- a/tests/integration/test_neo4j_entity_attributes_guard.py
+++ b/tests/integration/test_neo4j_entity_attributes_guard.py
@@ -49,6 +49,7 @@ Connection parameters are read from env vars with defaults matching the
 
 from __future__ import annotations
 
+import json
 import os
 from collections.abc import AsyncIterator
 from uuid import uuid4
@@ -97,6 +98,24 @@ def _entity(ns_id, name: str, attributes, entity_type: str = "PERSON") -> Entity
         name=name,
         entity_type=entity_type,
         attributes=attributes,
+    )
+
+
+def _entity_with_stamp(ns_id, name: str, attributes, occurred_at: str, entity_type: str = "PERSON") -> Entity:
+    """Like ``_entity`` but pins ``version_valid_from`` deterministically.
+
+    ``_derive_version_valid_from`` resolves ``metadata["occurred_at"]`` first, so
+    stamping it lets the version-chain tests assert an exact
+    ``version_valid_from`` value instead of a clock-derived ``created_at`` that
+    would make an "advances" assertion timing-flaky.
+    """
+    return Entity(
+        id=uuid4(),
+        namespace_id=ns_id,
+        name=name,
+        entity_type=entity_type,
+        attributes=attributes,
+        metadata={"occurred_at": occurred_at},
     )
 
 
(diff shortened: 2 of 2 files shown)


filippetrovic:
## 🤖 Multi-perspective review

**Overall assessment: ✅ APPROVE — no blocking findings.** The fix is correct, minimal, and correctly scoped; the regression tests are strong and non-vacuous. Seven independent reviewer perspectives (security, performance, correctness, tests, database/graph, docs, devil's advocate) plus a dedicated security audit produced **no CRITICAL / HIGH / MEDIUM findings**.

### What was validated
- **Correctness:** order-insensitive dict equality is correct and recursive (nested-dict key reorders also normalize). The `incoming_is_empty` empty/NULL guard is preserved and composes correctly via the `or` short-circuit. Malformed/`None` is handled by the deserialize helper (returns `{}`) and — because `incoming` is always valid serialized output and empty-incoming is short-circuited — this can never mask a genuine change nor wrongly create one. The raw `pre` value is still snapshotted verbatim, so no fidelity loss.
- **Scope:** correct as a single-site, compare-time fix. No sibling bug: relationships have no attribute version-detector, pgvector stores attributes as JSONB (already order-insensitive), and Memgraph does not snapshot `:EntityVersion` on upsert. Write-side `sort_keys` was rightly avoided — it wouldn't fix already-stored rows without a backfill and has a wide blast radius across all graph dict properties.
- **Tests:** hit the MATCH path (assert `is_new` is False **and** `id == seed_id`) so they cannot pass vacuously; preconditions assert same-dict / different-serialized-string; `version_valid_from` advance is pinned deterministically via `occurred_at`. Key-order, genuine-change, int/float, list-reorder, and empty/NULL are all covered.

### LOW / nits (optional, non-blocking)
- **[Performance / Devil's Advocate]** The compare now always runs two `json.loads` per matched entity. An optional raw-string fast-path (`pre == incoming` before parsing) would skip parsing in the common same-order no-change case — strictly behavior-preserving. Parse cost is negligible versus the surrounding Neo4j round-trips, so not required; a cheap win if a follow-up touches the line.
- **[Devil's Advocate / Correctness]** List-value **element order** is still treated as a change (`{"aliases": ["a","b"]}` vs `["b","a"]` still snapshots a version). This is deliberate — only mapping-key order is normalized — and is pinned by a test, but worth recording so future readers don't assume *all* attribute-reorder churn is eliminated (this comment records it).
- **[Database]** On a key-order-only re-upsert, `ON MATCH SET` still overwrites the stored attributes string with the reordered serialization (it deserializes to the same dict; `get_entity` returns the same logical map). Cosmetic only.
- **[Test]** Optional: add a nested-dict key-order case and a `version_valid_from`-unchanged assertion to the empty/NULL test for symmetry.
- **[Docs]** Optional one-line contributor-doc note that the change-detector compares parsed dicts, not serialized strings.

### Informational
- `int`/`float` (`1` vs `1.0`) and `bool`/`int` values now compare equal through JSON normalization, so those type-wobbles no longer version. Intended and benign for extracted attributes; pinned by a test.

## 🔒 Security Audit
**Verdict: PASS — no findings.**
- **Injection:** none — the change is an in-memory dict-equality gate; downstream writes use parameterized Cypher (`tx.run(_VERSION_CYPHER, version_rows=...)`) and the new test helpers use bound parameters (`$ns` / `$name` / `$etype`), not string-built queries.
- **Deserialization:** `_deserialize_dict` is `json.loads` on internally-serialized attribute strings (safe parser, no code-exec semantics; not pickle/yaml) — not a trust boundary, and consistent with the six existing read paths using the same helper.
- **Secrets / crypto / SSRF / file / redirect:** no surface. No secret-typing (ADR-084 style) fields added.
- **Supply chain:** no dependency or lockfile changes → no new supply-chain surface.

---
*Validation: the new Neo4j regression tests run in the CI integration lane (real Neo4j service); the local unit suite passed (10146 passed, 84% coverage), and lint/format/typecheck are clean.*

filippetrovic:
## Independent review — parsed-dict change-detection

Verified against the API patch at head `a66a8cf9` (local clone is on an old pre-B2 `main`, so B2's guard was verified from the patch's unchanged context lines, not the working tree). **APPROVE-WITH-NITS — nothing blocking; good to merge.** CI 26/26, codecov full, CodeRabbit clean.

**All load-bearing points confirmed:**
- **B2's empty-guard is preserved and composes** (the main risk): the line is `if incoming_is_empty or _deserialize_dict(pre["attributes"]) == _deserialize_dict(incoming_attrs): continue`. The `or` short-circuits on `incoming_is_empty`, so an empty/`NULL` re-upsert of a populated entity still creates zero versions — no B2 regression. The guard was placed in front of the compare, not replaced.
- **Order-insensitive parsed-dict compare** via `deserialize_dict` (mixins.py), which returns `{}` on `None`/malformed — correct.
- **Compare-only**: `old_attributes` still snapshots the verbatim stored string; write-side `json.dumps` is unchanged. No fidelity loss, no format churn.
- **Sibling-site hunt clean**: this is the only serialized-string attribute change-detector. Relationship upsert has no property-versioning; retire / soft-retire snapshot unconditionally (no compare); Memgraph/AGE/Neptune carry no `:EntityVersion`. Nothing analogous was left unfixed.
- **Tests are non-vacuous**: they drive the real `upsert_entities_batch` MATCH path (assert `is_new` is False AND `id == seed_id`) and count real `:EntityVersion`/`SUPERSEDES`. `test_key_order_reupsert_creates_no_new_version` fails without the fix; `test_empty_reupsert_creates_no_phantom_version` confirms B2's guard still holds under the composed condition.

**Behavior notes (both fine):** the parsed-dict compare now treats numerically-equivalent values (`1` vs `1.0`) as unchanged (benign — suppresses a re-encoding phantom) and still treats list-value reorders as a change (identical to the pre-fix string behavior — not a regression). Both are pinned by tests. Worth noting attribute values are *not* guaranteed strings at the storage boundary (`_serialize_dict` is bare `json.dumps`; `validate_attributes` only coerces for registered types), so numeric/list JSON can legitimately reach this compare — the fix handles it correctly.

**Nits (informational, non-blocking):**

### N1 — one test isn't fix-discriminating
`test_list_value_reorder_reupsert_creates_a_version` passes with *and* without the fix (both the old string compare and the new dict compare flag a list reorder as a change). It's a valid scope/documentation pin, but it does not guard the fix — don't count it as a regression test.

### N2 — no unit-level fallback for the compare
The guarantee rests entirely on the CI integration lane exporting `NEO4J_INTEGRATION_TEST=1`; there's no unit-level test of the parsed-dict compare. The detector is inline in `upsert_entities_batch`, so a unit test would need the compare extracted into a small helper. Optional — likely not worth the refactor for a one-line change.

Good to merge as-is.

probablyrobot (approved):

PR #1583, opened 2026-08-04
Stop SQLite migrations deleting child rows, and make documents.created_at NOT NULL

filippetrovic:
## Summary

Two changes. The second is the one the ticket asked for; the first is what made it safe to ship.

### The migration environment was deleting rows on every embedded upgrade

`db/migrations/env.py` issued `PRAGMA foreign_keys = ON` on the SQLite migration connection for the whole run (added in #411, with a comment whose rationale had the hazard backwards). Alembic's batch mode implements "alter column" as SQLite's documented table-rebuild procedure — create a temp table, copy rows, `DROP TABLE`, rename — and with enforcement on, that `DROP TABLE` performs an implicit `DELETE FROM` that fires every inbound `ON DELETE CASCADE` **before** the rename puts the table back.

Measured on a populated embedded database upgraded to head:

```
(documents, chunks, keyword_chunks, chronicle_events, chunks_fts)
(1, 1, 1, 1, 1)  ->  (1, 0, 0, 0, 0)
```

The cascade is transitive. Ten revisions batch-alter a table (001, 008, 010, 011, 012, 013, 016, 037, 055, 056). The `documents` rebuilds are the mild case; the `memory_namespaces` rebuilds destroy 11 tables each at their point in the chain. There is a second symptom beyond row loss: from revision `000` the chain cannot advance at all — migration 010's `DROP TABLE memory_namespaces` raises `SQLITE_CORRUPT` with child rows present, the per-migration transaction rolls back, and the database is stranded at its starting revision while `PRAGMA integrity_check` still reports ok.

The pragma is now explicitly `OFF` for the migration connection. It cannot be scoped narrower: the pragma must be issued outside a transaction to be reliable, migrations run under `transaction_per_migration=True`, and Alembic exposes no hook between "transaction begun" and "batch op" that sits outside the transaction. Two narrower forms were tried and rejected — an in-body `PRAGMA foreign_keys = OFF` silently becomes a no-op once a revision runs DML before its batch copy (pysqlite defers the real `BEGIN`), and `PRAGMA defer_foreign_keys = ON` does not stop the cascade at all, because it defers constraint *violation* checking and a cascade is an action, not a violation.

**This prevents future loss. It does not recover anything a prior upgrade already destroyed.** On the `sqlite_lance` stack LanceDB still holds embeddings keyed by the deleted chunk ids, so the residue is orphaned vectors rather than merely missing rows. CHANGELOG carries the per-revision damage table an operator needs to work out what they may be missing.

What is given up: the migration run no longer rejects FK violations it would previously have caught. Nothing in the chain does that today, and the constraints are still enforced on application connections. `PRAGMA foreign_key_check` is deliberately **not** wired in as a guard — it reports clean on the damaged database, because a cascade deletes children precisely so that no violation remains.

### `documents.created_at` is now NOT NULL

Migration `056_documents_created_at_not_null` aligns the physical schema with an ORM declaration it has contradicted since migration 000, which omitted `nullable=False` while `DocumentModel.created_at` has always declared `Mapped[datetime]`.

A NULL breaks a total order over `(created_at DESC, id DESC)` in two ways: SQLite sorts NULLs last under `DESC` and PostgreSQL sorts them first, so one namespace paginates differently per backend; and a keyset cursor predicate over the pair evaluates to NULL for such a row, so it is silently skipped on every page — never returned, never an error.

Existing NULLs are backfilled from `updated_at` where present (nullable too, so the fallback is reachable) and epoch-stamped otherwise, with both counts logged separately so the invented subset stays auditable. The backfill is not reversible; `downgrade()` drops the constraint only. NOT NULL rules out NULL — it does not make an epoch-stamped timestamp *correct*; such a row sorts dead last, which is visibly wrong but reachable, where today it is invisible.

No call-site change has to ship first: unlike #1579, no reachable write path emits a NULL `created_at`.

Not changed, because they already carry the guarantee: the `backend: sqlite` stack, whose hand-written DDL has always had `created_at TEXT NOT NULL` and whose single INSERT path coalesces a missing timestamp to now; and SurrealDB, whose field is `TYPE datetime` (not `option<datetime>`) and so already rejects `NONE`. Both are recorded in place.

## Test plan

- [x] **Populated-database regression test, proven red.** Parametrized over five starting revisions. Against a copy of the tree with `PRAGMA foreign_keys` forced back ON, all five fail — 013/016/054 on row counts, 000/009 on `database disk image is malformed` at migration 010. This mattered: `rows_documents`, `upgrade_completed`, `foreign_key_check`, `integrity_check`, `created_at_notnull` and `revision_at_head` all pass **both** ways, so the obvious test for a `documents` migration is green on the broken code. The subject table survives; only its cascade children die. The assertions had to be put on tables the migration never mentions.
- [x] Migration 056 lifecycle on SQLite: NULL row seeded below the revision is backfilled from `updated_at`, epoch-stamped when `updated_at` is NULL too, explicit-NULL INSERT then fails, indexes / FKs / `server_default` survive the rebuild, `upgrade -> downgrade -> upgrade` round-trips clean.
- [x] Epoch backfill written as a typed bindparam rather than a literal — against a copy of 056 using the raw literal `'1970-01-01T00:00:00+00:00'`, 4 tests in the SQLite lane fail. A literal stores a foreign format and round-trips timezone-aware while every other row is naive, which is the same lexicographic-ordering divergence class the change exists to remove.
- [x] Schema-drift gate: `documents.created_at` removed from the nullability ledger (it is asserted in both directions, so the stale entry was failing) and a nullability-agreement assertion added.
- [x] 17/17 new unit tests pass; 175 migration/drift tests pass.
- [x] `make format` clean, `make lint` exit 0 (107 diagnostics, unchanged from baseline).
- [ ] **The Postgres branch has not been executed anywhere yet.** No Docker and no Postgres was available in the development environment, so `SET NOT NULL`, its heap scan, the `ACCESS EXCLUSIVE` window and the advisory-lock interaction are reasoning from source and documentation, not measurement. Everything asserted about SQLite above *is* measured. CI is the first run of the PG legs — budget one red iteration.

## Operator note

`ALTER TABLE ... SET NOT NULL` takes `ACCESS EXCLUSIVE` and scans the whole heap; the revision's `SET LOCAL lock_timeout` bounds lock *acquisition*, never how long one is held. On a large `documents` table, pre-validate a `NOT VALID` CHECK before deploying to skip the scan (PostgreSQL 12+); the runbook is in the migration docstring and the CHANGELOG.

## Summary by CodeRabbit

* **New Features**
  * Document timestamps are now always populated and required, with missing values safely backfilled.
  * SQLite migrations now protect related data from unintended cascading deletions during schema changes.

* **Bug Fixes**
  * Improved migration reliability while preserving existing records, indexes, and relationships.

* **Documentation**
  * Added guidance on timestamp requirements, migration behavior, risks, and rollback considerations.

* **Tests**
  * Expanded coverage for timestamp backfilling, schema integrity, migration upgrades and downgrades, and data preservation.


filippetrovic:
## Consolidated Review — 7 reviewers + security audit

**Overall assessment: APPROVE (merge-ready), with one structural recommendation the author should decide on before merging.**

No reviewer found a correctness, security, or data-integrity defect. Every substantive claim in the PR was independently re-verified rather than taken on faith, and the two that were checkable by execution both held up.

Reviewers: Database, Correctness, Test, Security, Performance, Docs, Devil's Advocate. Plus a dedicated security audit over the 1763 added lines.

### What was independently confirmed

- **The diagnosis and both rejected alternatives.** Reproduced in bare `sqlite3` 3.53.1: `foreign_keys=ON` → `(parent, child) = (1, 0)`; `OFF` → `(1, 1)`; `defer_foreign_keys=ON` → `(1, 0)`, i.e. it does **not** stop the cascade; an in-body `PRAGMA foreign_keys=OFF` after DML reads back as `1` and the child is still deleted. Both alternatives fail exactly as the PR claims.
- **The regression test has real teeth.** Reverting `env.py` to the broken pragma fails all 5 legs of the populated-chain test (000/009 raise `SQLITE_CORRUPT`; 013/016/054 fail on row counts, `chunks` 1→0). Two further mutants confirm it: a literal-epoch mutant fails 5 tests, and removing the NOT NULL flip fails 4 including both drift-gate directions.
- **The batch rebuild preserves everything.** Full `sqlite_master` diff after 056 is the `notnull 0 → 1` flip and nothing else — all 10 indexes (including both partial and the unique-partial), the FK, the `CURRENT_TIMESTAMP` default, and the triggers survive.
- **Every count and revision list in the docs.** All thirteen figures re-measured against a rebuilt chain. All correct.
- **The backfill population is provably empty on khora-managed databases** — `server_default=now()` since migration 000, `created_at` excluded from both update paths, and an explicit `None` fires the ORM's Python-side default. The backfill is a belt-and-braces path, not a live one.

### HIGH

**H1 — [Devil's Advocate] Consider splitting the PR.** `origin/main` currently has both `PRAGMA foreign_keys = ON` and migration 055's `documents` batch rebuild, so any populated embedded database upgrading through 055 loses `chunks` / `keyword_chunks` / `chronicle_events` **today**. The `env.py` fix is one executable line plus a regression test and depends on nothing; 056 is an irreversible `ACCESS EXCLUSIVE` migration. Bundling means the urgent fix waits on the non-urgent one, and — the sharper point — `git revert <squash>`, the realistic incident response if 056 misbehaves on Postgres, would silently re-arm the data-loss bug.

The counter-argument (shipping 056 alone would ship a known-destructive migration) is satisfied by *ordering*, not by bundling.

This is a judgment call for the author, not a defect. Note `git tag --contains` on the 055 merge is empty — 055 is merged but unreleased, so no operator has run any of this yet, which lowers the urgency somewhat.

**H2 — [Docs] Filed but not substantiated.** The docs reviewer recorded one HIGH but did not produce the finding when asked twice. An independent pass over the verification-claim language in `CHANGELOG.md`, `CLAUDE.md` and the 056 docstring did not reproduce it, and the reviewer's own summary states all counts and revision lists verified correct. **Recording it rather than silently dropping it** — treat as unsubstantiated pending detail.

### MEDIUM

- **M1 — [Database, Performance, Security]** The Postgres lock story is under-documented in three converging ways: a queued `ACCESS EXCLUSIVE` blocks *new readers* for up to the 5s acquisition wait (not just writers, which the docstring implies); the heap-scan estimate is row-count-based when the scan is bytes-driven; and the cumulative `054 + 055 + 056` advisory-lock window is not disclosed anywhere, though each is documented individually.
- **M2 — [Database]** The SQLite table rebuild's transient disk cost is undisclosed — measured 314 MB → 622 MB permanent growth with `auto_vacuum=0` and no `VACUUM` in the path. Embedded deployments on constrained disks should know.
- **M3 — [Test]** `_is_lock_timeout` and its error path are untested; the `entities`-cascade children are unseeded (with a docstring that overstates what the test covers); and `_insert_row`'s FK heuristic can produce a misleading red.
- **M4 — [Performance]** The `NOT VALID` CHECK runbook is incomplete for the PostgreSQL 11 floor, where the optimization does not exist and the scan always runs.
- **M5 — [Devil's Advocate]** `documents.updated_at` carries the identical ORM/live drift and is still ledgered. Postgres accepts both `SET NOT NULL` clauses as one statement — one scan, one lock window — so folding it into 056 would collapse a future 057's fourth `ACCESS EXCLUSIVE` scan on this table into this one. Offered as a tradeoff, not a defect: it is outside the ticket's scope.
- **M6 — [Devil's Advocate]** The same ~800-word explanation is duplicated across `env.py`, `CHANGELOG.md` and `CLAUDE.md` with hand-counted figures that nothing in CI verifies. All correct today; this is drift risk. The populated-chain test already computes cascade closures, so it could assert the figures and let one location own them — the repo's established pattern for exactly this.

### LOW

Roughly a dozen, none blocking: the `env.py` compensating-control note slightly overstates `foreign_key_check`'s coverage (it only spans the 6 seeded tables); `CHANGELOG` omits the advisory-lock multi-service boot-failure mode that the migration docstring does document; the 056 docstring is 203 lines over ~40 executable statements (disproportionate, but it conforms to the convention 055 set one PR earlier — a repo-level conversation, not a strike against this PR); and a pre-existing `_SCHEMA_SQL.split(";")` footgun in `sqlite.py`, correctly flagged in a comment and out of scope here.

### Security Audit

**Verdict: PASS.** 0 CRITICAL, 0 HIGH, 0 MEDIUM.

- No injection: every value is bound. The three f-string SQL sites are test-only, interpolating table and column names reflected from a local scratch database, with values bound as parameters.
- No secrets: the only credential-shaped string is `khora:khora@localhost:5434` in a test docstring, matching the repo's own compose file.
- No dependency, endpoint, deserializer, file-path, redirect, or SSRF surface touched.
- **ADR-084 not applicable** — independently verified, no Pydantic field added or touched.
- **The FK-disable integrity trade-off was examined specifically for multi-tenant risk** and is sound: the only parent-row `DELETE` in the chain is Postgres-only (the pragma is SQLite-only), constraints remain enforced on all application connections, and there is no cross-tenant path.

### On the "unverified Postgres" caveat

Every reviewer was instructed to label Postgres claims as reasoning rather than measurement, because no Docker or Postgres was available during development. **CI has now closed that gap**: 26 checks green, 0 failures, including `conformance (postgres)`, `test-integration` and every backend leg. The Postgres branch has now executed. A tripwire (`KHORA_PG_REQUIRED=1` + a `pytest.exit` in `tests/integration/conftest.py`) means the PG lane cannot skip silently on a green job — verified at source, after one reviewer initially flagged that risk and then retracted it.

---

*None of the MEDIUM or LOW findings block merge. H1 is a structural call for the author.*

filippetrovic:
## Review findings addressed — `af7f058f`

The docs reviewer's HIGH turned out to be a real operator-facing defect, and it is now fixed. Recording it here because it is the kind of thing worth being explicit about.

**`CHANGELOG.md` was telling operators to destroy intact databases.** The entry described a database stranded by the cascade bug as "permanently unupgradable" and said "the chain can never complete". That is false: when the cascade trips, the per-migration transaction rolls back before writing anything, so no rows are lost and the database is simply parked at its starting revision. Re-running the upgrade on this version walks the whole chain to head with every row intact.

The failure mode was specific and plausible. The operator whose embedded upgrade is *visibly failing* is the most self-identifying group affected by this bug. They read "permanently unupgradable", then two sentences later "does not recover anything a prior upgrade already destroyed", and conclude the release does not help them. The remedy for a permanently unupgradable database is to rebuild it — so they discard a completely intact database that this release repairs. Data destroyed by the changelog rather than by the bug.

The stranded case is now stated as **the one case this change repairs rather than merely prevents**, in both `CHANGELOG.md` and `CLAUDE.md`. It was also the only claim in the PR that its own sibling documents contradicted — `env.py` and `CLAUDE.md` both said only "stranded at its starting revision", because that is what it is.

**Three further corrections:**

- The "am I affected" guidance now states that `PRAGMA foreign_key_check` reports **clean** on a damaged database — so the obvious check falsely reassures — and that comparing relational counts against the vector store only detects the `chunks` loss, not the other tables an early crossing destroyed. Added the `documents > 0 AND chunks = 0` signature.
- The `backend: sqlite` tier is no longer listed as flatly "Not affected". It writes `isoformat()` into a `TEXT` column ordered lexicographically, so its order still diverges across differing UTC offsets. The 056 docstring already scoped this carefully; the two documents readers actually consult did not.
- The `env.py` comment now records *why* `foreign_key_check` cannot serve as the guard that would make re-enabling the pragma safe, and names #411 as the PR that first set it wrong — so the next maintainer can see it was already tried. That comment is what someone reads immediately before turning the pragma back on, and "add `foreign_key_check` and re-enable" was the most likely future edit at that exact line.

Also retired a disclosure that would have become false on merge: the 056 docstring said the Postgres round-trip "still first executes in CI". Accurate at review time; the PG legs have since run green, so it is now past tense.

**Not addressed, deliberately:** the remaining MEDIUM/LOW findings are docstring precision and completeness items against a house standard this repo sets unusually high (lock-queue mechanics, SQLite transient disk cost, the `NOT VALID` runbook's PG 11 floor and pre-validation ordering, `updated_at` folding). None affects correctness. The `updated_at` suggestion in particular is outside this change's scope and is a real tradeoff worth its own decision, not a drive-by.

CI green on `af7f058f`: 26 checks, 0 failures, all Postgres legs included.

filippetrovic:
## Final state — CI green on `c197bf78`

**26 checks, 0 failures**, all Postgres legs included.

Three commits since the review comment above:

- **`af7f058f`** — the docs reviewer's HIGH (the "permanently unupgradable" claim) plus three smaller doc inaccuracies. Detailed in the previous comment.
- **`12435c04`** — the test reviewer's three MEDIUMs, each now proven closed by a mutant rather than asserted:
  - The `entities` leg asserted nothing. The docstring claimed seeding `entities` covered the 008 rebuild family; it does not, because **008 rebuilds `entities` itself and the rebuilt table is always the survivor** — the same trap that makes a `documents`-only assertion pass on broken code. Its children were unseeded, and every existing start revision was already *past* 008, so nothing crossed that rebuild at all. Seeds the children and adds a start revision below 008; all six legs now die on the broken tree.
  - `_insert_row` invented a UUID for any unsupplied `*_id` column. No such column is FK-constrained today, but the day one is, this gate would fail confidently naming the wrong cause — a dangling reference reported as the data loss it exists to detect. It now raises at seed time naming the column.
  - `_is_lock_timeout` and the upgrade error path had no coverage while the docstring makes a specific claim about them. Five server-free tests; a mutant narrowing the check to `pgcode` alone kills exactly the two that should die.
- **`c197bf78`** — removes an unrelated local tooling block that a blanket `git add` swept in. Caught by the backend implementer on review of my own commit, not by me. Net diff for that file is now only the two on-topic lines.

**One CI note for the record:** `test-integration` went red once on `c197bf78` with a pytest 120s timeout in `tests/integration/matrix/test_chronicle_pg.py`, while the log filled with HuggingFace `HTTP 429` responses revalidating the reranker model. External rate-limiting, not this change — the two commits before it touched only `tests/unit/db/*` and a markdown file, and the same job passed on both. Re-ran the job; green.

Ready to merge, subject to the two open questions from the review: whether to split the `env.py` fix into its own PR (H1), and whether to fold `updated_at` into 056 to collapse a future fourth blocking migration on this table.

msrdic (approved):

PR #1585, opened 2026-08-05
Read the migration head from the script directory in tests

filippetrovic:
## Summary

Migration tests spelled the chain head out as a literal (`055_documents_source_type_alignment`). Landing the previous migration therefore required the same mechanical bump in six files, and each of those bumps conflicts with any other migration PR in flight. This removes the bump.

- **One shared helper**, `tests/test_helpers/alembic_head.py::current_head()`, resolves the head through Alembic's `ScriptDirectory`. It is deliberately **uncached** — a memoized head would survive a migration being added mid-session and hand back a stale answer — and is documented as call-inside-the-test-body, because building the script directory at module scope turns a broken chain into a collection error that takes unrelated tests down with it.
- **Only *head* references became dynamic.** Explicit upgrade/downgrade targets keep their literals: those tests are about one specific revision and must not drift forward when the next migration lands. `test_migration_052/053/054`'s own target/predecessor constants are untouched.
- **`HEAD_REVISION` → `TARGET_REVISION`** in the 055 lane. That constant is 055's own upgrade target and merely *happens* to equal the head today; the rename records which of the two it is, so the next migration doesn't silently make it wrong.

### The file-count assertion: floor, not deletion

The exact `_EXPECTED_MIGRATION_COUNT = 56` had the same bump-on-every-PR problem, but deleting it outright would have opened a hole. It was load-bearing in one direction: it catches a **truncated tail**, and nothing else does. `test_migration_chain_is_contiguous` is invariant under truncation — delete the newest revision and the survivors are still a valid contiguous chain from `down_revision=None`. The fallback "each migration's own test pins its revision id" doesn't hold either; several revisions have no dedicated test module.

So the count becomes a **floor** (never bumped when a migration is added — only raised deliberately to tighten), plus a chain-resolution check that covers what a floor cannot: Alembic skips a file its *filename filter* excludes — an editor lockfile such as `.#zz.py` is skipped while still being globbed here — so a file can exist, count toward the floor, and belong to no chain; the walk count rejects it. Anything else unreadable raises outright rather than being skipped (`SyntaxError`, or `CommandError` for a file with no `revision`). Resolving the head additionally rejects a stray revision nothing points at, which reads as a second head.

**Known limits, stated rather than implied:** the floor catches a truncated tail only while the file count still equals it — once migration 056 lands, 57 files against a floor of 56 means deleting the newest revision passes every gate here. Both sides of the walk count read the same directory, so it is a self-consistency check: it sees a file present-and-unreachable and is blind to a file simply absent. The test docstring spells this out. A stray revision chained *onto* head is likewise not detected, where the old head literals would have caught it.

## Test plan

- [x] `make format` clean; `make lint` exits 0 (the `ty` diagnostics it prints are pre-existing warnings in `src/`, none in changed files)
- [x] Unit migration suites green — `tests/unit/test_migration_bundling.py`, `test_sqlite_migrations.py`, `test_migration_drift.py`, `tests/unit/db/`: **158 passed**
- [x] **The property this buys, demonstrated end to end.** Added a throwaway no-op revision chained to head, confirmed the resolved head moved to it, and re-ran the unit + integration migration suites with **zero test edits**: `172 passed, 54 skipped`. Throwaway then removed; `git status` clean under `src/`.
- [x] **Counterfactual, so the proof isn't vacuous.** Same throwaway against the *pre-change* tests fails exactly where expected — `test_versions_dir_is_fully_bundled` and `test_upgrade_head_fresh_sqlite` — `2 failed, 57 passed`.
- [x] Residue audit: every surviving revision literal under `tests/` is a migration's own revision, its predecessor, or an explicit walk destination. No head reference remains.
- [ ] **PostgreSQL lanes not executed locally** — no Docker in the dev container, so those tests skip rather than run (54 skips above). They run in CI.

Full-suite note: the unit run reports 69 failures / 20 errors in `langgraph`, `filter`, `sqlite_lance`, and `dream` suites. These reproduce identically on clean `main` in this container (missing optional deps) and are unrelated to this change, which touches only `tests/`.

## Summary by CodeRabbit

- **Tests**
  - Improved database migration validation to track the current migration state, reducing failures when new migrations are added.
  - Expanded migration packaging checks to detect missing, malformed, unreachable, or improperly bundled migrations.
  - Refined targeted migration tests to keep upgrade, rollback, and backfill scenarios isolated and reliable.
  - Added coverage for migration-chain updates and missing migration heads.

- **Chores**
  - Updated shared migration test utilities and revision naming for clearer, more maintainable test coverage.

### DIFF
--- a/tests/integration/db/test_migration_032_dream_runs.py
+++ b/tests/integration/db/test_migration_032_dream_runs.py
@@ -35,6 +35,7 @@ from alembic.config import Config
 from sqlalchemy.ext.asyncio import AsyncEngine, create_async_engine
 
 from khora.db.session import run_migrations
+from tests.test_helpers.alembic_head import current_head
 
 DATABASE_URL = os.environ.get(
     "KHORA_DATABASE_URL",
@@ -304,7 +305,7 @@ class TestMigration032OnSqlite:
                 async with engine.connect() as conn:
                     # Chain reached head (sanity).
                     result = await conn.execute(sa.text("SELECT version_num FROM khora_alembic_version"))
-                    assert result.scalar() == "057_drop_documents_created_at_index"
+                    assert result.scalar() == current_head()
 
                     # khora_dream_runs exists on SQLite (#896) so dream_history /
                     # dream_status work on the embedded sqlite_lance stack.
--- a/tests/integration/db/test_migration_041_chunks_denormalized_columns.py
+++ b/tests/integration/db/test_migration_041_chunks_denormalized_columns.py
@@ -41,6 +41,7 @@ from alembic.config import Config
 from sqlalchemy.ext.asyncio import AsyncEngine, create_async_engine
 
 from khora.db.session import run_migrations
+from tests.test_helpers.alembic_head import current_head
 
 DATABASE_URL = os.environ.get(
     "KHORA_DATABASE_URL",
@@ -322,7 +323,7 @@ class TestMigration041OnPostgres:
                 async with engine.connect() as conn:
                     # Chain reached head.
                     result = await conn.execute(sa.text("SELECT version_num FROM khora_alembic_version"))
-                    assert result.scalar() == "057_drop_documents_created_at_index"
+                    assert result.scalar() == current_head()
 
                     # The migration did not create the table.
                     result = await conn.execute(
@@ -353,7 +354,7 @@ class TestMigration041OnSqlite:
             try:
                 async with engine.connect() as conn:
                     result = await conn.execute(sa.text("SELECT version_num FROM khora_alembic_version"))
-                    assert result.scalar() == "057_drop_documents_created_at_index"
+                    assert result.scalar() == current_head()
             finally:
                 await engine.dispose()
 
--- a/tests/integration/db/test_migration_042_widen_chunks_source_external_id.py
+++ b/tests/integration/db/test_migration_042_widen_chunks_source_external_id.py
@@ -43,6 +43,7 @@ from alembic.config import Config
 from sqlalchemy.ext.asyncio import AsyncEngine, create_async_engine
 
 from khora.db.session import run_migrations
+from tests.test_helpers.alembic_head import current_head
 
 DATABASE_URL = os.environ.get(
     "KHORA_DATABASE_URL",
@@ -69,7 +70,6 @@ pytestmark = pytest.mark.integration
 
 
 _MIGRATIONS_DIR = Path(__file__).resolve().parents[3] / "src" / "khora" / "db" / "migrations"
-_HEAD = "057_drop_documents_created_at_index"
 _PREV = "041_khora_chunks_denormalized_columns"
 
 
@@ -296,7 +296,7 @@ class TestMigration042OnPostgres:
             try:
                 async with engine.connect() as conn:
                     result = await conn.execute(sa.text("SELECT version_num FROM khora_alembic_version"))
-                    assert result.scalar() == _HEAD
+                    assert result.scalar() == current_head()
 
                     result = await conn.execute(
                         sa.text(
@@ -326,7 +326,7 @@ class TestMigration042OnSqlite:
             try:
                 async with engine.connect() as conn:
                     result = await conn.execute(sa.text("SELECT version_num FROM khora_alembic_version"))
-                    assert result.scalar() == _HEAD
+                    assert result.scalar() == current_head()
             finally:
                 await engine.dispose()
 
--- a/tests/integration/db/test_migration_044_chunks_backfill.py
+++ b/tests/integration/db/test_migration_044_chunks_backfill.py
@@ -65,6 +65,7 @@ from alembic.config import Config
 from sqlalchemy.ext.asyncio import AsyncConnection, AsyncEngine, create_async_engine
 
 from khora.db.session import run_migrations
+from tests.test_helpers.alembic_head import current_head
 
 DATABASE_URL = os.environ.get(
     "KHORA_DATABASE_URL",
@@ -93,7 +94,6 @@ pytestmark = pytest.mark.integration
 _MIGRATIONS_DIR = Path(__file__).resolve().parents[3] / "src" / "khora" / "db" / "migrations"
 
 _PREV_REVISION = "043_khora_chunks_metadata_backfill"
-_HEAD_REVISION = "057_drop_documents_created_at_index"
 
 _TSV_TRIGGER = "khora_chunks_content_tsv_update"
 
@@ -781,7 +781,7 @@ class TestMigration044OnPostgres:
             try:
                 async with engine.connect() as conn:
                     result = await conn.execute(sa.text("SELECT version_num FROM khora_alembic_version"))
-                    assert result.scalar() == _HEAD_REVISION
+                    assert result.scalar() == current_head()
 
                     result = await conn.execute(
                         sa.text(
@@ -811,7 +811,7 @@ class TestMigration044OnSqlite:
             try:
                 async with engine.connect() as conn:
                     result = await conn.execute(sa.text("SELECT version_num FROM khora_alembic_version"))
-                    assert result.scalar() == _HEAD_REVISION
+                    assert result.scalar() == current_head()
             finally:
                 await engine.dispose()
 
--- a/tests/integration/db/test_migration_049_hook_subscriptions.py
+++ b/tests/integration/db/test_migration_049_hook_subscriptions.py
@@ -45,6 +45,7 @@ from khora.db.session import run_migrations
 from khora.hooks.dispatcher import HookDispatcher
 from khora.hooks.models import SemanticFilter
 from khora.hooks.subscription_store import HookSubscriptionStore, PersistentSubscription
+from tests.test_helpers.alembic_head import current_head
 
 DATABASE_URL = os.environ.get(
     "KHORA_DATABASE_URL",
@@ -71,7 +72,6 @@ pytestmark = pytest.mark.integration
 
 _MIGRATIONS_DIR = Path(__file__).resolve().parents[3] / "src" / "khora" / "db" / "migrations"
 
-_HEAD = "057_drop_documents_created_at_index"
 _PREV = "048_dream_conflicts_reconcile"
 
 
@@ -174,7 +174,7 @@ class TestMigration049Schema:
                     assert "ix_khora_hook_subscriptions_ns_event" in idx_names
 
                     ver = await conn.execute(sa.text("SELECT version_num FROM khora_alembic_version"))
-                    assert ver.scalar() == _HEAD
+                    assert ver.scalar() == current_head()
             finally:
                 await engine.dispose()
 
--- a/tests/integration/db/test_migration_052_entities_source_chunk_ids_gin.py
+++ b/tests/integration/db/test_migration_052_entities_source_chunk_ids_gin.py
@@ -64,7 +64,7 @@ pytestmark = pytest.mark.integration
 
 _MIGRATIONS_DIR = Path(__file__).resolve().parents[3] / "src" / "khora" / "db" / "migrations"
 
-_HEAD = "052_entities_source_chunk_ids_gin"
+_TARGET = "052_entities_source_chunk_ids_gin"
 _PREV = "051_documents_graph_mirror_pending"
 
 _INDEX_NAME = "ix_entities_source_chunk_ids_gin"
@@ -100,7 +100,7 @@ class TestMigration052GinIndex:
 
         # Migration 052 builds the GIN index. Idempotent when the shared dev
         # DB is already at head.
-        command.upgrade(cfg, _HEAD)
+        command.upgrade(cfg, _TARGET)
         rows = asyncio.run(_index_rows(DATABASE_URL))
         assert len(rows) == 1, f"expected the GIN index at head, found {rows}"
         assert rows[0][0] == _INDEX_NAME
@@ -112,7 +112,7 @@ class TestMigration052GinIndex:
             assert asyncio.run(_index_rows(DATABASE_URL)) == []
         finally:
             # Always restore the true chain head so the shared dev DB is never
-            # left rewound — "head" (not the pinned _HEAD) stays correct once a
+            # left rewound — "head" (not the pinned _TARGET) stays correct once a
             # future migration lands on top of 052.
             command.upgrade(cfg, "head")
 
--- a/tests/integration/db/test_migration_054_documents_namespace_created_at_id.py
+++ b/tests/integration/db/test_migration_054_documents_namespace_created_at_id.py
@@ -78,7 +78,7 @@ skip_no_pg = pytest.mark.skipif(
 
 _MIGRATIONS_DIR = Path(__file__).resolve().parents[3] / "src" / "khora" / "db" / "migrations"
 
-_HEAD = "054_documents_namespace_created_at_id"
+_TARGET = "054_documents_namespace_created_at_id"
 _PREV = "053_khora_chunks_bookkeeping_to_chunker_info"
 
 # The revision that first added the 2-column index, and the one immediately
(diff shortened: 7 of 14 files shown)


filippetrovic:
# Consolidated review

Seven reviewers (correctness, tests, database, security, performance, docs, devil's advocate) plus a security audit. Findings deduplicated and severity-ranked. Most were verified empirically against mutated copies of the migration chain rather than argued from reading — those are marked **[verified]**.

**Overall: approve, with two coverage regressions worth a decision before merge.** Nothing here is broken; CI is 26/26 green. The concerns are about guards this PR removes and one docstring claim that is factually wrong.

---

## HIGH

**H1 — A stray migration file chained onto head is now caught by nothing.** `tests/unit/test_migration_bundling.py` **[verified]**

Commit a leftover `alembic revision --autogenerate` scratch file under `versions/` whose `down_revision` points at the real head. It passes the floor (57 ≥ 56), resolves cleanly as the new head, matches the walk-length check, and satisfies contiguity — so it ships in the wheel and executes on the next `run_migrations()`.

The six head literals were the only thing that caught this: `command.upgrade(cfg, "head")` stamped the stray revision and every `assert == "055_..."` went red. `current_head()` now follows the stray, so the assertions agree with it. This is the one guard the sweep genuinely removed, and the PR body does not mention it.

*Options:* assert the head equals the highest-numbered file in `versions/` (cheap, catches the stray, still no per-migration edit); or accept it and say so in the docstring.

## MEDIUM

**M1 — The floor stops working the moment migration 056 lands, permanently.** `test_migration_bundling.py:24` **[verified]** — flagged independently by three reviewers.

`_MIGRATION_COUNT_FLOOR = 56` equals today's file count exactly, so truncation is covered *today* and never again: at 57 files against a floor of 56, deleting the newest revision passes the floor, passes contiguity (invariant under truncation), and shrinks the walk count in lockstep. `len(walk_revisions()) == len(migration_files)` cannot close this — both sides read the same directory.

The docstring's premise ("the count catches a truncated tail, and nothing else does") is also too strong. Deleting `055_...` from the tree fails **five** tests, four of which are the ORM drift gate: `test_orm_indexes_are_built`, `test_orm_not_null_is_installed`, `test_documents_alignment_is_not_in_the_ledger`, `test_upgrade_succeeds_when_index_pre_exists`. The drift ratchet is the real backstop for any migration with an ORM-visible footprint — which is most of them.

So the floor's non-duplicated coverage is "a truncated migration with no ORM-visible footprint (index-only, backfill-only, PG-only), during the window where the count still equals the floor" — and that window closes at the next migration. Raising the floor would reintroduce the exact per-migration edit and same-line conflict this PR removes, so it never will be raised.

*Recommendation:* delete `_MIGRATION_COUNT_FLOOR` and the `>=` assertion — the ticket's original instruction was right. `assert migration_files` covers "the versions dir ships nothing" without a number. Amend the docstring to credit `test_migration_drift.py` as the backstop and narrow the stated residual gap.

**M2 — The helper resolves the repo tree, not the installed package.** `tests/test_helpers/alembic_head.py:32` — flagged by four reviewers.

`current_head()` imports `MIGRATIONS_DIR` from `schema_drift` (`parents[2]/src/khora/db/migrations`, repo-relative), while `test_migration_bundling.py:20` deliberately uses `khora.db.migrations.__file__` — the *installed* package — because that is what makes it a packaging tripwire. Identical under the editable install CI uses; divergent under a wheel, where the swept tests would compare a version row written by the packaged chain against a head read from the repo tree.

Second cost: `schema_drift` pulls in `sqlalchemy`, `pgvector.sqlalchemy`, `khora.db.models` and `_schema_config` — **762 ms, ~2100 modules [verified]** — so a break anywhere in the drift gate's import chain becomes a collection error in six migration modules that exercise none of it.

*Fix:* `MIGRATIONS_DIR = Path(khora.db.migrations.__file__).parent` locally in `alembic_head.py`.

**M3 — The docstring's central claim is wrong; the case it privileges is the weakest, not the strongest.** `alembic_head.py:11-16` **[verified]** — flagged by three reviewers independently.

The docstring says the hook-subscriptions case carries real signal because `run_migrations()` returns `success=True, skipped=True` when the DB is ahead and the fixture checks only `success`. But the skip path (`env.py:237-245`) requires `current_rev is not None`, and that fixture's `_reset_public_schema` (`test_migration_049_hook_subscriptions.py:85-102`) does `DROP SCHEMA public CASCADE` and recreates `khora_alembic_version` **empty** — so `current_rev` is always `None` and ahead-detection can never fire. Every other failure sets `success=False`, which the fixture does assert.

So **all 12** swapped assertions are decoration today, not 11. That does not make the sweep wrong — the failures that matter raise out of `command.upgrade` before the assertion is reached — but the docstring inverts its own ranking and should say "would become load-bearing if the fixture ever seeded a revision row."

**M4 — Two claims overstate Alembic's behavior.** The PR body says "Alembic's loader skips a `versions/*.py` it can't read rather than failing." The filter is `(?!\.\#|__init__)(.*\.py)$` — it skips only `.#`-prefixed and `__init__` names; anything else raises. The code docstring states this correctly; the PR body does not. The walk-length check's genuine residue is therefore *only* an editor lockfile such as `.#zz.py` **[verified: `glob("*.py")` matches it, `startswith("__")` does not drop it, Alembic's regex does]** — which makes it more a false-positive generator (an unsaved buffer turns a packaging test red) than a safety net.

**M5 — `test_versions_dir_is_fully_bundled` lost its stated reason for existing.** The old docstring led with the purpose: `_MIGRATIONS_DIR` resolves through the *installed* package, so the test fails if a build config stops shipping `versions/*.py`, leaving `run_migrations()` silently building a partial schema. The rewrite spends 35 lines on the floor's limitations and never states that consequence. Restore one sentence — it is what stops someone "simplifying" `_MIGRATIONS_DIR` to a repo-relative path later.

## LOW

- **L1** `test_migration_bundling.py:500` — `assert script.get_current_head() is not None` is unreachable. A branched chain raises `CommandError: multiple heads` **[verified]** before the assert evaluates; the only `None` case is an empty versions dir, already excluded by the floor two lines above. Keep the bare call for its raise, drop the assert, or fold the branch note into the message.
- **L2** `alembic_head.py:51` — `assert` guards a control-flow invariant in a non-test module, so `python -O` strips it and the declared `-> str` is violated (returns `None`). Use `if head is None: raise AssertionError(...)`.
- **L3** `test_migration_052_...py:67`, `test_migration_054_...py:81` — both still name their own revision `_HEAD`, the exact confusion the `HEAD_REVISION` → `TARGET_REVISION` rename removed for 055. `_HEAD = "052_..."` is used as an explicit upgrade target; a future reader may "fix" it to `current_head()` and silently stop testing 052. Rename to `_TARGET`.
- **L4** No test covers `current_head()`. The property this PR exists to buy is proven once by hand, not enforced in CI — so adding `@lru_cache` to shave the measured 4-8 ms would silently reintroduce staleness with nothing failing. ~10 lines would close it (precedent: `tests/unit/test_diagnostics_helper.py`).
- **L5** `test_migration_bundling.py:469` — "several revisions... referenced nowhere in `tests/`" understates: **24 of 56 [verified]**. The argument holds a fortiori; "nearly half" reads truer without becoming a second number to bump.
- **L6** `test_migration_bundling.py:493` — the `startswith("__")` filter is wider than Alembic's, and the failure message omits the offending filename: `Chain resolves 2 revisions but 1 migration files are shipped: ['000_a.py']` **[verified]** names the file that is fine, not the one that isn't.
- **L7** `test_migration_drift.py:787` — comment says "Stop one revision short of 054"; the migration under test is 055. Pre-existing on `main`, not introduced here.

## Security audit

**PASS — no findings.** Test-only diff, no `src/` changes, no dependency manifests touched. Verified across all 16 categories: no secrets or DSNs introduced (the `KHORA_DATABASE_URL` localhost defaults are pre-existing context lines); no Pydantic model or secret-named field, so ADR-084 is not engaged; public-repo hygiene clean across diff, commit message, and PR body (no tracker IDs, ADR/PRD numbers, or internal hostnames). The one new execution surface — `ScriptDirectory.from_config` importing all 56 version modules — is not a new trust boundary: `run_migrations()` already imports and executes those same first-party modules from the same directory in these very tests. A bare `Config()` performs no `sys.path` mutation, loads no `.pyc`, reads no `alembic.ini` from cwd, and never touches `sqlalchemy.url`.

## Performance

**No findings.** Measured: `get_current_head()` costs **4.1 ms steady-state**; the full suite makes **10 script-directory builds totalling ~45 ms**. Every call site sits in a distinct non-parametrized test body, and xdist does not multiply it (each test runs on one worker). The uncached decision is defensible on measured cost — ~1% of the 0.47 s test it lives in — and the staleness argument for it is legitimate.

---

## Assessment

**APPROVE.** The core change is right and the strongest argument against it — that a literal is an independent oracle and a derived value is a tautology — does not survive: the alarm the literals produced was "your branch added a migration," and the only response to it was a mechanical bump. A tripwire whose sole response is "make it green" is not a tripwire.

Two things are worth deciding before merge rather than after: **H1** (stray-revision detection, genuinely lost) and **M1** (the floor, which buys one migration's worth of coverage and then decays silently). **M2** and **M3** are cheap, unambiguous corrections. The rest are follow-ups.

filippetrovic:
## Correction to the review above, and two additions

Late reports from the implementation team landed after I posted. One of my MEDIUM findings quoted text that is **not** in the shipped code, and I'd rather correct that than leave it standing.

**M1 is partly wrong — withdrawn in its "the docstring overstates" form.** I wrote that the docstring's premise is *"the floor catches a truncated tail, and nothing else does."* That sentence was in an earlier revision of this branch; QA caught the overclaim before the commit and the implementer had already fixed it. What actually shipped concedes the decay in full, unprompted:

> It catches a *truncated tail* only while the file count still equals the floor. Once a migration lands beyond it — 57 files against a floor of 56 — deleting the newest revision passes here, and passes every other gate too […] Both sides of that count come from this one directory, so it is a self-consistency check: it sees a file that is present and unreachable, and is blind to a file that is simply absent.

It also inverts the "each migration's own test pins its revision id" line into an explicit do-not-lean-on-this warning. So the code documents this limitation more honestly than my finding gave it credit for. My apologies for the misquote.

**What survives from M1 is the design question, not the documentation.** The floor still buys one migration's worth of truncation coverage and then decays permanently, and that is now a conscious, documented trade rather than an unnoticed one. If you want the property back without the per-migration edit, the database reviewer's suggestion is the cheapest form:

```python
assert "055_documents_source_type_alignment" in {r.revision for r in walked}
```

— an anchor revision bumped only when the floor is, which pins the chain to a known point instead of a bare count.

**M4 was accurate and the fault was mine** — the overstatement about Alembic's loader was in *my* PR description, not the code docstring, which had it right. The PR body is now corrected, and I've added an explicit "known limits" paragraph so the floor's decay and the H1 stray-revision gap are stated in the description rather than only in this thread.

**M3 confirmed independently, and it is the one documentation defect that is real.** `tests/test_helpers/alembic_head.py:11-16` still says the hook-subscriptions case is *"the one thing standing between a skipped migration and a green test."* Three reviewers reached this separately: `_reset_public_schema` does `DROP SCHEMA public CASCADE` and recreates `khora_alembic_version` empty, so `current_rev` is always `None`, `_DatabaseAheadError` cannot fire, and every other failure sets `success=False` — which the fixture does assert. All 9 sites are decoration today, not 8.

**M2 is stronger than I stated: it contradicts the team's own design ruling.** The architect's plan explicitly said to build a local 3-line `Config` and *not* to import from `schema_drift`, on the grounds that it "drags pgvector + the full ORM into unit-lane test modules." The shipped helper imports `MIGRATIONS_DIR` from `schema_drift` anyway, which is exactly the outcome that instruction existed to prevent. The one-line local `Path(khora.db.migrations.__file__).parent` fixes both this and the repo-tree-vs-installed-package mismatch.

**Coverage note worth recording:** `test_migration_049_hook_subscriptions.py:177` is the only site carrying net-new signal, and it was never executed locally — no Docker in the dev container, so every PG lane skipped. CI's Postgres jobs are green, which is the real gate for it.

H1 and M5 stand as written; L1–L7 unchanged.

filippetrovic:
## Review findings addressed — pushed in `427e4f3`

All items from the review above are resolved. Two required deviating from the literal instruction; both are called out rather than silently reinterpreted.

| Finding | Change |
|---|---|
| **M1** | `_MIGRATION_COUNT_FLOOR` and the `>=` assertion deleted; `assert migration_files` kept. Docstring now credits `tests/unit/test_migration_drift.py` (the ORM drift ratchet) as the truncation backstop instead of claiming nothing else catches it. |
| **H1** | Anchor-revision assertion added — `_ANCHOR_REVISION in {r.revision for r in walked}` — bumped only deliberately, never per migration. **See the caveat below: this does not close H1.** |
| **M2** | `alembic_head.py` is self-contained: `MIGRATIONS_DIR = Path(khora.db.migrations.__file__).parent`. Kills the `schema_drift` import chain (measured 762 ms / ~2100 modules into pure-Alembic unit modules) and reads the installed package, matching `run_migrations()`. |
| **M3** | Helper docstring corrected: all sites are decoration today; the hook-subscriptions case "would become load-bearing if that reset ever stopped clearing the version table." |
| **M5** | One-sentence purpose restored — installed-package resolution is what makes this a packaging tripwire. |
| **L1/L2/L3/L4/L6** | Bare `get_current_head()` kept for its raise (see below); `raise AssertionError` instead of bare `assert`; `_HEAD` → `_TARGET` in the 052/054 modules; new `tests/unit/test_alembic_head_helper.py`; orphan files now named in the failure message. |

### The anchor does not close H1 — verified, not assumed

I mutation-tested each hazard against scratch copies of `versions/`:

| mutation | result |
|---|---|
| anchor file deleted | **FAIL anchor** |
| tail truncated (newest) | **FAIL anchor** |
| branched / unreachable revision | **FAIL** `CommandError` |
| editor lockfile `.#lock.py` | **FAIL orphan** — names the file |
| syntax error / no `revision` | **FAIL** `SyntaxError` / `CommandError` |
| legitimate new migration | **PASS** (the bought property) |
| **stray revision chained onto head** | **PASS — still not caught** |

A stray chained onto head leaves the anchor reachable, so the anchor cannot see it. Only the head literals this PR removes would have caught it, and restoring them reinstates the per-PR edit the ticket exists to kill. `assert head == _ANCHOR_REVISION` would catch it but requires bumping on every migration — same problem. So H1 is closed only at a price the ticket forbids; it is now stated explicitly in the docstring rather than left to be discovered.

Note the anchor inherits the floor's decay in one respect: it catches tail truncation today because 055 *is* the newest, but once 056 lands and the anchor stays at 055, deleting 056 passes again. The drift ratchet is the real backstop there.

### A regression I introduced and caught

Dropping the "unreachable" assert (L1) initially removed branched-chain detection outright. `ScriptDirectory.from_config` and `walk_revisions()` both traverse two heads **without raising** — only `get_current_head()` raises. The bare call is now kept deliberately, with a comment saying why. The mutation row above is the guard.

### Verification

- Unit migration suites: **161 passed**. `make format` clean, `make lint` exit 0.
- Zero-edit property re-proven on the final tree: throwaway `056` chained to head → head resolved to it → **175 passed, 54 skipped** with no test edits. Throwaway and its `.pyc` removed (`test_no_orphan_pyc_files` caught the stale `.pyc` on the first attempt — the suite working as intended), `git status -- src/` empty.
- **CI 26/26 green**, including `test-integration`, which exercises `test_migration_049_hook_subscriptions.py:177` — the one site with net-new signal that no local run could reach without Docker.

### Two things left open, deliberately

- `_TARGET` in `test_migration_054_...py:81` is now correctly named but still **unused** — dead since `130bc5d3`. Renamed rather than deleted per the repo's "name unrelated dead code, don't delete it" rule. Say the word and it goes.
- **#1582 conflicts with this branch.** It adds `056_drop_documents_created_at_index` and touches `032`, `041`, `042`, `044`, `049`, `test_migration_bundling.py`, and `test_sqlite_migrations.py` — the same files. That is precisely the collision this ticket exists to prevent, happening live. Whichever lands second should take this branch's dynamic-head form; if #1582 lands first, its head-literal bumps become no-ops here.

msrdic (approved):

PR #1586, opened 2026-08-05
Add a bounded keyset scan primitive to the SQLAlchemy document stores

filippetrovic:
## Summary

Adds an internal `scan_documents` to the two SQLAlchemy-backed relational document stores (PostgreSQL and the embedded sqlite_lance adapter), plus the shared `DocumentScanStep` result type.

One call runs one bounded `SELECT`: a dialect-compiled pushdown fragment AND-composed with the namespace scope, the optional `status` / `updated_before` narrowing, and a keyset resume predicate, ordered by the total order pinned in #1576 and capped by a row bound. The backend resolves its own compiler through the registry keys added in #1577 and executes the step in its own session; no transaction spans steps.

**Nothing calls this yet.** The scan loop and the caller-facing surface are follow-up work, so this lands as an internal primitive with its tests.

Deliberate scope choices:

- The **public storage API is untouched** — `list_documents` is unchanged on both stores.
- `scan_documents` is **not** added to `RelationalBackendProtocol`. That protocol is `@runtime_checkable` and an existing test asserts `isinstance` against it; an abstract member would make that assertion false for the two stores that do not implement it yet. The protocol addition belongs with the change that lands all four.
- The step type is not exported publicly.
- The two remaining relational stores are out of scope.

Two things worth a reviewer's attention:

**The shared statement builder.** Both stores scan the same model, and only the compiled fragment's attachment is genuinely dialect-specific, so the statement is built once in `_documents_scan.py`. That is not only deduplication: the Postgres leg cannot be exercised without a live server, so building the predicate once — in a form the service-free embedded tests can execute — is what lets the locally-runnable leg prove the shared semantics.

**The keyset operands bind through the ORM column types** rather than being formatted by hand, which is load-bearing rather than stylistic. On the embedded store, `created_at` is stored as fixed-width text with a space separator and always-6-digit microseconds. A hand-formatted ISO-8601 cursor puts a `T` (0x54) where the stored form has a space (0x20), so it sorts *above* every row and the cursor compares less than its own row — a resumed walk returns that row forever instead of advancing. A space-separated form that drops the `.000000` sorts *below* its tie-mates and silently skips them. A test pins this: swapping the typed binds for `isoformat()` fails it.

Two smaller invariants that are invisible on a full-pushdown backend and wrong everywhere else:

- `last_scanned` is the key of the final row of the **raw** window, not the last row that survived filtering, so a resumed walk does not re-scan the non-matching gap.
- `exhausted` derives from the **raw** window count against the row bound, not the post-filter count, so it remains a sound termination signal.

The step reports `consumed_keys` verbatim from the compiler and carries no residual AST — no compiler produces a pruned tree, the caller already holds the filter it passed in, and a field named `residual` would invite exactly the mistake the compile-context docstrings warn against (the post-filter must evaluate the full filter unconditionally; `consumed_keys` is a reporting signal, not permission to skip leaves).

## Test plan

- [x] Unit tests pass — 11 new tests. The full CI-shaped unit invocation is green on this branch: `10658 passed, 35 skipped`.
- [x] Embedded (sqlite_lance) suite passes locally with no services: 10 tests.
- [x] **Mutation-verified**: replacing the typed cursor binds with `isoformat()` / `str()` fails 3 tests with exactly the predicted pathology (the cursor returns its own row instead of an empty tail); reverted to green. The guarantee is tested, not merely asserted.
- [x] `make format` clean, `make lint` exit 0, new module type-checks clean.
- [ ] Integration tests — the 11 Postgres tests **skip locally** (no Docker in the dev container) and run for the first time in CI against the real service. This is the one leg not verified locally.

Covered per backend: the row bound is honored; a mid-tie cursor resume over a tie-heavy corpus walks exactly-once in strict total order; the compile split is reported correctly for a forced-residual filter; an empty window reports exhausted; and `last_scanned` is asserted non-vacuously (the raw window deliberately ends on a non-matching row, and the test asserts the key differs from the last match).

Also pinned: the date system keys stay out of the embedded store's pushdown set, per #1580 — that store writes wall clock, so a pushed-down date compare would be wrong. A test asserts they route to the residual.

## Summary by CodeRabbit

* **New Features**
  * Added bounded document scanning for PostgreSQL and SQLite/Lance storage.
  * Supports status and update-time filters, configurable scan limits, and cursor-based continuation.
  * Scans now provide completion status and continuation metadata for reliable traversal.
  * Added filter pushdown support where available.

* **Bug Fixes**
  * Improved handling of timestamp and identifier ordering during resumed scans.

* **Tests**
  * Added comprehensive coverage for pagination, filtering, exhaustion, validation, and cursor resumption.

### DIFF
--- a/src/khora/storage/backends/_documents_scan.py
+++ b/src/khora/storage/backends/_documents_scan.py
@@ -1,0 +1,154 @@
+"""Shared query construction for the bounded documents scan — ``@internal``.
+
+The two SQLAlchemy-backed relational stores (PostgreSQL and the embedded
+sqlite_lance adapter) run the *same* bounded keyset scan over the *same*
+``DocumentModel``. Only one thing genuinely differs between them: how their
+dialect compiler's pushdown fragment attaches to the statement (PostgreSQL's
+compiler emits a SQLAlchemy expression; the SQLite compiler emits a string plus
+positional binds). Everything else — the namespace scope, the optional ``status``
+/ ``updated_before`` narrowing, the keyset predicate, the enumeration order, and
+the row bound — is identical, so it is written once here rather than twice.
+
+That is not only DRY. The keyset predicate is the one part of this scan whose
+correctness depends on a *serialization* detail, and it is the part that cannot
+be exercised without a live PostgreSQL. Building it once, in a form the embedded
+store's own tests can execute, means the locally-runnable leg proves the shared
+semantics and only the fragment attachment is left to a services-backed CI job.
+"""
+
+from __future__ import annotations
+
+from datetime import datetime
+from uuid import UUID
+
+import sqlalchemy as sa
+from sqlalchemy import Select
+
+from khora.db.models import DocumentModel
+from khora.storage.backends.base import DocumentScanKey
+
+__all__ = ["build_documents_scan_query"]
+
+
+def build_documents_scan_query(
+    namespace_id: UUID,
+    *,
+    status: str | None = None,
+    updated_before: datetime | None = None,
+    after: DocumentScanKey | None = None,
+    scan_limit: int = 100,
+) -> Select[tuple[DocumentModel]]:
+    """Build the bounded scan ``SELECT``, minus any compiled filter fragment.
+
+    ``@internal``. The caller ``AND``-s its own dialect's pushdown fragment onto
+    the result with a further ``.where(...)`` — appending a WHERE after
+    ``order_by`` / ``limit`` is well-defined and leaves both in place — and then
+    executes it in its own session.
+
+    ``ORDER BY created_at DESC, id DESC`` is the total enumeration order all four
+    relational stores adopted in khora #1576. ``created_at`` is ``NOT NULL`` as of
+    khora #1583; the keyset predicate below depends on that — a NULL key would
+    make the row-value comparison evaluate to NULL and silently drop the row from
+    every page.
+
+    **The explicit ``literal(value, <column>.type)`` on each cursor operand is
+    load-bearing. Do not simplify it away.** A right-hand ``tuple_`` types each of
+    its elements from the *value*, independently of the column it is compared
+    against — nothing propagates leftwards. ``tuple_(a_datetime, a_uuid)``
+    therefore renders correctly only because those two Python types happen to
+    infer to this table's own column types; the agreement is a coincidence, not a
+    mechanism. Hand it an operand one step off and the coincidence lapses
+    silently: a ``date`` where a ``datetime`` belongs infers ``Date`` and binds
+    ``'2026-01-31'``, which on the embedded store's TEXT column sorts below every
+    ``'2026-01-31 …'`` row and skips the whole day. Naming the column's type
+    forces the correct serialization instead, and turns the ``str``-instead-of-
+    ``UUID`` slip from a silently mis-ordering dashed 36-char bind into an
+    immediate ``AttributeError`` from the type's own processor.
+
+    Formatting the cursor by hand is wrong in two directions and neither is loud:
+    on the embedded store an ISO-8601 ``'T'`` separator sorts *above* the stored
+    space-separated form, so the cursor's own row compares less than itself and a
+    resumed walk returns it forever; while a space-separated form that omits the
+    ``.000000`` microseconds sorts *below* its tie-mates, silently skipping them.
+    Passing a bare ``datetime`` through an untyped textual bind picks the second
+    of those, via a deprecated driver-level adapter that warns and continues.
+
+    **Naming the type is a SQLite-side guard only — it is not what makes a cursor
+    correct.** On PostgreSQL no bind processor runs, so a *well-typed* operand
+    reaches the driver untouched under either spelling — though not a no-op even
+    there, because the spelling picks the cast: measured on the asyncpg dialect, a
+    ``date`` renders ``$1::DATE`` bare against ``$1::TIMESTAMP WITH TIME ZONE``
+    typed, and a ``str`` id renders ``$2::VARCHAR`` bare against ``$2::UUID``.
+    Neither spelling rescues the ``date`` — those two forms resolve midnight in
+    the server's zone and the client host's respectively. A ``str`` id diverges
+    the other way and is the sharpest asymmetry here: the typed spelling raises on
+    this store, because the column type's processor asks the operand for ``.hex``,
+    while asyncpg's uuid encoder takes the ``PyUnicode`` branch and its parser
+    skips ``-`` outright, so a dashed cursor decodes to the identical sixteen
+    bytes and is silently *correct* on PostgreSQL. The same bad input is loud on
+    one store and harmless on the other; neither store's behaviour predicts the
+    other's, which is the whole reason this rule lives here rather than in either
+    backend. That store has the same hazard by a different route, on two
+    different evidence standards. Its ``timestamptz`` encoder converts with
+    ``obj.astimezone(utc)``, so a naive cursor resolves against whatever zone the
+    process happens to run in — that is plain-Python arithmetic, run across four
+    zones, and the shift is the host's UTC offset, hence zero and invisible on a
+    UTC host. That the same encoder also wraps a bare ``date`` at midnight in the
+    host's local zone is **read from its source only**: unlike the uuid parser
+    above, it is a ``cdef`` with no Python entry point, so it cannot be exercised
+    without a live server. The guarantee that holds on both stores is therefore
+    the cheap one: **build a cursor from a row this store returned, never by
+    hand.** What the typed bind buys is that the SQLite path fails loudly, or not
+    at all, instead of quietly reordering — worth having, but not a substitute.
+
+    The cursor must not be *adjusted* on the way in either, and the two dialects
+    make opposite adjustments look harmless. The embedded store holds wall clock
+    with the writer's offset discarded, so attaching or stripping ``tzinfo``
+    there is a no-op while converting to another zone moves the position; a
+    ``timestamptz`` holds an instant, so converting is the no-op and stripping
+    ``tzinfo`` is what moves it — by the host's UTC offset, hence exactly zero on
+    a UTC host and invisible to a test that runs there. Neither store's rule
+    generalizes to the other. The one that holds on both is narrower than either:
+    bind the value you read, unmodified.
+
+    Uniform descending on both keys is what makes a single row-value comparison
+    legal; do not expand it to an ``OR`` form. Row values need SQLite >= 3.15,
+    which this package cannot be installed below: ``requires-python >= 3.13`` and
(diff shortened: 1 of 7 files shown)


filippetrovic:
## Consolidated review

An eight-perspective review was run across correctness, database, tests, security, performance, docs, decision-record drift, and a devil's-advocate pass. **A tooling caveat, stated up front:** the per-reviewer narrative reports did not survive the review session, so rather than reproduce them from memory this comment reports only findings the reviewing lead **re-verified directly against the code and CI in this session**. Every claim below has a citation or a command behind it. Nothing here is inferred from a report I cannot show.

**Overall assessment: APPROVE.** No CRITICAL or HIGH findings survived verification.

---

### Verified correct

**The three load-bearing invariants** (`sqlite_lance/relational.py:568-572`, mirrored in `postgresql.py`):

- `last_scanned=(rows[-1].created_at, rows[-1].id) if rows else None` — the final row of the **raw** window. `rows` is the SQL result, narrowed only by pushdown, never by a post-filter. Correct: a resumed walk will not re-scan the rejected gap.
- `exhausted=len(rows) < scan_limit` — derived from the **raw** window count, not a post-filtered subset. Correct: it stays a sound termination signal.
- `consumed_keys` is reported verbatim from the compiler, with the docstring explicitly stating it is *not* permission to skip leaves in a post-filter.

Each call runs in its own session (`async with self._get_session()`); no transaction spans calls.

**The cursor-bind guarantee is genuinely tested, not merely asserted.** Mutation-verified in this session: replacing the typed binds with `sa.literal(cursor_created_at.isoformat())` / `str(cursor_id)` fails 3 tests, with exactly the documented pathology — the cursor returns its own row instead of an empty tail. Reverted to green (10 passed). The distinction matters: this suite would catch the regression it claims to pin.

**The `?` → named-bind rewrite is sound** (`relational.py:1292-1321`). Two tripwires, both failing at construction rather than execution: a colon guard (`text()` would parse `:` as a bind name) and a placeholder/args count guard (a stray `?` would silently shift every later bind). The no-inline premise holds — `compile_lance` binds the user-derived metadata JSON path via `_jpath_bind` rather than inlining it, so the only text in the fragment is the compiler's own constants. Neither guard can fire today; keeping them as tripwires is the right call.

**Scope is clean.** Verified untouched: `backends/sqlite.py`, `backends/surrealdb/`, `engines/protocol.py`, `khora.py`, `storage/coordinator.py`. `scan_documents` is correctly **not** added to `RelationalBackendProtocol` — it is `@runtime_checkable` and an existing `isinstance` assertion would go false for the two backends that don't implement it yet.

**Public-repo hygiene: clean.** No internal ticket IDs, decision-record numbers, or internal hostnames anywhere in the diff, the new files, or this PR's text.

**No placeholders** — no TODO/FIXME/HACK/NotImplementedError in the diff.

---

### MEDIUM

**M1 (correctness / decision-record) — the superset-filter assumption is load-bearing and only documented.** `relational.py:530-536` is candid that resuming past pushdown-rejected rows skips nothing *only because the pushdown is a superset filter*, and that a compiler emitting a match-all placeholder inside a negation would make this scan drop documents silently. That property lives in the compilers and the compile context, not in this method, and it is currently guaranteed by the split-soundness gates from #1580. **This is correctly identified and honestly documented, not a defect in this PR** — but it is a cross-module invariant with no test in this PR asserting it, and the follow-up that wires up the coordinator should pin it explicitly rather than inherit it as prose.

**M2 (performance) — `scan_limit` bounds rows *returned*, not rows *examined*.** Documented in the builder docstring. A highly selective pushdown fragment can still make one call read an entire namespace. Acceptable for a bounded primitive with no callers, but the caller-facing tier needs to know that per-call work is not actually bounded by `scan_limit` alone.

### LOW

**L1 (docs) — `CLAUDE.md` contains two claims that are false in the current dev container:** "Docker Compose is always available" (docker is not installed) and "coverage ≥65%" (`pyproject.toml` sets `--cov-fail-under=77`). Pre-existing and out of this PR's scope, but they cost real time here — the first sends you looking for a Postgres that cannot exist.

**L2 (docs) — the docstrings are unusually long relative to the code.** Spot-audited and the strong factual claims hold up (stored byte formats, `'T'` 0x54 vs `' '` 0x20 ordering, the dashless 32-char id, the SQLite ≥3.15 row-value floor discharged via `requires-python >= 3.13`). Given that a wrong bind here is an infinite loop rather than a loud error, the prose is judged load-bearing rather than excessive.

---

### Verification status

| Check | Result |
|---|---|
| CI | **25/25 SUCCESS** |
| PostgreSQL integration tests | **Ran and passed in CI** — confirmed in the job log: `test_scan_documents_pg.py ...........` (11 passed, not skipped) |
| Full unit lane (CI-shaped, local) | 10658 passed, 35 skipped, 0 failed |
| Embedded suite (local, no services) | 10 passed |
| Mutation verification | 3 tests fail under the naive bind, then green after revert |
| `make format` / `make lint` | clean / exit 0 |

The one thing worth flagging about *this PR's own* verification story: 11 of its 21 tests could not run locally, so the Postgres leg had no local signal at all. That has now been discharged by CI rather than assumed.

filippetrovic:
## Correction to the consolidated review above

The eight reviewer reports reached me **after** I posted the consolidated review. That review reported only what I had re-verified myself, and said so — but its `APPROVE / no HIGH` verdict was therefore scoped to the checks I ran, and two findings I did not check came back HIGH from multiple independent reviewers, each **mutation-verified rather than argued**. Correcting the record, and both are being fixed on this branch.

### H1 — the compiled SQLite fragment is spliced into the WHERE ungrouped

`sqlite_lance/relational.py:1321`. `TextClause.self_group()` returns `self` for everything except `in_op`, so `.where()` does not parenthesize a `text()` fragment. Executed, not inferred:

```
select(DocumentModel).where(ns == X).where(text("a = 1 OR b = 2"))
->  WHERE documents.namespace_id = ? AND a = 1 OR b = 2
```

`AND` binds tighter than `OR`, so the namespace predicate ends up inside the left disjunct. Against a real two-namespace SQLite, an ungrouped fragment with a top-level `OR` returned **the other namespace's row**; the same fragment wrapped in parens did not.

**Latent, not live.** `compile_lance` self-parenthesizes every boolean node and every OR-emitting leaf (`$ne`, `$nin`), confirmed across 20 fuzzed filter shapes including hostile input — so nothing reachable today produces an ungrouped fragment. It is worth fixing anyway on this PR's own stated standard: `_lance_fragment_to_text` already ships two tripwires whose docstring says they "cannot fire today… kept for the day that stops being true." This is the third, and the only one whose failure mode is a silent cross-tenant read rather than a construction-time `ValueError`. Fix is `text("(" + rewritten + ")")`.

The PostgreSQL leg is unaffected — its compiler emits SQLAlchemy expressions, which `.where()` self-groups.

### H2 — namespace scoping has zero test coverage

Two reviewers independently mutated `build_documents_scan_query` to **delete the namespace predicate entirely**. All 22 tests passed.

The cause is structural rather than an oversight in any one test: the embedded fixture hands each test a private migrated database containing exactly one namespace, so no test ever has another tenant's rows to exclude, and the keyset predicate is namespace-blind on its own. The PG leg would catch it only incidentally, via residue from other tests in the shared CI database — contamination, not an assertion.

This is the invariant the repo already guards in three sibling modules, so it is a departure from local practice rather than an exotic ask. A two-namespace test under a filter closes it, and is the same test that would catch H1.

### Also worth recording

- **`updated_before` silently drops rows with NULL `updated_at`** (found independently by correctness and database review; measured 3000 of 3001 returned, with `exhausted=True`). `updated_at` is physically nullable; `updated_at < :bound` is NULL for such a row, so it is excluded from the raw window — never scanned, never resumed past, absent from the whole walk with no signal. **Inherited verbatim from the existing `list_documents`, so not a regression** — but it matters more here, on a primitive whose contract is exactly-once complete enumeration.
- **Two docstring claims in `_documents_scan.py` are factually wrong** — the untyped-textual-bind claim describes a *safe* construct as the trap (measured: SQLAlchemy infers `DateTime` and emits the correct form), and a "this store / that store" paragraph in a deliberately store-neutral module contradicts itself. Being fixed.
- **`scan_limit` bounds rows returned, not rows examined** — honestly documented in the source, but a deviation from the decision record's "hard per-call work cap" wording, and the measured cliff is ~3 orders of magnitude (0.072 ms unfiltered vs 91 ms for a metadata pushdown matching nothing, on 200k rows with no I/O). The mitigating half, which deserves stating: because `last_scanned` is the last *raw* row, a resumed step never re-examines the rejected gap, so **total** work across a full walk is O(namespace) regardless of selectivity. `scan_limit` fails as a latency bound and succeeds as a total-work bound. The follow-up that wires up a caller should not inherit the record's wording as a latency guarantee.

### What the reviewers confirmed clean

Row-value keyset genuinely seeks the index (`EXPLAIN QUERY PLAN`: `SEARCH ... USING COVERING INDEX ix_documents_namespace_created_at_id (namespace_id=? AND (created_at,id)<(?,?))`), and the OR-expanded form measurably does not — so the "do not expand to an OR form" instruction is verified, not stylistic. Keyset stays flat with depth (0.023 ms vs offset's 5.4 ms at 199k). No SQL injection: every operand routes through the compiler's bind path, verified by audit plus a 20-shape fuzz with hostile input, and all guard failures are loud at construction. No placeholders. Scope clean. Public-repo hygiene clean.

filippetrovic:
## H1/H2 landed — `4a12e4ea`

The correction comment above announced both HIGHs as being fixed on this branch; that commit had not landed. It has now.

**H1** — `_lance_fragment_to_text` wraps the rewritten fragment in parentheses, documented as the splice's third tripwire and the only one whose failure mode is silent (cross-tenant read) rather than a construction-time `ValueError`. Latent today: `compile_lance` self-parenthesizes every reachable shape, so no compiled-filter test can catch the regression — the new test drives the splice directly with an ungrouped top-level `OR` over a two-namespace corpus. Mutation-verified: removing the parens fails it.

**H2** — namespace isolation pinned on both stores: a second namespace seeded with the same varied corpus (every row a guaranteed filter hit), filtered walk at `scan_limit=1` so the scope predicate is load-bearing on every page including resumes. Mutation-verified: deleting the namespace predicate from `build_documents_scan_query` now fails 2 tests (previously 0 of 22).

**Also from the correction's list**: the two measured-wrong docstring claims in `_documents_scan.py` corrected (an untyped `sa.literal(datetime)` infers `DateTime` and is safe — the prose no longer names a safe construct as the trap; the PostgreSQL paragraph's store referents named); the `updated_before` NULL-`updated_at` exclusion documented in the builder Args + both store docstring pointers (behavior unchanged — the facade tier owns the contract call); `scan_limit` reframed as a total-work bound, not a latency bound.

Verification: embedded suite 24/24 (was 22); full `tests/unit/storage/backends/sqlite_lance/` 200 passed; both mutants killed and reverted; `ruff format`/`ruff check` clean; PG module collects 13 (new isolation test runs in CI).

probablyrobot (approved):

PR #1587, opened 2026-08-06
Add a bounded keyset scan primitive to the raw-SQL document stores

filippetrovic:
## Summary

Ports the bounded document-scan step that the two SQLAlchemy-backed stores already ship (#1586) to the two stores that hand-write their SQL: the raw SQLite backend and the SurrealDB relational adapter. Both are `@internal` with no callers yet, so the public storage API is untouched and nothing changes for existing users.

Each store emits a single `SELECT` combining the compiled pushdown fragment, the namespace scope, the optional `status` / `updated_before` narrowing and a keyset condition, ordered by `(created_at DESC, id DESC)` and bounded by `scan_limit`. Step semantics mirror #1586 exactly:

- `last_scanned` is the key of the final row of the **raw** window, not the last match
- `exhausted` means the raw window came back shorter than `scan_limit`
- `consumed_keys` is reported verbatim from the compiler, as a reporting signal only — never permission for a caller to skip leaves
- one step per call, in the store's own connection/session; no transaction spans steps
- `scan_limit` bounds rows **returned**, not examined — a total-work bound, not a latency bound

### Per-store differences that are not incidental

- **raw-SQLite** splices the compiler's positional binds natively and expresses the keyset as a row-value comparison. Cursor timestamps bind through the same serialization the writer uses, so cursor and stored TEXT sort identically across the microsecond-format boundary. The cursor id binds dashed — an undashed cursor sorts *above* its own row and walks the scan backwards forever.
- **SurrealQL has no row-value comparison**, so the keyset is written out as an explicit disjunction. `updated_before` binds the datetime *object* rather than a string: the column is `TYPE datetime`, and a string operand matches no row at all, which would report a populated namespace as exhausted on the first step. This diverges deliberately from `list_documents` on the same adapter, which has that defect; it is tracked separately and deliberately not fixed here.
- **Unsafe metadata path segments** make the SurrealDB compiler raise, which would otherwise escape the scan as an internal error. The scan path maps that one case to the public unsupported-filter error, with the catch scoped so unrelated compile failures still propagate.

### On the parentheses

The compiled fragment is parenthesized at the splice on both stores. Both compilers self-parenthesize every shape reachable today, so this is a **tripwire rather than a fix** — without it an ungrouped top-level `OR` reassociates past the namespace predicate and returns twice the rows, across tenants. Measured at 12 rows where the grouped form returns 6, on both stores.

## Test plan

- [x] Unit tests pass — `tests/unit/test_sqlite_scan_documents.py`, `tests/unit/filter/test_documents_scan_error_marker.py`
- [x] Integration tests pass — `tests/integration/storage/backends/surrealdb/test_relational_scan_documents.py` (executes rather than skips; the SurrealDB extra is installed)
- [x] Full suite green: `make test` → 560 passed, 356 skipped, coverage 85.43% (≥77% gate)
- [x] `make lint` exits clean; `make format` reports no changes
- [x] **Mutation-verified rather than assumed** — 14 mutations run against an isolated worktree, all 14 killed:
  - namespace predicate removed → fails on both stores
  - fragment-splice parentheses removed → fails on both stores
  - keyset disjunction parentheses removed (SurrealDB) → fails
  - cursor id bound undashed → fails
  - cursor timestamp forced to six µs digits, and separately truncated → each fails, and **only** at its own microsecond polarity, which is why the cursor tests are parametrized at both
  - cursor operands bound in the wrong order → fails
  - fragment args spliced at the wrong bind position → fails
  - `last_scanned` taken from the first row / the last match → fails
  - `updated_before` reverted to `.isoformat()` (SurrealDB) → fails
  - compiler's guard message reworded two modules away → fails, via the co-located marker test
- [x] Adjacent regression sweep (sqlite backend, sqlite sort index, surrealdb backend + unit suites, filter suites): 1973 passed

## Summary by CodeRabbit

* **New Features**
  * Added document scanning for SQLite and SurrealDB storage.
  * Supports bounded results, cursor-based pagination, status filtering, update-time limits, and compiled filters.
  * Provides resume information to continue scans reliably and detect when results are exhausted.
  * Maintains namespace isolation and consistent ordering across pages.
* **Bug Fixes**
  * Improved handling of unsupported or unsafe filter expressions with clear storage-specific errors.
  * Prevented filtering, cursor, and namespace conditions from affecting unrelated documents.

### DIFF
--- a/src/khora/storage/backends/sqlite.py
+++ b/src/khora/storage/backends/sqlite.py
@@ -24,7 +24,7 @@ from khora.core.models.entity import Entity
 from khora.core.models.recall import DocumentProjection
 from khora.core.models.tenancy import TenancyMode
 from khora.storage.backends._fts5 import escape_fts5_query
-from khora.storage.backends.base import PaginatedResult
+from khora.storage.backends.base import DocumentScanKey, DocumentScanStep, PaginatedResult
 
 if TYPE_CHECKING:
     from khora.filter.ast import FilterNode
@@ -506,6 +506,144 @@ class SQLiteRelationalBackend:
         )
         rows = await cursor.fetchall()
         return [self._row_to_document(r) for r in rows]
+
+    async def scan_documents(
+        self,
+        namespace_id: UUID,
+        *,
+        filter_ast: FilterNode | None = None,
+        status: str | None = None,
+        updated_before: datetime | None = None,
+        after: DocumentScanKey | None = None,
+        scan_limit: int = 100,
+    ) -> DocumentScanStep:
+        """Scan one bounded keyset window of a namespace's documents.
+
+        ``@internal``. Not part of the public storage API — the offset-based
+        :meth:`list_documents` is unchanged. One ``SELECT`` on this backend's own
+        connection; no transaction spans steps and no consistent snapshot is
+        claimed. A caller walks the namespace by chaining
+        :attr:`DocumentScanStep.last_scanned` back in as ``after`` until the step
+        reports ``exhausted``.
+
+        The same scan the two SQLAlchemy stores run (khora #1586), written out
+        here rather than shared: this store has no ORM, builds SQL text and a
+        positional bind list by hand, and owns its own ``documents`` DDL. Binds
+        being positional, conjunct order *is* bind order — namespace, ``status``,
+        ``updated_before``, the two cursor operands, the fragment's ``args``, the
+        row bound — and appending out of order shifts every later bind by one.
+
+        **The compiled fragment is spliced in parentheses, a cross-tenant
+        tripwire.** Measured, two namespaces of 6 rows each: grouped returns 6
+        rows, ungrouped returns 12, no error and no warning. Latent today
+        (``compile_lance`` self-parenthesizes every shape it emits), which is why
+        it is enforced rather than assumed. Nothing else from the sqlite_lance
+        splice is ported: that store's ``:`` and ``?``-count guards protect a
+        rewrite into SQLAlchemy named binds, and this driver is natively qmark —
+        ``:`` is not bind syntax here, and a count mismatch is already a loud
+        ``sqlite3.ProgrammingError``. No ``CompileError`` mapping is wrapped
+        around the compile call either, unlike the SurrealDB store; that
+        asymmetry is genuine, since ``compile_lance`` has no ``raise
+        CompileError`` site and raises :class:`RecallFilterUnsupportedError` only
+        under ``on_unsupported="raise"``, which this context never selects.
+
+        **Cursor serialization inverts from sqlite_lance.** ``created_at`` binds
+        through :func:`_dt_to_str` (``datetime.isoformat()``), byte-for-byte what
+        :meth:`create_document` wrote into a TEXT column SQLite compares
+        lexicographically; ``id`` binds as ``str(uuid)``, the **dashed** 36
+        characters that same writer stored. sqlite_lance holds 32 *undashed* hex
+        characters, so a dashed cursor is the bug there and the only correct form
+        here. An undashed cursor sorts *above* every stored id it shares leading
+        hex with — the stored form carries ``-`` (0x2D) at index 8 where the
+        undashed form carries a hex digit — so ``id < :cursor`` matches strictly
+        more rows, including the cursor's own. Measured from mid-tie ``…0003``:
+        dashed yields ``[…0002, …0001, …0005]``, ``.hex`` yields ``[…0004, …0003,
+        …0002, …0001, …0005]``, so the walk runs backwards and never terminates.
+        Positions are store-local: build one from a row this store returned.
+
+        This store therefore enumerates in **stored-text order — a deterministic
+        total order, but not an instant order**, since :func:`_dt_to_str`
+        preserves the writer's offset with no UTC coercion. Pre-existing, and the
+        same property that keeps the date keys out of ``_PUSHABLE_SYSTEM_KEYS``.
+        It does not threaten the walk: keyset correctness needs the cursor
+        predicate to agree with ``ORDER BY``, not with chronology, and both are
+        lexicographic over the same column. Callers must not read this order as
+        chronological.
+
+        ``created_at`` cannot be NULL, belt and braces: ``_SCHEMA_SQL`` declares
+        it ``NOT NULL`` independently of and predating the Alembic chain this
+        store does not have, *and* :meth:`create_document` is the only INSERT and
+        coalesces a missing timestamp to now. A NULL key would make the row-value
+        comparison evaluate to NULL and drop the row from every page.
+        ``updated_before`` narrows with ``updated_at < ?``, the same shape #1586
+        ships; the NULL-``updated_at`` exclusion that tier documents does not
+        arise here, because ``updated_at`` is ``TEXT NOT NULL``.
+
+        ``scan_limit`` bounds rows **returned**, not rows **examined** — a
+        selective fragment can still make one call read the whole namespace, so
+        it is a total-work bound, never a per-call latency bound. What it buys is
+        that a full walk stays O(namespace): ``last_scanned`` is the last *raw*
+        row, so a resumed step never re-examines the rejected gap. Resuming past
+        those rows skips nothing **only because the pushdown is a superset
+        filter** — a property of the compiler and this store's compile context,
+        not of this method.
+
+        ``consumed_keys`` is reported verbatim from the compiler: the leaves the
+        SQL already enforced. It is a reporting signal and **never** permission to
+        skip them — the post-filter must evaluate the whole AST unconditionally.
+        The date-valued system keys are deliberately unpushable here (see
+        :func:`_documents_compile_context`), so they stay out of it by design.
+
+        Raises:
+            ValueError: ``scan_limit`` below 1, matching the shared SQLAlchemy
+                builder. Raised before anything is compiled or executed, so a bad
+                bound is never masked by a compile error.
+        """
+        if scan_limit < 1:
+            raise ValueError(f"scan_limit must be >= 1, got {scan_limit}")
+
+        conditions = ["namespace_id = ?"]
+        params: list[Any] = [str(namespace_id)]
+        if status:
+            conditions.append("status = ?")
+            params.append(status)
+        if updated_before is not None:
+            conditions.append("updated_at < ?")
+            params.append(updated_before.isoformat())
+        if after is not None:
+            cursor_created_at, cursor_id = after
+            conditions.append("(created_at, id) < (?, ?)")
+            params.extend([_dt_to_str(cursor_created_at), str(cursor_id)])
+
+        consumed_keys: frozenset[str] = frozenset()
+        if filter_ast is not None:
+            compiler = CompilerRegistry.get("relational.sqlite", "documents")
+            compiled = compiler(filter_ast, _documents_compile_context())
+            consumed_keys = compiled.consumed_keys
+            # Parenthesized — see the docstring. Ungrouped, a top-level OR in the
+            # fragment absorbs the namespace predicate and leaks other tenants.
+            conditions.append(f"({compiled.predicate})")
+            params.extend(compiled.params["args"])
+
+        where = " AND ".join(conditions)
+        params.append(scan_limit)
+        cursor = await self._conn.execute(
+            f"SELECT * FROM documents WHERE {where} ORDER BY created_at DESC, id DESC LIMIT ?",  # noqa: S608
+            params,
+        )
+        rows = await cursor.fetchall()
+
+        # ``last_scanned`` and ``exhausted`` both describe the RAW window: the
+        # final row scanned, and whether SQL ran out of rows filling it. Neither
+        # may be derived from a post-filtered subset — that would re-scan the
+        # rejected gap on resume, and would call a full window exhausted.
+        documents = [self._row_to_document(r) for r in rows]
+        return DocumentScanStep(
+            documents=documents,
+            last_scanned=(documents[-1].created_at, documents[-1].id) if documents else None,
+            exhausted=len(rows) < scan_limit,
+            consumed_keys=consumed_keys,
+        )
 
     async def claim_orphaned_documents(
         self,
--- a/src/khora/storage/backends/sqlite.py
+++ b/src/khora/storage/backends/sqlite.py
@@ -539,9 +539,15 @@ class SQLiteRelationalBackend:
         (``compile_lance`` self-parenthesizes every shape it emits), which is why
         it is enforced rather than assumed. Nothing else from the sqlite_lance
         splice is ported: that store's ``:`` and ``?``-count guards protect a
-        rewrite into SQLAlchemy named binds, and this driver is natively qmark —
-        ``:`` is not bind syntax here, and a count mismatch is already a loud
-        ``sqlite3.ProgrammingError``. No ``CompileError`` mapping is wrapped
+        rewrite into SQLAlchemy named binds, and this driver is natively qmark, so
+        a count mismatch is already a loud ``sqlite3.ProgrammingError`` in both
+        directions. **The count check is the whole of the protection — do not read
+        this as "``:`` is inert here".** ``sqlite3`` does treat ``:name`` as a
+        parameter and binds it positionally from a sequence, so a stray ``:`` in a
+        spliced fragment shifts the binds exactly as it would under
+        ``text()``; it merely cannot do so *silently*, because the count stops
+        agreeing. Today it warns (``DeprecationWarning``) and from Python 3.14 it
+        raises ``sqlite3.ProgrammingError`` outright. No ``CompileError`` mapping is wrapped
         around the compile call either, unlike the SurrealDB store; that
         asymmetry is genuine, since ``compile_lance`` has no ``raise
         CompileError`` site and raises :class:`RecallFilterUnsupportedError` only
--- a/src/khora/storage/backends/surrealdb/relational.py
+++ b/src/khora/storage/backends/surrealdb/relational.py
@@ -426,10 +426,19 @@ class SurrealDBRelationalAdapter:
         single comparison**, and there are two such sites at different
         severities. This one is **live**: we write the ``OR`` rather than
         inheriting it from a self-parenthesizing compiler, and it is reached on
-        every resumed step. Ungrouped, ``AND`` binds tighter, so the namespace /
-        status / filter conjuncts all land inside the left disjunct and the right
-        one returns every tenant's rows tied on the cursor instant — measured, 3
-        rows grouped against 7 ungrouped, 4 of them foreign. The compiled
+        every resumed step. Ungrouped, ``AND`` binds tighter, so the disjunction
+        splits the conjunct list at its own position: everything appended *before*
+        it (namespace, status, ``updated_before``) lands inside the left disjunct,
+        and the right one carries only the tie clause plus whatever is appended
+        *after* — today just the compiled fragment. Either way the right disjunct
+        has no namespace scope, so it returns other tenants' rows tied on the
+        cursor instant. Measured with no ``filter_ast``, where the right disjunct
+        is bare: 3 rows grouped against 7 ungrouped, 4 of them foreign. With a
+        fragment the leak is narrowed by that fragment rather than removed, so
+        the magnitude drops but the cross-tenant read does not. **The order of
+        the appends is therefore load-bearing to this paragraph, not to the
+        defect** — regrouping the list changes which conjuncts leak, never
+        whether they do. The compiled
         fragment splice is the second site, a tripwire rather than a live hazard
         (``compile_surrealdb`` self-groups every boolean node today) with an
         identical failure mode, measured at 6 rows becoming 12. The inner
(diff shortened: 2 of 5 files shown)


filippetrovic:
# Consolidated review — 9-reviewer team + security audit

Reviewers: security, performance, correctness, tests, database, docs, spec-drift, observability, devil's advocate. Findings are deduplicated; severity is the synthesizer's, not each reviewer's. **Where a finding says *measured*, a reviewer ran it** — most of this review is execution results, not reading.

**Overall: request changes on one finding; everything else is optional or a follow-up.** The keyset machinery is correct — boundaries, ordering, bind composition and type round-trips held under every probe, on both stores. Test power is unusually high and was independently confirmed: multiple reviewers rebuilt the branch in throwaway worktrees and reproduced the docstrings' mutation claims exactly, including the counterintuitive ones. Against this PR's own premise — #1586's suite stayed green with the namespace predicate deleted — **that hole is closed on both stores, verified by deleting it.**

---

## HIGH

**H1 — [Performance / Database] `surrealdb/relational.py:503` — the keyset disjunction defeats index selection entirely; every resumed step is a full cross-tenant table scan, and the docstring claims the opposite.**

Found independently by three reviewers, each with their own `EXPLAIN` run, all agreeing:

```
namespace_id = $ns                                      -> Iterate Index  (ns-scoped)
namespace_id = $ns AND created_at < $ts                  -> Iterate Index  (prefix + range)
namespace_id = $ns AND (created_at < $ts OR (…))   SHIPPED -> Iterate Table  (whole table, all tenants)
```

Any top-level `OR` collapses the planner to `Iterate Table` — losing not just the `created_at` range but the `namespace_id` prefix. So each resumed step reads and deserializes every `document` row in the database, all tenants included, then filters and sorts in memory. The sort index added one commit earlier for exactly this `ORDER BY` is reachable (row 2 proves it); this method is the one shape that cannot use it.

Measured magnitude: a resumed step returning 50 rows from a 200-doc namespace goes **4.9 ms → 65.4 ms** when 5,000 foreign rows exist (13×, entirely from rows the caller can never see); **531.8 ms/step** at 20 KB bodies, because a table scan materializes full record values, so cost tracks *total corpus bytes*. A full walk is quadratic — **3.50× wall time for a 2× namespace**, with per-step time barely falling as the walk progresses, the signature of re-reading rather than seeking.

Two rescues were tried and **do not work**, so nobody repeats them: the `WITH INDEX` hint still plans `Iterate Table`, and a redundant `created_at <= $ts` guard alongside the OR also still plans `Iterate Table`.

Consequences beyond speed: `relational.py:251-255` asserts *"a full walk is O(namespace) whatever the selectivity."* On this store it is O(namespace × steps) plus a cross-tenant term. It is also **strictly worse than the offset `list_documents` it improves on** — that one keeps a namespace-scoped index plan. And a compiled `$or` fragment degrades even the *first* step, since it carries its own top-level OR into the same WHERE.

No correctness impact — the WHERE still filters, no rows leak — and the method is `@internal` with no callers, so nothing regresses today. But this is the one property that distinguishes a keyset scan from the offset scan 100 lines above it, and neither test file contains an `EXPLAIN` assertion.

*Fix that measured clean:* split the step into two index-eligible queries (`created_at = $ts AND id < $id`, then `created_at < $ts`) and merge client-side, capped at `scan_limit`. Honest caveat from the reviewer who measured it: this does **not** make a step `O(scan_limit)` — 2.x still sorts in memory — it restores namespace scoping, i.e. O(one tenant) instead of O(whole database). *Minimum acceptable:* keep the code, delete the O(namespace) claim, and add one `EXPLAIN`-asserting test so the plan is pinned rather than assumed.

Raw-SQLite is clean by contrast and measured so: every emittable shape plans `SEARCH documents USING INDEX idx_docs_ns_created_id (namespace_id=? AND (created_at,id)<(?,?))`, no temp b-tree, and adding 20,000 foreign rows changed a walk from 18 ms to 17 ms. The row-value form does exactly what it claims. The asymmetry is entirely SurrealDB's.

---

## MEDIUM

**M1 — [Security] `surrealdb/relational.py:530` — unguarded bind-dict merge can silently replace the namespace scope.** `params.update(compiled.params)` lets compiled binds win, so a compiled bind named `ns` replaces the tenant scope in `namespace_id = $ns` and the query still executes normally. **Reproduced:** a compiler returning `params={"f_0": …, "ns": str(other_ns)}` made a scan of namespace A return a row belonging to namespace B — no error, no log. Not reachable at head (`param_namespace` defaults to `"f"`), but it is a default, not a structural guarantee, and the knob is exercised in existing tests. The finding is the **asymmetry**: this PR chose to *enforce* rather than assume the identical latent cross-tenant hazard one line earlier (the fragment parentheses, "a tripwire rather than a fix"), while its sibling got a code comment. Two lines before the update: `collisions = params.keys() & compiled.params.keys()` → raise. Raw-SQLite has no counterpart; its binds are positional.

**M2 — [Correctness / Database] `sqlite.py:643`, `surrealdb/relational.py:548` — `last_scanned` is derived from the converted `Document`, not the raw row; if the converter's fallback fires the walk never terminates.** Both `_row_to_document` carry `_parse_dt(row["created_at"]) or datetime.now(UTC)`. #1586's siblings read the raw row and cannot invent a value. A fabricated `now()` cursor sorts above every row, so the next step re-matches the whole window — **an unbounded loop, not a dropped row. Measured:** a row with `created_at = ''` (which satisfies `TEXT NOT NULL`, and `_parse_dt` maps to `None`) produced an identical 3-row window for 8 straight steps, `exhausted=False` throughout, 24 yielded / 3 unique. Unreachable through either store's write API, so latent. What makes it worth raising: the docstrings' safety argument covers *SQL evaluating a NULL key* and not *the Python converter firing on a non-NULL value that satisfies the constraint* — and the uncovered case fails worse than the one defended against. One line per store. Note this derivation has **already diverged in form** from the two siblings (`documents[-1]` vs `rows[-1]`) — equivalent today, which is what drift looks like on day zero.

**M3 — [Correctness / Docs / Database] `sqlite.py:611` — `updated_before` compares wall clock, not instant, and the two stores in this PR now disagree on it.** The bound binds as `.isoformat()` against a TEXT column compared lexicographically, while the writer stores the caller's offset with no UTC coercion. **Measured**, three rows whose instants are all strictly before the bound: raw-sqlite returns 2, SurrealDB returns 3 (correct). The dropped row is 90 minutes before the bound. Opposite direction too: a naive `updated_at` exactly *equal* to the bound is wrongly *included*, because it is a prefix of the bind. This is the same defect the file's own compile-context docstring spends four paragraphs on when withholding the date keys from pushdown — but `updated_before` is a direct parameter, so **no post-filter can recover the rows**. Inherited from `list_documents` (measured identical), so not a regression; raised because the SurrealDB sibling in this same PR deliberately fixed exactly this and documented the divergence at length, while the SQLite docstring asserts sameness.

**M4 — [Tests] `sqlite.py:631` — the `id DESC` tie-break is pinned by no test.** Mutating `ORDER BY created_at DESC, id DESC` → `ORDER BY created_at DESC` leaves **all 62 tests green**, including the one whose docstring claims "a wrong-but-complete order has to fail too". **Measured mechanism:** the store ships `idx_docs_ns_created_id`, and SQLite walks it backwards supplying `id DESC` as residual order for free. Drop the index on the same fixture and the two forms diverge immediately. So the seed is correctly designed — the *planner* masks the mutant. This matters because the cursor is `(created_at, id) < (?, ?)`, sound only if ORDER BY yields the same total order; anything that steers the planner off that index desynchronises sort from cursor and silently skips or repeats rows inside every tie block, with the suite green throughout. The SurrealDB sibling does **not** have this gap (same mutant kills 2 there). Fix, shape verified: one test that `DROP INDEX`es before scanning and asserts exact order.

**M5 — [Tests] 4 of 22 superset parametrizations are structurally vacuous.** Instrumented oracle/window sizes on both stores: `pushable_exists` (`source_url $exists False`) = oracle 0 / window 0, and `unpushable_key` (`occurred_at $gte`) = oracle 0 / window 6. `$exists: False` on a system key is constant-false (system keys are documented always-present) and `occurred_at` has no documents backing, so `oracle <= window` can never fail — for *any* seed, not fixable by a different corpus. `pushable_exists` is blind precisely where it looks strongest: one of only two shapes where the pushdown consumes a key *and* returns an empty window, the "pushdown rejected everything" direction the test exists to catch. The docstring sells eleven shapes; it is nine. Both replacements measured non-vacuous: `metadata.tier $exists False` → 4/4, or `source_url $exists True` → 6/6. Also add an `assert oracle` anti-vacuity guard, or a future edit to the seed hollows out more parametrizations silently.

**M6 — [Docs] `surrealdb/relational.py:429-431` — inverted mechanism.** Claims the "namespace / status / **filter** conjuncts all land inside the left disjunct". `conditions` appends ns → status → updated_before → keyset → **fragment last**, so ungrouping parses as `(ns AND … AND created_at < $ts) OR ((created_at = $ts AND id < $id) AND (frag))` — the filter conjunct lands in the **right** disjunct, and that disjunct is filter-narrowed rather than bare. The conclusion (unscoped right disjunct → cross-tenant leak) holds; the mechanism does not, and the quoted 3-vs-7-rows measurement is exact only in the no-`filter_ast` case. This is the paragraph someone reorders `conditions` against.

**M7 — [Docs] `sqlite.py:543` — "`:` is not bind syntax here" is false for this driver.** **Executed:** `con.execute("SELECT ?, :foo", [1, 2])` returns `(1, 2)` — a `:name` placeholder *is* counted and bound positionally, with `DeprecationWarning: … Starting with Python 3.14 an sqlite3.ProgrammingError will be raised.` So a stray `:` shifts binds exactly as in the sibling store; what protects this store is the **count check**, not absent bind syntax. The sibling half of the sentence is correct (count mismatch raises, both directions, verified). This sentence is the stated justification for omitting a guard the sibling keeps, and it becomes actively wrong on 3.14.

**M8 — [Correctness] the two stores disagree on legal-but-hyphenated metadata keys.** Same filter, same corpus: raw-sqlite returns 3 rows for `metadata.foo-bar $eq`; SurrealDB raises `RecallFilterUnsupportedError`. Both agree (no raise) when the hyphen sits inside a deferred `$or`. The docstring discloses the sibling-dependence *within* SurrealDB but not that the other store handles the key fine — so a backend-agnostic scan loop gets rows on one store and a hard failure on the other for legal, common JSON. The asymmetry is genuine (`compile_lance` has no `raise CompileError` site at all), but the prose implies a universal capability gap.

**M9 — [Spec drift] `base.py:47-62` — the shared contract was not updated for the two new stores.** The `DocumentScanKey` comment says "the two backends do not agree on its shape" and that the key "is read off a document row and bound back **through the same column type**". After this PR there are four stores and three key shapes, and neither new store binds through a column type. Raw-sqlite is a third shape the taxonomy doesn't cover — aware *or* naive depending on what the writer passed. `base.py` is the only cross-store contract doc for this primitive, so a future coordinator author applying its aware-on-PG / naive-on-embedded rule to raw-sqlite gets it wrong. Same gap on the exception surface: the method now raises a store-specific error on exactly one of four stores and nothing shared records it.

**M10 — [Database] heterogeneous record-id types make the SurrealDB walk non-terminating, silently.** SurrealDB orders `Id::String` and `Id::Uuid` as distinct variants, so once both exist in one table `id < $rid` stops agreeing with `ORDER BY id DESC`. **Measured:** 4 of 8 cursor positions returned the wrong row set; a walk at `scan_limit=1` visited 5 distinct rows in **60 steps without terminating**, no error. khora's own writes are safe (always `RecordID(table, UUID)`), so latent — but SurrealDB is the unified backend users write to directly, and the failure is a hang rather than an exception. Worth one docstring precondition line; this docstring is otherwise scrupulous about stating its preconditions.

**M11 — [Devil's advocate] ~120 lines of test helper duplicated at 88–100% similarity, while the designated shared home already exists.** `_filter_ast` 100%, `_write` 98%, `_seed_varied` 98%, `_SUPERSET_SHAPES` 98%, `_seed` 96%, `_make_namespace` 90%, `_seed_from_ladder` 88% — and both files already import from `tests/test_helpers/document_scan.py`, then define near-copies alongside. `_seed_from_ladder` is the shared `scan_seed` with caller-supplied ids: a two-line `ids=None` parameter, copied as 28 lines twice instead. Only `_two_namespace_ladders` genuinely diverges, and there it is the *docstrings* that differ (one store needs the id ordering load-bearing, the other doesn't) — which argues for one function with a store-specific note. Notably the reviewer who raised this **flagged their own fingerprints**: they pushed the "no shared abstraction" line hard and early, aimed at the *implementation*, and it appears to have been generalised to the tests where a shared module already existed.

**M12 — [Devil's advocate] the dismissed `CompileError` subclass is the better design, and would delete ~110 lines.** The inline defense is that the guard "has no error subclass to key on, and adding one means editing a compiler this scan does not own". But `CompileError`'s own docstring says it signals *"an internal compiler fault … a bug, not a capability gap"* — an unsafe metadata segment is caller input, so **the guard is raising the wrong class already**. A 2-line `class UnsafeMetadataPathError(CompileError)` is purely additive (every existing `except CompileError` still catches it) and removes the marker constant, the 20-line defense comment, the 89-line marker test module, and the fragility it monitors.

**M13 — [Devil's advocate] the duplication most likely to bite is *intra-file*, and this PR created it.** The `namespace_id` / `status` / `updated_before` condition-building block is now verbatim twice per store, ~110 lines apart (`sqlite.py:484-499` vs `605-616`; same on SurrealDB). #1586's incident was *a deleted namespace predicate tests didn't catch*; the response here writes that predicate in two more places and buys test coverage to watch them. A private `_documents_where(...)` per file would state each store's namespace scope exactly once and let `list_documents`' existing tests cover the shared half.

**M14 — [Devil's advocate / spec drift] `scan_documents` is on no Protocol, and #1586's stated reason for that is now discharged.** #1586 kept it off `RelationalBackendProtocol` because two stores didn't implement it yet and an abstract member would break `isinstance` for them. All four now comply. Leaving the public protocol untouched *in this PR* is correct, but the follow-up is unblocked and unrecorded, and four identical signatures are currently type-checked against nothing.

---

## LOW

- **[Observability / Correctness]** `relational.py:524` — `RecallFilterUnsupportedError("metadata", …)` hardcodes `path`; all six other construction sites in the repo pass the real dotted path, and the class stores them separately so callers can report structurally. Not fixable at this call site (the compiler raises a bare `CompileError`) — pairs with M12.
- **[Docs]** `sqlite.py:559-561` — "the walk runs backwards and never terminates" overstates it. The byte mechanism is *correct* (verified: `-` = 0x2D sorts below every hex digit, so an undashed cursor sorts above and `id <` re-matches its own row). But `exhausted = len(rows) < scan_limit` means the quoted example (5 rows at `scan_limit=10`) *terminates* on that step. Hanging needs a full window to repeat.
- **[Docs]** `test_sqlite_scan_documents.py:43-45` — "with `uuid4` ids the whole defect is invisible" is wrong. The cursor's own row shares *all* leading hex by construction, so `str(u) < u.hex` holds for any seed; only the rows-above-the-cursor leg disappears. The own-row re-match — the half that stalls a production walk — survives. The argument for keeping the ladder seed survives the correction.
- **[Docs]** `relational.py:443-444`, `485-486` — a permanent docstring in a public repo citing "the team lead's written authorisation" means nothing to a future reader; the adjacent measurement and a tracking-issue pointer already carry the argument. (Flagged independently by three reviewers.)
- **[Docs]** `relational.py:505` — hardcoded `compilers/surrealdb.py:260` reference rots on the first edit above that line, and the marker test doesn't guard the number.
- **[Docs]** the SurrealDB docstring never mentions that its tiebreak rests on `id < $rid` agreeing with `ORDER BY id DESC` — an empirically-established engine property with a single named regression guard. It was the highest-risk unknown going in; a reader of the method alone won't learn it.
- **[Devil's advocate]** `sqlite.py:612` vs `:616` — two spellings of one serialization contract four lines apart (`updated_before.isoformat()` vs `_dt_to_str(cursor)`). Identical today; the docstring spends fifteen lines on why the cursor must route through `_dt_to_str`, and the adjacent bind bypasses it. Route both through the helper.
- **[Tests]** `test_sqlite_scan_documents.py:758`, `test_relational_scan_documents.py:635` — `assert not isinstance(doc_id, bool)` immediately after `assert isinstance(doc_id, UUID)` cannot fail; its own comment says why.
- **[Tests]** two docstring counts are understated in the safe direction: the raw-row mutant fails 6 not 5 (`:622`), and the keyset-parens mutant fails 4 not the 3 named. Worth correcting in modules this rigorous about not overclaiming elsewhere.
- **[Tests]** `test_sqlite_scan_documents.py` has no `pytestmark`; both sibling modules declare one. Zero impact today (both lanes select by path) — consistency only, with `--strict-markers` on.
- **[Devil's advocate]** the *compiled-fragment* paren tripwire monkeypatches `CompilerRegistry` to install a fake ungrouped compiler, so it cannot fire on the realistic regression — a real compiler starting to emit ungrouped output would leave it green. Keep the parentheses (free, precedented) but note the invariant they lean on has no test where the compilers live. The *keyset-disjunction* tripwire is genuine: real code path, no monkeypatching, hand-written OR reached on every resumed step.
- **[Performance]** the filter AST is recompiled every step (measured 47.5 µs, ~5% of a 100-row SQLite step). Not worth changing inside the method; a free win if a walk helper lands later.
- **[Performance / Info]** `scan_limit` bounds rows, not bytes — both stores `SELECT *`, so `scan_limit=100` over 1 MB documents is ~100 MB resident. Inherited from the #1586 contract; whatever walk helper lands should size by expected row width.
- **[Observability]** `DocumentScanStep` carries no observability dict, so a future step that needs to *degrade* rather than raise has nowhere contract-shaped to record it. #1586's dataclass, untouched here — on the record before the first caller lands.
- **[Observability]** `connection.py:227` returns `[]` for an unexpected driver result shape with no log or record. Previously benign; now `exhausted = len(rows) < scan_limit` derives the walk's only termination signal from that same length, so a populated namespace could report exhausted on step one. Rated unlikely (upstream re-raises driver errors); flagged because this method enlarges the blast radius.
- **[Database]** sub-microsecond `created_at` silently skips rows — SurrealDB stores nanoseconds, the SDK truncates to microseconds, so a tie-mate older than the truncated cursor matches neither disjunct and is dropped permanently. Unreachable through khora's write paths today, but `DEFAULT time::now()` in the schema produces ns precision the moment a write path stops setting the field.

---

## Security audit

Ran the 16-category sweep scoped to added/modified lines. **No dependency manifest or lockfile changes.** No SSRF sinks, no deserialization, no file/path handling, no regex construction, no secrets, no auth or access-control surface — this diff adds no entry point and no trust boundary. The only injection-relevant sinks are the two f-string `SELECT`s, and injection was **cleared by execution**, not by reading:

- Every operand binds. A `"x' OR 1=1 --"` payload as a system-key value, a metadata value, and an `$in` element never reaches predicate text through either compiler using the stores' real compile contexts.
- Metadata path segments cannot escape. Raw-SQLite **binds** the JSONPath as a parameter, so user key text never enters SQL at all (4 hostile keys executed, 0 rows, no error). SurrealDB must interpolate, and its `_SAFE_SEGMENT_RE` blocked **15 of 15** hostile segments; the `\Z`-rather-than-`$` anchor is load-bearing and correct (`"tier\n"` is rejected, which `$` would have passed).
- Cross-tenant reachability: 8 adversarial filter shapes plus cursor-resume-with-`$or` against a real two-namespace store — **0 foreign rows in every case.** The parentheses are genuine defense-in-depth today: enumerating both compilers' top-level emissions, no reachable fragment is ungrouped.
- A third party cannot shadow the documents compiler — `CompilerRegistry.register` raises on a conflicting function for an occupied key.

**Verdict: PASS.** One MEDIUM (M1), report-only per the audit's own policy — query-construction changes are never auto-fixed.

---

## Adjacent, pre-existing, out of scope for this PR

Raised so they are on the record; all three have or want their own issue, and **not fixing them here is correct scope discipline**:

1. **SurrealDB `list_documents(updated_before=…)` returns `[]` unconditionally.** Measured: no bound 6 rows, datetime bind 6, string bind **0**. It binds `.isoformat()` against a `TYPE datetime` field and nothing in the repo passes that parameter to that store, which is why it survived. Already filed. This is *why* `scan_documents` diverges.
2. **That same `list_documents` has no namespace-isolation regression test** — a reviewer neutralized its `namespace_id` predicate and the whole SurrealDB integration directory still passed (78). The new `scan_documents` is now the better-covered of the two enumeration methods on that adapter.
3. **`CLAUDE.md` documents `make format` as "black, isort, ruff"**; the target is `ruff format` + `ruff check --fix` and there is no black or isort in the toolchain. Pre-existing drift that cost one reviewer a false alarm.

---

## Bottom line

**H1 is the only finding I would gate on**, and it is narrow: either restructure the SurrealDB step to keep the index, or delete the false complexity claim and pin the plan with a test. Everything in MEDIUM is a real improvement and none of it is a correctness risk *today* — the method has no callers, which is exactly why now is the cheap moment to take M1, M2 and M4.

Worth stating plainly, because a long findings list obscures it: **the engineering here is strong.** Two implementers refused instructions from their own ticket that would have put false claims in shipped docstrings, and verified the refusals by measurement. Two QA agents pushed back on reviewer instructions to delete tests, with measurements, and were right both times — the reviewers conceded and corrected their own specs. Every mutation claim in both test modules was independently reproduced. That is the culture this PR's premise calls for.

sale87 (approved):

Issue #1464, opened 2026-07-09
Crash mid-window permanently half-ingests documents - checksum dedup then skips them on re-run


probablyrobot:
## Problem (DA CONFIRMED - correctness bug)

A crash mid-window in `_remember_batch_impl` permanently half-ingests documents, and re-runs skip them:

- The window loop (engine.py:3378) has no try/except and no `mark_failed`; `mark_completed` only fires at window end (:3830).
- Chunks commit at stage 3; if the process dies before stage 6, entities/relationships are never written and the document stays PENDING.
- `get_documents_by_checksums` filters `status != FAILED` (storage/backends/postgresql.py:797), so the PENDING half-ingested doc is returned as a dedup hit and SKIPPED on re-ingest. The docstring shows the intent was FAILED-only re-ingest; PENDING was overlooked.

Result: un-repairable state - chunks exist, graph does not, and no retry path exists short of manual DB surgery.

## Fix

Flip the dedup filter to `status == COMPLETED` (+ regression test simulating a mid-window crash and asserting the re-run re-processes the doc). Optionally mark docs FAILED from a window-level exception handler for observability.

Effort: S. Wave A.

PR #1588, opened 2026-08-06
Bind datetimes as objects in the SurrealDB relational adapter

filippetrovic:
## Summary

Four sites in `storage/backends/surrealdb/relational.py` bound a `datetime` as an ISO-8601 string via `.isoformat()`. The timestamp columns are `TYPE datetime`, and SurrealDB compares across types **without reaching the values**: `datetime < string` is always false and `datetime >= string` always true, whatever the string contains (verified with a past date, a future date, one exactly equal, the non-timestamp `'zzz'`, and the empty string).

A string bind therefore does not *misorder* a predicate — it **pins** it, silently, with no error. That one fact explains why the sites failed in opposite directions. Each now binds the `datetime` object, matching the adapter's own write path and `scan_documents`, which was already correct and is unchanged.

Measured on python SDK 2.0.0, identical on `memory://` and `surrealkv://` (both on the SDK-bundled core **2.3.10**, read off the extension binary rather than inferred from the Python package version) and a standalone `ws://` **2.3.7** server. Every leg is 2.3.x, so the behaviour is measured on **2.3.x only** — not below 2.3, and not on 3.x. The SDK pin is `>=2.0.0,<3.0` explicitly for 3.x support, so that configuration is reachable and uncovered.

| Site | Operator | Before | After |
|---|---|---|---|
| `list_documents(updated_before=…)` | `<` | **0 rows** | 3 of 6 |
| `_checksum_reingestable_clause` | `>=` | **5 checksums** (clause inert) | 4 |
| `claim_orphaned_documents` SELECT | `<` | **0 rows** | 2 |
| `claim_orphaned_documents` UPDATE | write | rejected (unreachable) | persists a real datetime |

## ⚠️ Operator impact — two behaviours activate

Both features exist today but have **never run** on the SurrealDB backend. This is convergence onto the PostgreSQL behaviour shipped in #1464, not new semantics. Operators running `backend: surrealdb` should expect a change in behaviour, not just a bug fix.

**1. Crash-abandoned re-ingest (#1464) starts working.** The stale-PENDING dedup exclusion compares `updated_at >= $pending_stale_before`. Pinned true, the disjunct never narrowed and the clause collapsed to the legacy `status != 'failed'` path — byte-equivalent to the feature being off. **A PENDING row left behind by a crashed ingest was a permanent dedup hit, so its document could never be re-ingested** — no error, no log, the content simply never came back. Such documents will now re-ingest on their next submission. Fresh PENDING rows still dedup, so the concurrent in-flight guard is unaffected.

**2. Orphan reclamation starts working, with a first-run backlog burst.** The claim filters were pinned false, so the query returned zero rows on every call and no orphaned document was ever reclaimed. The pending processor loops per namespace in batches of 100 until drained, so **the first sweep after upgrade claims the entire accumulated backlog** — one LLM extraction per item, which is the first-order cost. Sizing it means counting PENDING/PROCESSING rows older than the configured cutoffs (defaults: 5 min pending, 900 s processing).

This burst only applies if you actually run the pending processor — it is started explicitly via `start_pending_processor()`. Deployments that never start it get behaviour 1 and no burst.

**Interaction between the two, worth knowing before you upgrade:** both features derive their cutoff from the same `pending_processor_grace_period_minutes`. So a single stale PENDING row is simultaneously a dedup *miss* — a resubmit mints a **new** document row, since the checksum index is not UNIQUE — and *claimable* by the sweep, which processes the **old** row. Two documents, two extractions, one piece of content. Pre-existing on PostgreSQL; on SurrealDB both switch on at once, against a backlog.

**The claim is not atomic on this backend.** SELECT-then-per-row-UPDATE with no locking, where PostgreSQL uses `FOR UPDATE SKIP LOCKED`. Two processors sweeping the same namespace can both claim and both enqueue the same orphan — duplicated extraction, not corruption. Previously unreachable; now live, and most likely precisely during the first-run burst. Note `khora.py` currently asserts claims "are atomic" and qualifies only the PostgreSQL leg.

### Query-plan costs (pre-existing, unchanged by this PR, logged for follow-up)

Neither is a regression — the SQL text is identical before and after — but both are worth stating since this PR is what makes the first one execute:

- The **claim query's** top-level `OR` demotes it to `Iterate Table`; `EXPLAIN` confirms the plan is byte-identical for the string and object binds, while each split leg alone plans as `Iterate Index`. Draining N orphans costs ~`ceil(N/100)` full scans per namespace, each reading every `document` row across all namespaces.
- The **dedup query** has the same shape on the *hot* path: its nested `OR` demotes `idx_document_ns_checksum` to `Iterate Table`, and `pending_stale_before` is never `None` in production, so every `remember()` on this backend already full-scans `document`. Higher volume than the claim query.

Splitting those disjunctions into index-eligible queries is the follow-up `scan_documents` already names. Folding it in here would turn a binding fix into a query rewrite.

### No data repair needed

The claim path's write was unreachable (the SELECT ahead of it returned nothing), and SurrealDB **rejects** a string into a `TYPE datetime` field rather than coercing it — discarding the whole statement, so a multi-field `UPDATE` leaves every field untouched. No existing row can hold a string timestamp; there is nothing to migrate or clean up. (Precisely: the *string* failure mode cannot have applied partially, since one timestamp is bound for every row and row 1 raises first. A transient error mid-loop could leave earlier rows flipped, but those self-heal on the next sweep.)

### Scope

SurrealDB only. PostgreSQL, `sqlite_lance` and the `backend: sqlite` stack are unaffected — `storage/backends/sqlite.py`'s `.isoformat()` is correct, since that backend stores `TEXT`. A sweep of the whole `surrealdb/` package found no fifth site: `vector.py`, `graph.py`, `event_store.py` and `_entity_to_bindings` already bind objects, and `_parse_dt` (read direction) is unaffected — the SDK returns native `datetime` objects. The metadata-column `.isoformat()` in the filter compiler is a *different*, still-correct divergence (metadata round-trips through a `FLEXIBLE object` column, which preserves the written type, gated with `type::is::string`) and is deliberately untouched.

### Also in this change

`scan_documents`' docstring described its object bind as a **deliberate divergence** from `list_documents`, with a companion comment telling readers not to "correct" it back. Both are false now that the two agree — and the docstring's generalization ("an ISO string compares against no row at all") is true only for `<`, which is plausibly what produced the `>=` site in the first place. Replaced with a single module-level note stating the mechanism; the same over-generalization is scoped to `<` at the two remaining sites that stated it as a property of the type pair. No behaviour change in `scan_documents`.

Review follow-ups in the second commit: the module note's core version is corrected from an inferred 2.0.0 to the measured 2.3.10 with its provenance recorded; three reads (`claim_orphaned_documents`, and the single and batch checksum lookups) whose `namespace_id` predicate no test pinned now have coverage that fails when the predicate is neutralized; and claimed documents carry the refreshed `updated_at` rather than the pre-claim value, matching the PostgreSQL sibling.

## Test plan

- [x] `pytest tests/integration/storage/backends/surrealdb/` — **91 passed**
- [x] `pytest -k surreal` across unit + integration — 558 passed, 0 failed
- [x] **Every new test verified RED against the pre-fix binds**, not merely asserted. Reverting all four binds at once (the true pre-fix state) fails **8 of the 11** datetime tests; each bind was also reverted individually, and the two claim-path binds fail *differently* (silent under-select vs. hard `InternalError`), confirming each is independently load-bearing.
- [x] **Namespace-isolation guards added and each verified to bite.** Neutralising `list_documents`' `namespace_id` predicate previously left all 78 SurrealDB tests green; it now fails exactly the two new tests with 87 others passing. The same hole on `claim_orphaned_documents` and both checksum reads is now covered — the claim-path assertion checks foreign rows are neither returned **nor re-stamped**, since that path writes. The seed pins foreign ids *above* scanned ones so a leak displaces expected rows inside a bounded page rather than falling off the end.
- [x] `ruff check`, `ruff format --check`, `ty check` clean
- [ ] Postgres/Neo4j lanes not run locally — Docker unavailable in this environment. Unaffected by this change (SurrealDB-only), and CI covers them.

Pre-existing unit failures (69, all `sqlite_lance` / `langgraph` / `dream` / filter-conformance, rooted in an `ImportError` on `SQLiteLanceRelationalAdapter`) reproduce identically with these changes stashed — 69 failed / 50 passed both ways.

### Known follow-ups, deliberately not in this PR

- Split the claim and dedup `OR` disjunctions into index-eligible queries (both full-scan today, unchanged by this PR).
- Batch the claim's per-row `UPDATE` loop, as PostgreSQL does — on a remote connection each row is a round trip.
- Eight further `documents` accessors on this adapter share the untested-`namespace_id` shape; inert with respect to this diff.
- A `.isoformat()` lint gate scoped to `storage/backends/surrealdb/` would need zero exclusions today.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved SurrealDB date and time filtering, deduplication, document scanning, status updates, and orphan claiming.
  * Ensured timestamp boundaries and namespace filtering behave consistently across document listings and scans.
  * Prevented immediate duplicate orphan claims and preserved accurate update timestamps.

* **Tests**
  * Added comprehensive integration coverage for datetime handling, filtering, isolation, status transitions, and orphan processing.

* **Documentation**
  * Clarified SurrealDB datetime comparison and write behavior.


filippetrovic:
## Multi-perspective review

Eight reviewers ran independently against this diff — security, performance, correctness, tests, database, docs, observability, devil's advocate. A dedicated security audit ran afterwards. Findings are consolidated and deduplicated below.

**Overall assessment: APPROVE WITH NITS.** No CRITICAL or HIGH findings. Two MEDIUM items were already found and fixed before the PR opened; the remaining items are advisory and none block merge.

### Verification standard applied

The defect being fixed survived to production *because no test ever exercised the parameter*. A review that only reads the diff would reproduce that failure mode, so every central claim here was re-measured rather than accepted:

- The type-semantics claim was independently reproduced on three modes — `memory://`, `surrealkv://` (both SDK-bundled core 2.0.0) and a standalone `ws://` **2.3.7** server started for the purpose. The operator truth table is value-independent across five string operands including a non-timestamp and the empty string. The write rejection discards the *whole statement*, verified on a multi-field `UPDATE` that leaves both fields untouched.
- The RED claims were redone from scratch by reverting **all four binds simultaneously** — the true pre-fix state, which no individual test run had exercised: **7 failed / 82 passed**. Each bind was also reverted individually. The two claim-path binds fail *differently* (silent under-select vs. hard `InternalError`), confirming each is independently load-bearing.
- The namespace predicate was neutralised directly: exactly the two new tests fail, 87 others stay green. That reproduces the reported hole (previously all 78 SurrealDB tests stayed green under the same mutation) and demonstrates it is now closed.

### MEDIUM — found in review, already fixed in this PR

**M1 · docs — stale prose the fix falsified.** `test_relational_scan_documents.py` still asserted that `scan_documents` "diverges deliberately from `list_documents`, which binds `.isoformat()` and is defective for it… tracked separately and deliberately not tested here." False on all three clauses after this change. The equivalent prose inside `relational.py` had been corrected but the sweep missed the test file. **Fixed.**

**M2 · tests — a vacuous assertion.** `test_claim_leaves_untouched_rows_byte_for_byte` passed even with all four binds reverted: an under-selecting claim trivially touches nothing, so it never bit on the real bug. A `len(claimed) == 2` guard was added and verified to fail (`assert 0 == 2`) under the SELECT-revert mutant where it previously passed. **Fixed.**

### LOW / advisory — not blocking

**L1 · devil's advocate — one PR bundles a pure bug fix with two behaviour activations.** `list_documents` is separable even though the claim-path SELECT/UPDATE are not. Accepted as-is: single root cause, one-line fix per site, and the activations are convergence onto already-shipped PostgreSQL behaviour rather than new semantics. Splitting would leave a knowingly half-fixed adapter on `main` for no risk reduction. Recorded so the tradeoff is explicit rather than implicit.

**L2 · performance / devil's advocate — the first-deploy drain has no throttle.** Disclosure lives in the PR body, which is a weaker mitigation than a cap or a dry-run count would be. The cost is real but bounded and one-time, and it is the *backlog* that is expensive, not the change. Flagged for the author's judgement; a follow-up guard is defensible.

**L3 · performance / database — the claim query plans as a full table scan, before and after.** The top-level `OR` costs the index, not the bind (`EXPLAIN` confirms `Iterate Table` identically pre- and post-fix, while `namespace_id` alone and each split leg plan as `Iterate Index`). So draining N orphans costs ~`ceil(N/100)` scans per namespace, each reading every `document` row across all namespaces. **Pre-existing and merely exposed by this PR, not introduced by it** — splitting the `OR` into two index-eligible queries is the follow-up `scan_documents` already names, and folding it in here would turn a binding fix into a query rewrite. Correctly deferred.

**L4 · docs — SurrealDB 3.x is unmeasured, and the caveat is load-bearing.** The SDK pin is `>=2.0.0,<3.0` explicitly *for* 3.x support, so a 3.x engine is a reachable, uncovered configuration. The comment says so plainly rather than implying coverage it lacks — the right call, worth preserving through any future edit.

**L5 · docs — the register entry is drafted but unapplied.** A `CLAUDE.md` Backend Specifics bullet recording the type-rank semantics is written and verified but deliberately left for the author to apply, being project configuration rather than ticket scope. Repo convention (cf. #1582) is to land such bullets in-PR; recommended, not required.

**L6 · correctness — misleadingly named `now_iso` locals elsewhere in the package** already hold `datetime` objects and bind correctly. Renaming is not traceable to this change; leaving them was the right restraint. Noted only so the next reader is not alarmed by the names sitting under a module note that says "never `.isoformat()`".

### Security audit

Clean — **PASS**. No injection sinks (the four changed lines are parameterized binds; no f-string query construction added or modified), no secrets, no new entry points, no untrusted-input path, no dependency or manifest changes, no deserialization/SSRF/path surface. The `# noqa: S608` SELECT in this file interpolates only an internally-constructed conjunct list, not user input, and is untouched.

Worth stating positively: this PR **improves** the tenant-isolation posture. `list_documents`' `namespace_id` predicate was structurally untested — an IDOR-family exposure that a prior mutation confirmed no test would catch. It now has a regression guard designed so a leak *displaces* expected rows inside a bounded page rather than falling off the end, which is what makes it detectable at all.

### Domain verdicts

| Domain | Verdict |
|---|---|
| Security | Pass — clean; isolation posture improved |
| Performance | Pass with advisories (L2, L3) — both pre-existing exposures, not regressions |
| Correctness | Pass — fix complete at all four sites; no placeholders |
| Tests | Pass — after M2; RED verified independently |
| Database | Pass — type-semantics claim reproduced on all three modes; sweep confirmed no fifth site |
| Docs | Pass after M1 — every cited number reproduced; the one inference is labelled as one |
| Observability | Pass — no instrumentation surface added; see L2 on visibility of the drain |
| Devil's advocate | Merge with the tradeoffs recorded (L1, L2) |

### Note on CI

The failing checks on this PR are a **GitHub Actions platform outage**, not code: `The job was not acquired by Runner of type hosted even after multiple attempts` and `Failed to resolve action download info. Error: Service Unavailable`. Every job that has actually obtained a runner has passed, including `skeleton_surrealdb` — the check that most directly exercises this change. Re-runs are in progress.

filippetrovic:
## Corrected consolidated review — supersedes the summary above

My previous comment was posted before the full reviewer reports had come back, and it understated the findings. It said "no CRITICAL or HIGH findings"; three reviewers independently found HIGH items, one of which is a factual error in permanent source comments. The verification-standard section of that comment stands — it was based on real measurement — but the findings table did not. This comment replaces it.

**Overall assessment: APPROVE WITH CHANGES.** No CRITICAL. Four HIGH, of which three are being fixed in this PR and one is disclosure-only. Every reviewer independently re-measured rather than trusting the PR body, and two of the HIGH findings come from two reviewers arriving at them separately.

### HIGH — fixing in this PR

**H1 · docs — a wrong number asserted as measured, in the module note.** The note credits the embedded legs to "the SDK-bundled core 2.0.0". That is inferred from the Python package version and is wrong. Verified directly:

```
strings .venv/.../surrealdb/_surrealdb_ext.abi3.so | grep -oE "surrealdb-core-[0-9.]+"
  349  surrealdb-core-2.3.10
```

`surrealdb-2.0.0` appears exactly once — the SDK crate, not the engine. Two consequences: the version is wrong, and since all three measured legs are therefore **2.3.x** (2.3.10 embedded, 2.3.7 server), the closing claim of "SurrealDB 2.x" over-claims — nothing below 2.3 was measured. This is the same defect class the PR exists to fix — a real measurement generalized past what was run — fifteen lines above the fix, in the register that is trusted precisely because it is specific. Being corrected to core 2.3.10 and narrowed to 2.3.x.

**H2 · security + tests — the namespace predicate is untested on the path this PR activates.** Found independently by two reviewers. Mutating `namespace_id` to a tautology in `claim_orphaned_documents` (`relational.py:~625`) and in `get_document_by_checksum` (`:~778`) leaves **all 89 tests green** — the identical blind spot this PR exists to close, still open on two paths. The claim path is the urgent one: this PR makes it execute for the first time, and it **writes** (status → PROCESSING, `updated_at` re-stamped) and drives one LLM extraction per claimed row, so a scope leak there is a cross-tenant destructive write plus unbudgeted spend, not a read leak. Second-namespace coverage is being added, asserting foreign rows are neither returned nor re-stamped.

A further eight `documents` accessors on this adapter have the same untested predicate. Pre-existing and inert with respect to this diff — worth a follow-up issue, not a blocker.

**H3 · performance — the claim path issues one `UPDATE` per row.** PostgreSQL batches (`UPDATE … WHERE id IN (…)`). Measured embedded: 193 ms vs 127 ms per 500 rows — only 1.5×, but on `ws://` each row is a network round trip, so a 10k backlog at 2 ms RTT is ~20 s of serialized round trips before the first document is processed. Newly reachable, since the loop never executed before. Small fix mirroring PG; deferring to a follow-up is defensible given it is a shape change rather than a bind fix, but it should be logged rather than dropped.

### HIGH — disclosure only, not fixable here

**H4 · devil's advocate — the two activated features share one cutoff.** `_pending_stale_cutoff()` and the sweep's `pending_before` both compute `now - pending_processor_grace_period_minutes`. So one stale PENDING row is *simultaneously* a dedup miss — and a resubmit mints a **new** row, since the checksum index is not UNIQUE — *and* claimable by the sweep, which processes the **old** row. Two documents, two extractions, one piece of content. Pre-existing on PostgreSQL and not introduced by this diff, but on SurrealDB both switch on at once against an accumulated backlog. Added to the operator-impact section.

### MEDIUM

**M1 · devil's advocate + docs — the falsified over-generalization survives in two more places**, one of them in a docstring this PR already edits (`test_relational_scan_documents.py:23-26` and `:694-696`). Both state as a property of the type pair something true only for `<`. Both are safe in context because both sites use `<` — which is precisely how the `>=` defect propagated. Being scoped.

**M2 · correctness + database — the claim path is not atomic, and this PR makes that reachable.** SELECT-then-unguarded-per-row-UPDATE, no locking, where PostgreSQL uses `FOR UPDATE SKIP LOCKED`. Two processors sweeping the same namespace both claim and both enqueue: duplicated extraction, not corruption. Previously unreachable. Compounds with the first-run burst, which is exactly when a second instance is most likely sweeping. Note `khora.py`'s docstring asserts claims "are atomic" and qualifies only the PG leg — a reader on SurrealDB will take that at face value. Disclosed; the conditional-UPDATE fix is a follow-up.

**M3 · database — the dedup query full-scans, on the hot path.** Confirmed by `EXPLAIN`: the nested `OR` demotes `idx_document_ns_checksum` to `Iterate Table`, and `pending_stale_before` is never `None` in production, so **every `remember()` on SurrealDB already full-scans `document`**. Not a regression — the SQL text is unchanged — but it is higher-volume than the claim-query scan the PR does document, and the PR is silent on it. Being added to operator impact; fix is a follow-up of the same shape.

**M4 · tests — a nondeterministic failure, observed twice on unmodified source, not reproducible in ~900 subsequent runs.** The reviewer declined to guess at a cause, which is the right call. Flagged because a nondeterministic guard on a silent-data-loss path deserves a CI loop before merge rather than a shrug.

**M5 · tests — mode coverage is `memory://` only.** The PR measured `surrealkv://` and `ws://` by hand but nothing in the suite exercises them, and `surrealkv://` is precedented elsewhere in this directory. Worth a parametrised fixture.

**M6 · docs — no CHANGELOG entry.** Convention is mixed for pure index/scan PRs, but the comparable *behaviour-activating* ones (#1583, #1579) each carried one. Two features going from inoperative to live on a shipped backend belongs under `[Unreleased] ### Fixed`.

### Corrections to the PR body itself

- **"7 failed / 82 passed" should be 8.** Independently measured: the all-four-binds revert fails 8 of the 11 new tests. The PR's number predates its own added guard.
- **"Partial application is impossible" is too broad.** True for the string-rejection mode only (one timestamp bound for all rows, so row 1 raises first). A transient error at row *k* leaves rows `0..k-1` flipped while the caller gets nothing — self-healing, and the *conclusion* (no repair needed) survives, but the justification as written overreaches. The in-code comment is correctly scoped; only the prose is not.
- **The off-peak recommendation reads as unconditional.** The sweep only runs if the operator calls `start_pending_processor()`; deployments not running it get the re-ingest change and no burst.

### Where the PR is right (verified, not assumed)

Recorded because several reviewers actively tried to break these and could not: the type-rank claim reproduces exactly on all three modes, including the reverse operand order, which corroborates the labelled inference without over-claiming it; the write rejection discards the whole statement, verified on a multi-field UPDATE; the "no fifth site" sweep holds across every sibling adapter; the filter-compiler `.isoformat()` carve-out is correct and correctly untouched (a reviewer verified `FLEXIBLE object` columns really do preserve the written type); splitting `list_documents` out would buy nothing, since no production caller passes `updated_before`; the claim SELECT/UPDATE really are inseparable; and 669 test lines for four binds is not disproportionate — exact row sets, verified RED, with the `pending_stale_before=None` control being what stops "stale re-ingests" passing for a mutant that made *everything* re-ingestable.

### Security audit — PASS

No injection sinks (all four changed lines are parameter binds), no secrets, no new entry points, no dependency changes, no deserialization/SSRF/path surface. The `# noqa: S608` SELECTs interpolate only internally-assembled literal conditions; the sole non-literal fragment is guarded by a genuine identifier allowlist (`^[A-Za-z_][A-Za-z0-9_]*\Z`), untouched here. The one LOW is cost amplification on the activated drain — same shape as PostgreSQL since #1464, correctly mitigated for this PR by the off-peak recommendation.

### CI

Failures on this PR are a **GitHub Actions platform outage** — `The job was not acquired by Runner of type hosted even after multiple attempts` and `Failed to resolve action download info: Service Unavailable`. Every job that has obtained a runner has passed, including `skeleton_surrealdb`, `vc_full`, `skeleton_pgvector`, `skeleton_sqlite_lance` and `no-agents-md`. Re-runs continuing.

filippetrovic:
### CI status: blocked by a GitHub Actions outage, not by this change

GitHub Actions has been in a **major outage** since 15:22 UTC — status-page incident `qcvjkzcs7j74`, impact `critical`, still unresolved. Workflow runs for the latest commit were never created (the `pull_request` webhook was dropped), and the runs from the previous commit have sat `queued` for hours. Closing/reopening the PR and repeated `gh run rerun --failed` both failed to produce new runs.

Every job that has managed to obtain a runner on this branch has **passed**:

`skeleton_surrealdb` · `vc_full` · `skeleton_pgvector` · `skeleton_sqlite_lance` · `no-agents-md` · `conformance (surrealdb)` · `conformance (python)` · CodeRabbit

Every failure carries one of two infrastructure errors — `The job was not acquired by Runner of type hosted even after multiple attempts` or `Failed to resolve action download info: Service Unavailable` — i.e. the job never executed any of this repository's code.

Verified locally on the current head (`d3a5a6c9`) instead:

| Check | Result |
|---|---|
| `pytest tests/integration/storage/backends/surrealdb/` | 91 passed |
| `pytest -k surreal` (unit + integration) | 558 passed, 0 failed |
| `ruff check` / `ruff format --check` | clean |
| `ty check` | clean |

Not run locally: the Postgres and Neo4j lanes, since Docker is unavailable in this environment. They are unaffected by this change — it is SurrealDB-only — but they have not been exercised, so **CI should be confirmed green before merge** once Actions recovers.

Pre-existing unit failures (69, all `sqlite_lance`/`langgraph`/`dream`/filter-conformance, rooted in an `ImportError` on `SQLiteLanceRelationalAdapter`) reproduce identically with this branch's changes stashed — 69 failed / 50 passed both ways — so they are not attributable here.

filippetrovic:
### CI update: two of three workflows now green; the third is stuck behind the outage

Since the previous comment, runners partially recovered and two workflows completed on this branch:

| Workflow | Result |
|---|---|
| **E2E Engine Matrix** | ✅ success — includes `skeleton_surrealdb`, `vc_full`, `vc_embedded`, `chronicle`, `skeleton_pgvector`, `skeleton_sqlite_lance` |
| **Filter Conformance** | ✅ success — all 7 legs including `conformance (surrealdb)` |
| **CI** | ⏳ still `queued` after several hours |

No workflow runs exist for the two most recent commits. That is not a configuration problem on this PR — GitHub's own incident update states it directly:

> Webhook triggers remain throttled to aid recovery, so many push and pull request events are not triggering new workflow runs. […] we are processing approximately 15% of webhooks

Which is why pushing new commits and closing/reopening this PR both failed to schedule anything. Actions remains at `major_outage` (incident `qcvjkzcs7j74`, critical, unresolved since 15:22 UTC).

**Where that leaves the change.** The engine and conformance suites passed on the first commit, which carries the entire behavioural change — all four datetime binds. The two later commits add test coverage, correct a version number in a comment, refresh `updated_at` on returned claim objects, and add a changelog entry; verified locally at head:

| Check | Result |
|---|---|
| `pytest tests/integration/storage/backends/surrealdb/` | 91 passed |
| `pytest -k surreal` (unit + integration) | 558 passed, 0 failed |
| `ruff check` / `ruff format --check` / `ty check` | clean |

**Please confirm a full green CI run before merging** once Actions recovers — a re-run or a rebase-push should schedule one. The `test-unit`, `test-integration`, `lint` and `security` jobs in the CI workflow have not executed against any commit on this branch, and the Postgres/Neo4j lanes could not be run locally (no Docker in that environment).

filippetrovic:
### ✅ CI is green on the head commit

The Actions outage throttled webhooks to ~15%, so `pull_request` events were never scheduling runs. All three workflows support `workflow_dispatch`, which bypasses the webhook path — dispatched against this branch, every job ran and passed on `29f2e009`.

**25 check-runs on the head commit, all `success`, zero failures:**

| Workflow | Jobs |
|---|---|
| **CI** | `install` · `lint` · `security` · `secret-typing` · `no-agents-md` · `test-unit` · `test-integration` · `weaviate-integration` · `test-google-adk` · `examples-smoke` |
| **E2E Engine Matrix** | `skeleton_surrealdb` · `skeleton_pgvector` · `skeleton_sqlite_lance` · `skeleton_weaviate` · `vc_full` · `vc_embedded` · `chronicle` |
| **Filter Conformance** | `surrealdb` · `postgres` · `cypher` · `chronicle` · `sqlite_lance` · `weaviate` · `python` |
| plus | `codecov/patch` · CodeRabbit |

This closes the gap flagged in my earlier comments: `test-unit` and `test-integration` cover the Postgres and Neo4j lanes that could not run locally (no Docker there), and they pass. Nothing about this change is now unverified.

One caveat on the PR's checks list: because these were `workflow_dispatch` runs, GitHub does not surface them in the PR's own check summary — only `pull_request`-triggered runs appear there. They are recorded against the head commit and visible via `gh api repos/DeytaHQ/khora/commits/29f2e009/check-runs`. Once Actions fully recovers, a rebase-push will populate the PR checks list normally if you'd prefer that before merging.

msrdic (approved):

PR #1589, opened 2026-08-07
Harden the document-store scan primitive across backends

filippetrovic:
## Summary

Follow-up hardening for the bounded keyset `scan_documents` primitive across all four document stores (raw-SQLite, SurrealDB, and the two SQLAlchemy stores). These are latent-defect and test-gap fixes found reviewing the primitive before it gains callers.

- **Cross-tenant bind guard (SurrealDB):** the compiled filter's binds were merged over the scan's own binds, so a compiled bind named `ns`/`lim`/`status` could overwrite the tenant scope and silently return foreign-namespace rows. Reject any bind-name collision explicitly, with a test that drives a colliding stub compiler.
- **Keyset cursor derived from the raw row (all four stores):** the cursor is now built from the raw result row via a shared `build_scan_step` helper. Previously two stores read the *converted* document, whose converter coalesces a missing `created_at` to `now()` (and a non-UUID id to a derived UUID); either fallback floats the cursor above the whole window and turns the walk into an unbounded loop. `exhausted` is derived from the raw row count, never the post-filtered document list.
- **Metadata path segments defer instead of raising (SurrealDB):** a filter whose metadata path has a non-identifier segment (e.g. `metadata.foo-bar`) now defers to the coordinator's residual on the `documents` (`split`) context in every position, matching the other backends, rather than raising and diverging by position. The emit-time identifier guard is retained as raise-mode defense-in-depth (injection protection on direct/temporal compiles).
- **Serialization + observability:** raw-SQLite `updated_before` routes through the shared datetime serializer; the SurrealDB connection logs the unexpected-driver-result-shape branch (which otherwise reads as clean exhaustion).
- **Contract + de-duplication:** refreshed the cross-store scan-key contract note (four stores, three key shapes); extracted the shared where-clause; lifted ~120 lines of duplicated test helpers into the shared test-helper module.
- **Test gaps closed:** a tie-break test that drops the sort index so the `id DESC` ordering is actually exercised; anti-vacuity assertions so a constant-empty filter shape can no longer pass the superset invariant silently; raise-path, cross-store parity, and connection-warning tests.

No public API change (the primitive is intentionally not yet on the backend protocol).

## Test plan
- [x] Unit lane green (`make test-unit`: 10675 passed, 84% coverage)
- [x] Lint/format/type gate green (`make lint`)
- [x] Scan integration lanes green locally (PostgreSQL: 13 passed against a live server; SurrealDB: 34 passed)
- [ ] Full integration/e2e matrix passes in CI

## Summary by CodeRabbit

- **Bug Fixes**
  - Unsafe metadata filter paths are safely deferred for post-filtering rather than included in queries.
  - Explicit strict mode continues to reject unsafe filter paths.
  - Scan cursors and exhaustion indicators remain accurate across filtering, pagination, and timestamp ties.
  - Unsupported filters no longer incorrectly terminate scans.
  - Document scanning is more consistent across storage backends.
  - Unexpected database results now generate a warning and safely return no rows.
  - Invalid scan cursor data is detected instead of silently accepted.


filippetrovic:

Reviewed by an 8-perspective panel (security, correctness, database, performance, test-quality, docs, observability, devil's advocate) plus a dedicated security audit. Each reviewer worked the diff independently.

**Overall: APPROVE.** No CRITICAL / HIGH / MEDIUM findings. The design is sound; the keyset primitive is index-served on the three SQL tiers; the cross-tenant guard, raw-row cursor, and five-position metadata deferral are all verified correct. The only findings were doc-accuracy nits in the PR's own contract prose — now fixed in this branch — plus two non-actionable LOWs (one follow-up, one pre-existing).

### LOW — doc accuracy (FIXED in this PR, latest commit)
- **L1 [docs/correctness]** `base.py` `build_scan_step` docstring claimed the `on_unsupported="split"` invariant was unenforced ("nothing enforces it… worth adding a test"). It is in fact enforced by `test_documents_compile_contexts.py::test_context_declares_the_physical_schema`, parametrized over all four contexts. Corrected to point at that test.
- **L2 [docs]** `base.py` over-claimed what `build_scan_step` centralizes. Scoped precisely: it centralizes the `exhausted` computation; `last_scanned` is extracted per store and stored verbatim (the `raw_row_count` parameter name is an advisory guardrail, not an enforced one).
- **L3 [docs]** `test_documents_compile_contexts.py` map-header + inline comment described only one of the two fixes SurrealDB needed for split-mode soundness and still read "not yet sound under split". Both fixes now named (placeholder inversion under `$not`/`$or`, and the non-identifier metadata-segment raise).

### LOW — not addressed here (rationale)
- **L4 [observability]** The new `surrealdb/connection.py` unexpected-result-shape `WARNING` has no paired `degraded_total` counter, so the (silent) "reads as clean exhaustion" failure is greppable but not alertable. A bare WARNING is proportionate for a rare driver-contract anomaly at this primitive layer (no result observability dict exists there to attach a structured `Degradation` to). Adding the counter also requires a `docs/telemetry-contract.json` update — better as a small follow-up than scope creep here.
- **L5 [performance]** SurrealDB keyset pagination is O(corpus × steps), not O(scan_limit): SurrealQL lacks row-value comparison so the cursor expands to a top-level `OR` the 2.x planner can't index, forcing a cross-tenant table scan per step. **Pre-existing and unchanged by this PR** — exhaustively documented in `surrealdb/relational.py` with measurements and a tracked follow-up. Operator-facing note, not a blocker.

### Security audit — PASS
No dependency/manifest changes (no supply-chain surface). Production query construction binds every value via `$params` / positional placeholders and interpolates only validated identifiers (closed whitelist + `_SAFE_SEGMENT_RE` with a `\Z` anchor that blocks the trailing-newline bypass); the only raw SQL strings in the diff are static test fixtures. The new bind-collision guard *strengthens* multi-tenant isolation (a compiled filter bind can no longer overwrite the tenant scope), and the namespace predicate is present in both readers on all four stores. No crypto / deserialization / file / redirect / SSRF / endpoint surface. No findings.

### Confirmations (INFO)
- All new/changed tests are mutation-verified and non-vacuous: the §-tie-break test kills the `id DESC` mutant (drops the index so it can't mask it); anti-vacuity `assert oracle` guards can fail on a constant-empty shape; the metadata-deferral parity test compares against the `compile_python` oracle and asserts the leaf lands in the residual, not `consumed_keys`; the bind-collision and raw-row-cursor tests each fail on their mutant.
- `exhausted` derives from the raw SQL row count, never the post-filtered document list.
- Scope discipline held: `scan_documents` was **not** added to `RelationalBackendProtocol` (deferred to its own PR); no Linear/internal identifiers in the diff; CLAUDE.md untouched.

sale87 (approved):

PR #1590, opened 2026-08-07
Split OR-defeating SurrealDB document queries into index-eligible legs

filippetrovic:
## Summary

An `OR` anywhere in a `document`-table `WHERE` collapses the SurrealDB 2.x planner to a full table scan — and the collapse also drops the `namespace_id` index prefix, so the statement reads every `document` row in the database, all tenants included, then filters and sorts in memory. On a resumed keyset walk this makes each step's cost track total corpus bytes rather than the caller's page size or namespace, turning a bounded page into a cross-tenant scan.

The scope of that behavior is the `document` table specifically: 2.x *can* union index scans across a disjunction, but only when every disjunct is a comparison on a single-field index, and `document` carries composite `(namespace_id, …)` indexes and nothing else — so no disjunct on it ever qualifies. (Measured with `EXPLAIN`; the reasoning is recorded in a note atop the adapter, and a test keeps the premise honest against future engine changes.)

Three affected reads are restructured into `OR`-free conjunctive legs merged in Python:

- **Keyset resume (`scan_documents`)** — the disjunctive cursor becomes a tie-block query (`created_at = cursor AND id < cursor_id`) and a strictly-older query (`created_at < cursor`), concatenated in that order. Because every tie-block row sorts strictly above every strictly-older row under `created_at DESC, id DESC`, the concatenation is the same total order as the old single query — byte-identical windows, no comparator, no dedup. The second query runs only for the remaining limit and is skipped entirely when the first fills the window. The shared narrowing (namespace, `status`, `updated_before`, compiled filter fragment) is factored into one helper carried by both legs.
- **Checksum dedup** — the nested `OR` becomes two disjoint status legs, the settled leg probed first so a settled row still outranks a fresh-pending one.
- **Orphan-claim sweep** — one query per status/cutoff pair, merge-sorted by `updated_at` and sliced to the limit before claiming.

A repo-wide sweep of remaining `OR`-in-`WHERE` sites was done; the ones on other tables that share this defect class (larger, schema-level) are recorded separately and left out of this change.

## Test plan

- [x] Unit tests pass (new plan-assertion module: 31 tests, `memory://`, no docker)
- [x] Integration tests pass (existing `scan_documents` suite unchanged; 578 integration passed)
- [x] Full suite green (`make test`), coverage 85.48% (≥77% gate)
- [x] Lint/format clean (`make format`, `make lint` — zero new type diagnostics vs baseline)
- New tests drive the real adapter through a recording connection and re-issue each captured statement with `EXPLAIN`, asserting a namespace-scoped `Iterate Index` for every `document` read (keyset step 1, both resumed-step statements, both dedup legs, both orphan legs); a companion test pins that the old disjunctive shapes really do `Iterate Table`, so the split is not cargo-culted. No-filter shapes only — a filtered `$or` fragment still table-scans by construction and is out of scope.

## Summary by CodeRabbit

* **Performance Improvements**
  * Improved database query planning for document scans, checksum lookups, and orphan recovery.
  * Increased use of indexes by splitting complex searches into efficient query operations.
  * Preserved correct ordering, filtering, deduplication, and namespace isolation during scans.
* **Bug Fixes**
  * Ensured settled documents take precedence over newer pending records during checksum lookups.
  * Improved handling of resumed scans and concurrent orphan claims.
* **Tests**
  * Added comprehensive coverage for query plans, index usage, cursor boundaries, and result consistency.

### DIFF
--- a/src/khora/storage/backends/surrealdb/relational.py
+++ b/src/khora/storage/backends/surrealdb/relational.py
@@ -62,6 +62,46 @@ if TYPE_CHECKING:
 # ---------------------------------------------------------------------------
 
 # ---------------------------------------------------------------------------
+# Disjunctions and index selection
+#
+# **No ``document`` read in this module may contain an ``OR``.** Three of them
+# used to, and each one made the statement read every ``document`` row in the
+# database — all tenants included, not just the namespace asked for.
+#
+# The rule, measured on 2.x rather than assumed. SurrealDB *can* answer a
+# disjunction as a union of index scans, but only when EVERY disjunct is a
+# comparison on a field carried by a **single-field** index. A composite index
+# does not qualify, not even on its own leading column:
+#
+# * ``status = 'a' OR status = 'b'`` with a one-field ``status`` index — two
+#   ``Iterate Index`` operations, and it stays indexed when wrapped in an outer
+#   ``AND`` (three operations, the extra one serving the conjunct).
+# * ``checksum = $a OR checksum = $b`` with only the two-field
+#   ``idx_document_ns_checksum`` — ``Iterate Table``.
+#
+# When one disjunct cannot be served, the fallback is not scoped to that
+# disjunct: the whole statement becomes ``Iterate Table`` and the conjuncts
+# around it lose their indexes too. That is why the ``namespace_id`` scope
+# disappears along with the keyset — the tenant filter still *filters*, it just
+# stops *narrowing*, so correctness is unaffected and cost is not.
+#
+# ``document`` carries composite ``(namespace_id, …)`` indexes and nothing else,
+# by design: the one-field ``idx_document_namespace`` was dropped as a redundant
+# prefix of the sort index. So on this table the qualifying case cannot arise and
+# **every** ``OR`` scans, whatever its shape — verified over eight, including a
+# bare ``OR`` of two composite leading columns and an ``OR`` of two full
+# composite matches. Defining one-field indexes to buy the union back is not a
+# trade worth making: it is permanent write amplification on every ``document``
+# insert to salvage a predicate that splits into conjunctive legs for free.
+#
+# The measurements above are on ``memory://`` with the SDK-bundled core; the
+# adapter's three former ``OR`` sites were each confirmed ``Iterate Table``
+# before the split and ``Iterate Index`` per leg after. Any predicate added here
+# should be checked the same way — the plan tests re-issue every statement this
+# adapter sends with ``EXPLAIN`` appended, so a new ``OR`` fails there.
+# ---------------------------------------------------------------------------
+
+# ---------------------------------------------------------------------------
 # Adapter
 # ---------------------------------------------------------------------------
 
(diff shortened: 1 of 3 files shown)


filippetrovic:

Seven independent reviewers (security, performance, correctness, test, database, docs, devil's-advocate) plus a dedicated diff-scoped security audit. **Verdict: APPROVE** — no CRITICAL / HIGH / MEDIUM findings. Two LOW (descriptive, non-blocking) and a few INFO notes.

### Highlights
- **Net security improvement.** The old ungrouped `OR` could drop the `namespace_id` scope from a disjunct and read other tenants' rows; splitting into `OR`-free conjunctive legs makes that shape unrepresentable — every one of the six legs is namespace-scoped. The bind-collision guard was also hardened from a dynamic `params ∩ compiled.params` check to a static reserved-key set, so a compiled filter can no longer shadow the cursor or tenant scope.
- **Central premise verified and CI-enforced.** The claim "on the `document` table no `OR` disjunct can index-union, because it carries composite `(namespace_id, …)` indexes and nothing else" was checked against the live schema (the one-field namespace index is removed) and is pinned by a test that re-`EXPLAIN`s the three original disjunctive shapes and fails if a future engine learns to index them — so the split is not cargo-culted.
- **Semantic equivalence proven.** The keyset split's concat(tie-block, strictly-older) is byte-identical to the old single window (disjoint by the `created_at` equality; `LIMIT = scan_limit − len(tie)` so no trailing slice). The dedup split's two status legs union to exactly the old predicate (truth-table checked, intersection empty), settled leg first so it still outranks a fresh-pending row. Orphan legs are disjoint by status, merge-sorted and sliced before claiming.
- **Strong tests.** Plan assertions go beyond `Iterate Index` to the seek detail (a namespace-prefix seek with the cursor as a residual filter also prints `Iterate Index` while still scanning the namespace), and an exact per-path statement count catches an accidental extra query. A deterministic `PENDING→PROCESSING` flip between the two orphan `SELECT`s tests the one race the split newly introduces, with a guard against the mutant that would make it pass vacuously.

### LOW (optional, non-blocking)
- **[Docs/Test]** Historical drift in the *unchanged* integration test file: two tests (`test_every_narrowing_leg_composes_in_one_statement`, `test_filtered_walk_puts_a_cursor_and_a_compiled_fragment_in_one_statement`) and a stale body comment in a third still describe the removed single-`WHERE` disjunction mechanism (the resumed step is now two statements). All three assert on **rows**, which hold either way, and the suite's stated philosophy is "assert the outcome, not the mechanism" — so this is descriptive only. Fix is a rename or a one-line "name is historical" note. Left for a follow-up.
- **[Database]** `get_documents_by_checksums`: the batch leg keeps the pre-existing `checksum IN $checksums`, which over a composite-only index seeks the `namespace_id` prefix and evaluates the `IN` as a residual filter — i.e. O(namespace docs), not O(checksums). Pre-existing, correct, tenant-safe, not a regression; only noting that the "one index-eligible statement per leg" wording reads slightly stronger than the plan.

### INFO
- The resumed step is now two non-transactional statements, so a tie-mate inserted between them can be missed — explicitly disclaimed in the docstring and consistent with the method's no-snapshot / no-cross-step-transaction contract.
- The `Iterate Index` guarantee is verified on the embedded core (`memory://`) only; a remote deployment could differ in `EXPLAIN` output, but the `OR`-free legs return identical rows on any engine (correctness is engine-independent — only the plan-verification scope is bounded).
- A compiled filter fragment containing `$or` reintroduces a top-level `OR` and table-scans even the first step. Out of scope here (a compiler push-down/defer decision), documented in the complexity note, and the plan tests correctly cover the no-filter shapes only.

### Security audit
Diff-scoped (added/modified lines only). Injection: the four f-string-into-SQL sites interpolate only static structural fragments plus the compiler's own `$`-bound predicate — no caller value reaches SQL text; all query values are bound parameters. Access control: every read remains namespace-scoped. No secrets, weak crypto, unsafe deserialization, or dependency/manifest changes. **PASS.**

**Overall: APPROVE — ready to merge once CI is green. The two LOW items are optional polish.**

msrdic (approved):

PR #1594, opened 2026-08-07
Wire the document enumeration surface: facade, DocumentPage, engine threading, coordinator scan loop

filippetrovic:
## Summary
- Add `DocumentPage` — a `Sequence[Document]` (iteration / `len` / truthiness keep working) carrying `next_after` (last-scanned keyset position; `None` iff exhausted), `exhausted` (the only termination signal), and `post_filtered_keys` — and `DocumentCursor` `(created_at, id)`. Both exported from the top-level package. Kept distinct from the storage tier's offset-shaped `PaginatedResult`.
- Rewrite `Khora.list_documents` to the keyset contract: validate + lower the recall filter, then reject the non-enumerable `occurred_at` key (the chunk event-time axis has no document-row column) with a structured `RecallFilterValidationError` (`code="key_not_enumerable"`, the nine enumerable keys in `allowed`, copy naming `source_timestamp`/`created_at` as non-equivalent substitutes); enum-validate `status` (unknown → `ValueError`, never a silent zero-match); normalize `after` (`DocumentCursor | Document` → keyset); resolve the config-overridable scan bound; return a `DocumentPage`. No `offset`.
- Thread `filter_ast`/`status`/`updated_before`/`limit`/`after`/`scan_bound` through the engine protocol and all three engines (skeleton, vectorcypher, chronicle) as one-line delegations — no engine interprets the AST.
- Add `StorageCoordinator.scan_documents_page`: the keyset scan loop over the bounded `scan_documents` primitive. Compiles the in-memory residual predicate once per page (not per step), sizes each step to the shortfall (`limit - len(matches)`) so a page never exceeds `limit` and `next_after` is always a raw last-scanned position (never derived from a converted document), giving exactly-once resume. Scan bound `max(limit × 10, 1000)`, configurable, never a caller parameter. The offset-based `list_documents` is left intact for its existing callers.
- Declare the internal `scan_documents` primitive on `RelationalBackendProtocol` so the coordinator calls it across the protocol boundary cleanly.
- Add `document_scan_overfetch_multiplier` (10) / `document_scan_min_bound` (1000) to the query settings.

## Test plan
- [x] Unit tests pass (`DocumentPage` sequence contract; `occurred_at` rejection with structured error + asserted copy, including `$or`/`$not` nesting; enumerable keys pass)
- [x] Integration tests pass (sqlite_lance: multi-page cursor walk — exactly-once / ordered / terminates on `exhausted` / Sequence-compat; resume from a returned `Document`; facade↔storage parity for `status` / `updated_before`; filter narrowing + `occurred_at`/unknown-status errors)
- [x] `make format` + `make lint` clean; full unit suite green locally
- [ ] Full integration + conformance matrix (runs in CI across all backends)

## Summary by CodeRabbit

* **New Features**
  * Added paginated document enumeration with cursor-based continuation.
  * Added filtering by status, update time, system fields, and metadata.
  * Added configurable scan bounds for controlling enumeration limits.
  * Exposed document page, cursor, and status types through the public API.

* **Bug Fixes**
  * Added clear errors for unsupported time-based enumeration filters and unknown statuses.
  * Improved pagination ordering, resumption, page filling, and termination behavior.

### DIFF
--- a/src/khora/__init__.py
+++ b/src/khora/__init__.py
@@ -43,7 +43,7 @@ Example usage:
 """
 
 from .config import KhoraConfig
-from .core.models.document import DocumentSource
+from .core.models.document import DocumentCursor, DocumentPage, DocumentSource
 from .core.models.event import EventType
 from .core.recall_context import context_text
 from .dream import DreamConfig, DreamMode, DreamResult, DreamRunInfo, DreamScope, OpKind
@@ -103,6 +103,8 @@ __all__ = [
     "SearchMode",
     "KhoraConfig",
     "DocumentSource",
+    "DocumentPage",
+    "DocumentCursor",
     # Engine functions
     "create_engine",
     "list_engines",
--- a/src/khora/__init__.py
+++ b/src/khora/__init__.py
@@ -43,7 +43,7 @@ Example usage:
 """
 
 from .config import KhoraConfig
-from .core.models.document import DocumentCursor, DocumentPage, DocumentSource
+from .core.models.document import DocumentCursor, DocumentPage, DocumentSource, DocumentStatus
 from .core.models.event import EventType
 from .core.recall_context import context_text
 from .dream import DreamConfig, DreamMode, DreamResult, DreamRunInfo, DreamScope, OpKind
@@ -105,6 +105,7 @@ __all__ = [
     "DocumentSource",
     "DocumentPage",
     "DocumentCursor",
+    "DocumentStatus",
     # Engine functions
     "create_engine",
     "list_engines",
--- a/src/khora/config/schema.py
+++ b/src/khora/config/schema.py
@@ -1578,6 +1578,20 @@ class QuerySettings(BaseSettings):
         default=0.05, ge=0.0, le=1.0, description="Minimum entity similarity threshold"
     )
 
+    # Document enumeration (list_documents) keyset scan bound. When a filter is
+    # active the coordinator over-fetches raw rows to absorb post-filter
+    # rejection before giving up on filling a page; this caps the total rows a
+    # single page may scan at ``max(limit × multiplier, min_bound)``. Not a
+    # public list_documents parameter — tuned here only. Raising the multiplier
+    # trades more scan work per page for fewer under-full pages under a very
+    # selective filter; the floor keeps small page limits from starving.
+    document_scan_overfetch_multiplier: int = Field(
+        default=10, ge=1, description="Per-page scan bound = max(limit × this, document_scan_min_bound)"
+    )
+    document_scan_min_bound: int = Field(
+        default=1000, ge=1, description="Floor for the per-page document enumeration scan bound"
+    )
+
     # Fusion weights. Wired onto the default VectorCypher engine in #1406
     # (previously dead there - the engine read only VectorCypherConfig). The
     # defaults were canonicalized to the values the engine already used
--- a/src/khora/config/schema.py
+++ b/src/khora/config/schema.py
@@ -1581,12 +1581,12 @@ class QuerySettings(BaseSettings):
     # Document enumeration (list_documents) keyset scan bound. When a filter is
     # active the coordinator over-fetches raw rows to absorb post-filter
     # rejection before giving up on filling a page; this caps the total rows a
-    # single page may scan at ``max(limit × multiplier, min_bound)``. Not a
+    # single page may scan at ``max(limit x multiplier, min_bound)``. Not a
     # public list_documents parameter — tuned here only. Raising the multiplier
     # trades more scan work per page for fewer under-full pages under a very
     # selective filter; the floor keeps small page limits from starving.
     document_scan_overfetch_multiplier: int = Field(
-        default=10, ge=1, le=1000, description="Per-page scan bound = max(limit × this, document_scan_min_bound)"
+        default=10, ge=1, le=1000, description="Per-page scan bound = max(limit x this, document_scan_min_bound)"
     )
     document_scan_min_bound: int = Field(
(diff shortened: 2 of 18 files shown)


filippetrovic:
## Multi-perspective review

Reviewers: correctness, devil's-advocate, database, security, performance, test-quality, docs (7 independent passes over the diff). Consolidated below; most findings already addressed in a follow-up hardening commit on this branch.

**Overall: APPROVE** with two follow-ups (limit clamp, CHANGELOG-at-release). No correctness or security blockers — the keyset walk is exactly-once, namespace isolation is sound, and there is no SQL injection.

### HIGH
- **H1 — Per-step sizing `min(need, remaining)` amplifies round-trips under a selective, non-pushed filter.** Consensus across correctness/perf/db/devil's-advocate. Not a wrong-result bug (results and exactly-once resume stay correct); the tail steps shrink toward 1 row, so a page can cost up to `scan_bound` round-trips. **Accepted & documented** (code comment added): per-step over-fetch is a deliberate non-goal because safe truncation would require deriving a resume cursor from a converted `Document`, which the scan primitive forbids (now()/uuid5 masking). A future primitive that exposes per-row keys could lift this.
- **H2 — Short page is not an end signal, but `DocumentPage` is truthy/`len`-able**, so `while len==limit` / `while page:` are silently wrong; only `while not page.exhausted`. **Mitigated**: facade + `DocumentPage.exhausted` docstrings now state the correct loop, and a new test asserts continuation across a short/empty non-exhausted page. (An `async for` iterator helper is a reasonable future add.)

### MEDIUM
- **M1 — `RelationalBackendProtocol.scan_documents` was a non-abstract `...` no-op.** ✅ Fixed: default now raises `NotImplementedError` (kept non-abstract so nominal test doubles still construct).
- **M2 — `limit` has no upper clamp → memory/DoS** (`SELECT *` × `scan_bound = limit×10`). ⚠️ Flagged: authenticated caller passing a huge `limit` can materialize a whole namespace. Needs a product decision on a max-page ceiling; the two scan-bound *config* knobs now have `le` bounds, but `limit` itself is not clamped.
- **M3 — No CHANGELOG entry** for the public-surface change (`list_documents` return `list[Document]` → `DocumentPage`; new exports). ⚠️ Flagged for the release step (khora's CHANGELOG is release-coupled).
- **M4 — `RecallFilterUnsupportedError` in the facade `Raises:` is unreachable** (enumeration compiles in split mode). ✅ Fixed (removed).
- **M5 — `DocumentPage.exhausted` docstring described the non-exhausted case.** ✅ Fixed.
- **M6 — `DocumentStatus` is in the public signature but was not exported.** ✅ Fixed (added to `__all__`).
- **M7 — `after` accepting a raw `Document`** (foreign/hand-built → silent wrong resume). ✅ Mitigated via docstring caveats on `after` and `DocumentCursor`.
- **M8 — Scan-bound default duplicated** (coordinator `max(limit×10,1000)` vs the config knobs). ⚠️ Minor: the fallback only fires on direct coordinator calls; noted.

### LOW
- **L1 — `scan_bound` was unvalidated** (`<=0` broke the `next_after is None iff exhausted` invariant for direct callers). ✅ Fixed (raises `ValueError`).
- **L2 — `post_filtered_keys` unions `consumed_keys` across steps.** Exact today (compile is deterministic per backend+dialect+filter); ✅ comment added asserting the invariance it rests on.
- **L3 — `DocumentCursor` is a transparent public dataclass documented as opaque.** Doc caveat added; an opaque token is a possible future hardening.

### Security Audit
Change surface is a namespace-scoped internal read path — no new endpoints, auth, crypto, or dependencies, so the OWASP/dependency/crypto categories are N/A. Findings:
- **IDOR / tenant isolation: SOUND.** Every path resolves stable→row namespace id at the facade and threads `namespace_id` unchanged to `build_documents_scan_query`, which unconditionally applies `namespace_id ==`. A forged/foreign `after` cursor only sets a keyset position; the query stays namespace-scoped.
- **Injection: NONE.** Filters compile to bound SQLAlchemy params / a Python predicate — no string interpolation.
- **DoS: see M2** (unclamped `limit`) — the one actionable security item.
- **Info disclosure: none** — `FieldError` copy is static schema metadata; the cursor carries only the caller's own `(created_at, id)`.

### Tests
The co-located suite was extended to cover the gaps reviewers flagged: intra-page re-stepping under a post-filtered `created_at`, the scan-bound-consumed short/not-exhausted branch with exactly-once resume, resume across a `created_at` tie block, empty namespace, resume past the end, the `scan_bound` guard, and validation-before-engine-dispatch. `post_filtered_keys` is now asserted by value.

msrdic (approved):

filippetrovic:
Addressed CodeRabbit's actionable comments in 20b505ee:

- **Ambiguous Unicode `×` (RUF001/RUF003)** in the scan-bound comments/description across `config/schema.py`, `khora.py`, `coordinator.py`, `tests/unit/helpers.py`, `tests/unit/test_khora.py` → replaced with ASCII `x`.
- **`DocumentPage.__slots__` not sorted (RUF023)** → reordered.
- **Filter tests asserted membership, not order** → now compare ordered lists via `_expected_order` so a post-filter ordering regression fails.

Already handled in the earlier hardening commit (5f720d6):
- **`scan_bound` validation** (Major) — guarded with `ValueError` for `< 1`.
- **Scan-bound exhaustion path coverage** — `test_multistep_page_fill_and_short_page_not_exhausted` exercises the short/not-exhausted branch (`scan_bound=3`) with exactly-once resume.

Verified: `ruff --select RUF001,RUF002,RUF003,RUF023` clean on the touched files; `make lint` + format clean; enumeration/facade tests green.

Issue #1221, opened 2026-06-15
Khora.forget() / re-ingest never deletes old chunks on backend=surrealdb


damirkrstanovic:
## Summary

On `backend=surrealdb`, deleting a document's chunks is a silent no-op.
`Khora.forget()` and re-ingest (`Khora.remember()` with a matching `external_id`)
both route to the SurrealDB vector adapter's `delete_chunks_by_document()`, which
counts and deletes with `WHERE document = $doc AND namespace_id = $ns`. But the
chunk-write path sets only the `namespace` record-link and never populates the
scalar `namespace_id` column, so that condition matches **0 rows**. The old chunks
survive, and the method returns `0` as if it had cleanly succeeded.

Consequences: `forget()` leaves the content recallable; re-ingesting a changed
document accumulates stale + duplicate chunks (recall returns outdated text); the
`chunk` table grows unbounded; and every caller sees a clean return value. Every
sibling read method in the same adapter uses the correct compound condition
`(namespace = $ns_rid OR namespace.namespace_id = $ns_str)` — only the
delete/count queries use the scalar-only form, and the writes never set the scalar
they filter on.

## Environment

- Khora 0.20.0, commit `6f931ed4`
- Backend: `surrealdb` (behavioral repro run on `mode=memory`; the defect is in the
  query text / write field-set and is mode-independent)
- The repro drives the real `SurrealDBVectorAdapter` against an in-process
  SurrealDB engine — no server needed.

## Where the asymmetry is

Write — `create_chunk` / `create_chunks_batch` set the record-link only
(`vector.py` ~120-136 and ~150-167); `namespace_id` is never in the field set:

```sql
CREATE $rid SET namespace = $ns_rid, document = $doc_rid, content = $content, ...
```

The schema declares the scalar optional with **no DEFAULT** (`schema.py:90`), so it
is always `NONE`:

```sql
DEFINE FIELD IF NOT EXISTS namespace_id ON chunk TYPE option<string>;
```

Delete + count filter on that never-populated scalar (`vector.py:240, 248`):

```sql
SELECT count() AS cnt FROM chunk WHERE document = $doc_rid AND namespace_id = $ns GROUP ALL
DELETE FROM chunk WHERE document = $doc_rid AND namespace_id = $ns
```

Every read method uses the compound condition instead (`vector.py:178, 198, 218,
327, 406, 515, 533, 560, 606`):

```sql
[...]

PR #1595, opened 2026-08-07
Restructure OR-defeating SurrealDB chunk and live-filter reads into index-eligible legs

filippetrovic:
## What & why

Follow-up to the SurrealQL `OR`-in-`WHERE` sweep in #1590. Two disjunction shapes collapsed the SurrealDB 2.x planner to `Iterate Table` on **recall hot paths** — a full cross-tenant scan of the entire corpus. All plans below were EXPLAIN-measured on the bundled embedded core (`memory://`).

The corrected premise from #1590 holds: an `OR` unions index scans **only when every disjunct is a comparison on a single-field index**; a record traversal (`namespace.namespace_id`) never qualifies, and one unservable disjunct drops the whole statement — namespace prefix included — to a table scan.

## chunk dual-form namespace scope

`(namespace = $ns_rid OR namespace.namespace_id = $ns_str)` → **Iterate Table**. The traversal leg is unservable. Fix:

- Rewrite the traversal to the **scalar `namespace_id`** (denormalised onto every chunk since #1221) and add a single-field **`idx_chunk_namespace_id`**. Both legs now union index scans.
- `get_chunk`, `get_chunks_batch`, `get_chunks_by_document`, `search_fulltext`, and the delete predicate now plan `Iterate Index`.
- **`search_similar`** is split into two OR-free legs merged in Python: its `embedding IS NOT NULL` guard poisons *any* in-query disjunction (measured), and `vector::dot()` errors on a `NONE` embedding so the guard can't move to Python. Each leg plans `Iterate Index`; results merge by similarity, de-duplicated by id.
- **`delete_chunks_by_document`** now counts via `DELETE … RETURN BEFORE` instead of `count() … GROUP ALL` — the aggregate count double-counts rows matching both union legs on this engine (measured 80 for 20 rows), while the `SELECT`/`DELETE` themselves are correct.

## valid_until live filter

`(valid_until IS NONE OR valid_until > time::now())` on `list_entities` / `list_relationships` → **Iterate Table**. Split into two OR-free legs merged in Python (the #1590 pattern) via a shared `_query_live_ordered` helper, preserving `created_at DESC` ordering and `offset`/`limit` pagination. NONE-semantics unchanged — no schema/behaviour change, no backfill.

## Design decision (kept both namespace forms)

The scalar disjunct is **not** dropped: `update_last_accessed` deliberately expands to all namespace row-ids, confirming the scalar genuinely matters under namespace versioning. Keeping both forms is provably no-coverage-loss vs. today; the change is query-shape + one additive index only.

## Tests

New `test_chunk_valid_until_query_plans.py` drives the **real adapters** through a recording connection and re-issues each captured statement with `EXPLAIN`, asserting no read falls to `Iterate Table` — plus correctness (retired-row hiding, pagination walk, namespace scoping, accurate delete count) and a premise-pin that the replaced disjunctive shapes really do table-scan on this engine. Existing mock-based unit tests updated for the new query structure.

- New plan+correctness tests: 13 passed
- Full unit suite: 10212 passed, 35 skipped
- SurrealDB embedded integration tests: passed

## Summary by CodeRabbit

- **Performance Improvements**
  - Improved namespace-scoped chunk, entity, relationship, similarity, and full-text searches for faster indexed query execution.
  - Enhanced live-record filtering, ordering, de-duplication, and pagination.
  - Improved chunk deletion accuracy and efficiency.

- **Bug Fixes**
  - Prevented retired records from appearing in live results.
  - Preserved namespace isolation and correct newest-first ordering.

- **Tests**
  - Added comprehensive coverage for query plans, indexing, pagination, filtering, de-duplication, and deletion behavior.

- **Documentation**
  - Clarified migration locking behavior, timeout expectations, and deployment guidance.

### DIFF
--- a/src/khora/db/migrations/versions/054_documents_namespace_created_at_id.py
+++ b/src/khora/db/migrations/versions/054_documents_namespace_created_at_id.py
@@ -4,6 +4,27 @@ Revision ID: 054_documents_namespace_created_at_id
 Revises: 053_khora_chunks_bookkeeping_to_chunker_info
 Create Date: 2026-08-04
 
+Operational profile — read this first
+-------------------------------------
+* **This migration can be slow, and it holds the migration advisory lock for
+  the whole time it runs.** A btree ``CREATE INDEX CONCURRENTLY`` over a large
+  ``documents`` table takes as long as it takes, and ``run_migrations()`` takes
+  a session-scoped ``pg_advisory_lock`` *before* Alembic's transaction
+  demarcation and holds it for the entire run — including the autocommit
+  ``CREATE INDEX CONCURRENTLY`` block below.
+* **A concurrent caller gets 60s, then fails to boot.** It waits only 60s for
+  that lock before raising ``TimeoutError``, which surfaces as
+  ``MigrationResult(success=False)`` and then ``RuntimeError: Database
+  migration failed`` at startup. A concurrent build over a large ``documents``
+  table can easily exceed 60s, so on a large deployment every other service
+  booting with ``run_migrations=True`` would fail to start for the duration of
+  the build.
+* **Run this migration out-of-band on such deployments**, rather than during a
+  rolling deploy.
+* **Migration 029's BRIN build is not a precedent for "concurrent builds are
+  fine".** It does not carry this risk — BRIN indexes are KB-sized and build in
+  seconds.
+
 ``list_documents`` now pins a total order — ``ORDER BY created_at DESC, id
 DESC`` — across the relational backends so that offset pagination cannot drop
 or repeat a row when two documents share a ``created_at``. The pre-existing
@@ -30,27 +51,15 @@ added the 2-column index for ``get_last_activity_at()``'s
 query on its ``(namespace_id, created_at)`` prefix. The new index is created
 **before** the old one is dropped so that query never runs unindexed.
 
-Both dialects are handled. Postgres builds concurrently (see the caveats
-below); every other dialect takes the plain-DDL branch, which is what migration
-019 already did when it created the 2-column index unconditionally. That keeps
+Both dialects are handled. Postgres builds concurrently (see the operational
+profile above and the invalid-index caveat below); every other dialect takes the
+plain-DDL branch, which is what migration 019 already did when it created the
+2-column index unconditionally. That keeps
 the invariant simple: wherever 019 ran, 054 runs, so the ORM declaration and the
 physical schema agree on every backend. The embedded ``sqlite_lance`` stack runs
 this chain too, and its own full-drain pagination callers are the ones that
 motivated the change, so skipping SQLite would have left the regression in place
 on the stack it was measured on.
-
-Operational caveat — this migration can be slow, and it holds the migration
-advisory lock while it runs. ``run_migrations()`` takes a session-scoped
-``pg_advisory_lock`` *before* Alembic's transaction demarcation and holds it for
-the whole run, including the autocommit block below. A concurrent caller waits
-only 60s before raising ``TimeoutError``, which surfaces as
-``MigrationResult(success=False)`` and then ``RuntimeError: Database migration
-failed`` at startup. A btree ``CREATE INDEX CONCURRENTLY`` over a large
-``documents`` table can easily exceed that, so on a large deployment every other
-service booting with ``run_migrations=True`` would fail to start for the
-duration of the build. Run this migration out-of-band on such deployments rather
-than during a rolling deploy. (Migration 029's BRIN build does not carry this
-risk — BRIN indexes are KB-sized and build in seconds.)
 
 Invalid-index recovery — a failed ``CREATE INDEX CONCURRENTLY`` (deadlock,
 cancellation, connection loss) leaves an **INVALID** index behind: ignored by
--- a/src/khora/storage/backends/surrealdb/graph.py
+++ b/src/khora/storage/backends/surrealdb/graph.py
@@ -141,6 +141,54 @@ def _row_to_relationship(row: dict[str, Any]) -> Relationship:
     )
 
 
+_LIVE_LEGS: tuple[str, str] = ("valid_until IS NONE", "valid_until > time::now()")
+"""The two OR-free legs of the ``(valid_until IS NONE OR valid_until > now)`` live
+filter, run separately and merged in Python (#1595).
+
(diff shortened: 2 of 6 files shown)


filippetrovic:
## Multi-perspective review — PR #1595

_Reviewed across five independent lenses: database/query-plan, Python correctness, performance, adversarial, and security/tests/docs._

> Infrastructure note: this environment lacks the agent-teams layer (`TeamCreate` unavailable), so the review was run as five parallel review agents rather than a formal `/team:review` team. Same substance, no team-config artifact.

**Overall: REQUEST CHANGES** — one real pagination regression (M1) and one defensive-consistency gap (M2) are worth fixing before merge; H1 needs an author call. Everything else is nits or acknowledged tradeoffs. The core plan reasoning (index-eligible legs, the subset guarantee, tenant isolation on every leg, the `DELETE … RETURN BEFORE` count fix) was verified sound by multiple reviewers.

### HIGH
- **H1 — traversal→scalar may not be coverage-equivalent for row-id-linked chunks.** The removed `namespace.namespace_id` traversal dereferenced the linked namespace row's *stable* id; the new scalar leg matches only a chunk whose denormalised `namespace_id` equals the stable id. A chunk linked by a namespace *row* id with a non-stable/NULL scalar matched the old traversal but neither new leg — recall would silently return fewer rows and `delete_chunks_by_document` would under-delete. Reads pass the stable id (not namespace-resolved), and `update_last_accessed` resolves all namespace row-ids precisely because this mixed case is treated as real. _Mitigant:_ shipped `count_chunks`/`list_chunks` already scope by `namespace = $ns_rid` alone, so the new reads are a **superset** of already-relied-on chunk scoping — any row-id-linked chunk this misses is already invisible to those methods. Entity reads correctly keep the traversal (it plans `Iterate Index` there — the ticket's "already fine" set — whereas on `chunk` it table-scanned). _Action:_ confirm no live write path produces a row-id-linked chunk (or accept alignment with `count_chunks` precedent); the "covers the same rows" comment is being scoped to stop over-claiming.

### MEDIUM
- **M1 — `_query_live_ordered` can drop/duplicate rows across pages under `created_at` ties.** Each leg's SQL is `ORDER BY created_at DESC LIMIT $window` with no secondary key, but the Python merge re-sorts by `(created_at, id)`. When a tie group exceeds the page window, the per-leg truncation is arbitrary w.r.t. the id tiebreak, so consecutive `offset` pages can skip some tied rows and repeat others. The prior single `LIMIT/START` statement walked one stable server order and didn't. **Fixing:** add `id DESC` to each leg's `ORDER BY` (the same tiebreak migration #1590/054 uses), so the fetched window is a true prefix of the merge order. Adding a tied-timestamp pagination regression test.
- **M2 — `get_chunks_by_document` / `search_fulltext` lack the id-dedup that `delete`/`search_similar` got.** They have the same `(namespace = … OR namespace_id = …)` union; the PR's own DELETE rationale says the union double-produces rows on the measured engine. Embedded 2.0.0 returns unique rows (measured), but `ws://` 2.3.x is unverified. **Fixing:** dedup both by chunk id for consistency and remote-mode safety.
- **M3 — `search_similar` computes `vector::dot` ~2× (perf).** The two legs overlap for dual-form rows, so similarity is computed twice per chunk on every recall. Acceptable vs. the ~200× reduction from cutting the whole-corpus table scan to a tenant-scoped index scan; a scalar-`namespace_id` backfill (to drop the record-link leg) would remove it. _Deferred — fast-follow._
- **M4 — `_query_live_ordered` deep-offset amplification (perf).** Fetches `2×(offset+limit)` rows + a Python sort; fine for the default `offset=0` and the multi-tenant regime, a regression for single-tenant deep drains. Fast-follow: far-future `valid_until` sentinel (collapses to one indexed statement) or keyset. _Deferred._
- **M5 — `min_similarity` path is untested.** **Fixing:** adding coverage.

### LOW
- **L1** — `idx_chunk_namespace_id` may overlap `idx_chunk_ns_session`'s leading column; kept deliberately for cross-engine-version robustness of the OR-union (documented in schema).
- **L2** — sort key mixes naive/aware datetimes only if `_parse_dt` ever returns naive alongside a `None` fallback; latent (SurrealDB returns UTC-aware). Hardening the key.
- **L3** — dead `ef`/`limit` bindings passed to the `search_similar` legs (SQL inlines the limit). Minor cleanup.
- **L4** — the `DELETE` predicate is guarded by a text tripwire but not an EXPLAIN assertion (it's not a SELECT). Measured `Iterate Index Count` manually.
- **L5** — `search_similar` docstring says `vector::similarity::cosine`; code uses `vector::dot` (pre-existing). Fixing the docstring while here.
- **L6** — the count-double-count comment was measured on 2.3.x; scoping it since embedded is 2.0.0.

### Security / tenant-isolation — clean
Every split preserves the namespace conjunct on **both** legs (guarded by `test_search_similar_stays_namespace_scoped` and the delete cross-tenant-survivor assertion). No untrusted f-string interpolation — window binds via `$__live_window`, metadata keys go through `_sanitize_field_name`, `LIMIT` is int-cast. No new endpoints/clients/crypto/deps, so no separate OWASP/dependency audit surface. No IDOR introduced.

filippetrovic:
### Review addressed (commit 3b4e0c1e)

- **M1 (fixed)** — added `id DESC` secondary sort to each `_query_live_ordered` leg so the fetched window is a true prefix of the merge order; created_at ties no longer drop/repeat across offset pages. New regression test (`test_list_entities_pagination_survives_created_at_ties`).
- **M2 (fixed)** — `get_chunks_by_document` and `search_fulltext` now de-dup by chunk id, matching `search_similar`/delete.
- **M5 (fixed)** — added `test_search_similar_respects_min_similarity`.
- **L3/L5/L6 (fixed)** — dropped dead `ef`/`limit` bindings, corrected the `vector::dot` docstring, scoped the count-double-count comment to the 2.3.x core.
- **H1 (scoped, not code-changed)** — comment now states the rewrite covers the record-link + scalar forms and aligns with `count_chunks`/`list_chunks` (which already scope by record-link alone), and does *not* reproduce the old traversal's row-id-link dereference — no read path relies on that today. A true fix belongs with the namespace dual-ID normalisation, not here.
- **M3/M4 (deferred, fast-follow)** — `search_similar` 2× compute and deep-offset amplification are acknowledged tradeoffs vs. the ~200× table-scan reduction; both dissolve with the scalar-`namespace_id` backfill / far-future `valid_until` sentinel noted in the review.

msrdic (approved):

PR #1596, opened 2026-08-07
Add documents-enumeration conformance legs to the recall-filter harness

filippetrovic:
## Summary

Extends the internal recall-filter conformance harness so the existing filter-case corpus also runs against the **document-enumeration** surface, not just the chunk-recall surface. Concretely, it drives the storage coordinator's document keyset-scan primitive (`scan_documents_page`) — the primitive that backs the public `list_documents(filter=…)` enumeration API — and asserts that every backend's surviving row set equals the in-memory `compile_python` oracle.

- **New executor seam.** A single `DocumentsExecutor` plus an injected `DocumentsRunner` walks the real coordinator document-scan path to exhaustion and maps surviving rows back to seed ids. Seeding goes exclusively through the `create_document` write API, namespace-partitioned per case (xdist-safe).
- **Both read modes per case.** Natural pushdown (backend prefilter + coordinator post-filter) and a forced-residual mode (raw enumeration with `filter_ast` withheld, then the identical `compile_python` post-filter applied over the round-tripped document row), so oracle equality is checked on every backend, not only where residuals occur naturally.
- **Case-family carryover.** F-OP over the nine enumerable document keys, plus F-EXISTS, F-COERCE, F-OBJEQ, F-DOTKEY, F-LOGIC, F-NULLVAL, F-SEL and the metadata-datetime shapes of F-DATES. The chunk-only event-time key is not enumerable on this surface and is encoded as a structured-rejection case with its error copy asserted.
- **SurrealDB representation exclusions carried over** (metadata datetime ISO-form divergence; present-JSON-null drop on write), each with a documented capability reason — never a silent skip. The unsafe metadata-segment case defers to the post-filter under split semantics rather than raising.
- **Coverage meta-test extended** to the enumerable-key set: an enumerable key with zero documents-target operator case fails CI.
- **CI + local coverage.** New no-Docker embedded unit smoke legs (raw-sqlite, sqlite_lance, embedded SurrealDB) in both modes, plus a new CI matrix job covering all four relational backends. This is the first conformance coverage of the raw `sqlite` relational backend.

Test-harness change only — no production behavior is modified (`src/` diff is a single additive file, `+488/−0`).

## Test plan

- [x] Unit tests pass (`tests/unit/filter/` incl. the new documents module and the extended coverage meta-test; full unit suite green, coverage 84%)
- [x] Embedded documents-target legs pass in both modes on raw-sqlite, sqlite_lance, and embedded SurrealDB (no Docker)
- [x] Chunk-target conformance legs unregressed
- [ ] Integration matrix (Postgres + SurrealDB documents legs) — exercised by the new CI job

## Summary by CodeRabbit

- **New Features**
  - Added document-enumeration conformance coverage alongside existing chunk-recall checks.
  - Added support for validating document filtering across SQLite, SQLite/LanceDB, SurrealDB, and PostgreSQL.
  - Added deterministic document seed data and backend-specific test execution modes.

- **Bug Fixes**
  - Added validation and clear errors for unsupported document filters, including `occurred_at`.
  - Added coverage checks to ensure enumerable system fields are tested consistently.

### DIFF
--- a/.github/workflows/filter-conformance.yml
+++ b/.github/workflows/filter-conformance.yml
@@ -1,14 +1,27 @@
 name: Filter Conformance
 
 # ---------------------------------------------------------------------------
-# Recall-filter conformance matrix. One leg per engine; each leg provisions
+# Recall-filter conformance matrices. One leg per engine; each leg provisions
 # ONLY the live store it needs via an `if:`-guarded `docker run` (Postgres for
 # the postgres leg, Neo4j for the cypher leg, Weaviate for the weaviate leg),
 # so a leg that runs in-process never waits on an idle container.
 #
+# Two jobs, because there are two SURFACES with different corpora and different
+# contracts:
+#
+# * `conformance` — the chunk RECALL surface. Postgres/SurrealDB are total-exact
+#   there and a filter they cannot express RAISES, so the corpus carries
+#   `expect_unsupported` legs. Selected by KHORA_CONFORMANCE_BACKEND.
+# * `documents-conformance` — the document ENUMERATION surface. Uniformly split +
+#   post-filter on every backend, so nothing raises and every case is a plain
+#   row-set comparison; each case also runs in both a natural (backend pushdown)
+#   and a residual (post-filter-only) mode. Selected by
+#   KHORA_CONFORMANCE_DOC_BACKEND, and its Postgres leg has its own corpus, its
+#   own seed invocation and its own seed-map path.
+#
 # This file is intentionally separate from ci.yml: GitHub schedules a separate
-# workflow file as its own check, so this matrix runs concurrently with the
-# main `test` job and gates the PR on its own.
+# workflow file as its own check, so these matrices run concurrently with the
+# main `test` job and gate the PR on their own.
 # ---------------------------------------------------------------------------
 
 on:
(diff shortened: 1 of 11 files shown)


filippetrovic:

Eight independent reviewers (security, performance, correctness, test, database, docs, spec-drift, devil's advocate) plus a dedicated security audit. **No CRITICAL or HIGH findings. Overall: APPROVE.** This is an additive, test-harness-only change (`src/` diff is a single file, `+488/−0`, no production behavior modified); reviewers ran the embedded legs end-to-end (raw-sqlite, sqlite_lance, embedded SurrealDB) in both modes with zero oracle mismatches across all 193 cases.

### Highlights the reviewers called out as well-done
- **Oracle leg is falsifiable, not assumed** — `documents_oracle_survivors(case) == expected_ids` re-derives every carried-over expectation against a real document-row mapping, and correctly does **not** synthesize the chunk-only event-time key.
- **Coverage meta-test asserts the tighter invariant** — the nine enumerable keys are covered **and** the event-time key's *absence* is asserted, with a falsifiability guard.
- **Single-executor + injected-runner design** is the right, smaller call (the enumeration surface compiles nothing in-harness — one thing to drive, a bool for the mode).

### MEDIUM
- **M1 · [CI/efficiency]** `.github/workflows/filter-conformance.yml` — the three in-process embedded legs (sqlite / sqlite_lance / surrealdb) each run as a separate runner paying its own `uv sync`; only the postgres leg needs a container + out-of-band seed. Folding the three embedded legs into one job (or into the existing `conformance` job's in-process legs) would drop two dependency-install cycles per PR for identical coverage. Non-blocking — the current split mirrors the existing conformance job's per-leg structure. **Suggested follow-up.**

### LOW
- **L1 · [Docs]** Residual-mode docstrings imply datetime round-trip coverage on the embedded tiers that natural mode already provides there (date keys are withheld from pushdown on embedded, so a deferred leaf hits the identical `compile_python` post-filter in both modes). Residual's unique value is concentrated on the postgres leg (real `timestamptz` pushdown cross-check). Worth softening the wording; residual is cheap (a `parametrize`, not a runner) so keep it.
- **L2 · [Coverage]** `_documents_surreal_excluded` prunes the metadata-`$date` bucket class-wide (12 cases) though only ~4 equality shapes are measured to diverge; the 8 passing range shapes are then not run on any surreal leg, forfeiting live regression detection there. Defensible over-prune (under-prune would fail loudly); could be narrowed to a per-case empirical prune later.
- **L3 · [Consistency]** `_reject_non_enumerable_keys` (production) matches the event-time key by exact tuple while the harness/`_documents_eligible` match by path prefix — functionally identical for a single-segment system key; align if touched.
- **L4 · [Docs]** `tests/unit/filter/test_documents_conformance.py` module docstring says "186-case" — the corpus is now **193**. Stale numeral.
- **L5 · [Lifecycle]** Embedded and PG read-only coordinators are `lru_cache`d for the process lifetime and never `disconnect()`/dispose their engines. Acceptable for a test harness (process exit reclaims; matches the chunk-leg pattern; namespaces are private + reads namespace-scoped, so no cross-test contamination). The seed entrypoint path *does* dispose cleanly.
- **L6 · [Performance]** `run_case_for_backend` builds a `records` mapping list that `DocumentsExecutor.survivors` ignores (live rows are truth); ≤10 tiny mappings per test — negligible.
- **L7 · [Test]** The no-Docker unit smoke leg runs a real Alembic chain + embedded-SurrealDB schema init under `pytest.mark.unit` (~2s). Deliberate, but blurs the unit boundary with migration I/O.
- **L8 · [Docs, pre-existing / out of scope]** `src/khora/filter/conformance.py:147` comment says "eight denormalized document system keys" but `_DOC_STRING_KEYS` holds seven. Not introduced by this PR (the new docstrings correctly say "seven"); flagged for a future pass.

### Security Audit — PASS
No CRITICAL/HIGH/MEDIUM across the 16 categories. Verified: seeding is exclusively through the `create_document` write API (no raw SQL/SurrealQL emitted); the SurrealDB field-name safe-segment guard is `\Z`-anchored and shared between the boolean gate and the emit path (no drift), with user values always parameter-bound; the metadata unsafe-segment case defers to the in-memory post-filter under split semantics. INFO only: the Postgres CI leg uses an ephemeral, localhost-bound, torn-down test container credential (identical to the existing conformance job — not a real secret); GitHub Actions are tag-pinned (`@v6`/`@v7`) consistent with the existing workflow; temp-dir and seed-map file I/O use non-user-controlled paths.

### Assessment
**APPROVE.** No blocking findings. M1 is a CI-efficiency follow-up; the LOW items are documentation/coverage polish, none affecting correctness or CI. Full acceptance-criteria alignment confirmed by the spec-drift review, including the two intentional, documented deviations from the original spec (forced-residual realized by withholding the filter AST since no capabilities-override hook exists; the unsafe-segment case deferring rather than raising, per the later amendment folding that carve-out into an ordinary split case).

msrdic (approved):

